PR #1, opened 2017-07-22
Make it cool again ;)

ignatenkobrain:


ignatenkobrain:
```python
from gi.repository import Modulemd
mmd = Modulemd.ModuleMetadata()
print(mmd.props.buildrequires)
mmd.props.buildrequires = {'foo': 'bar'}
print(mmd.props.buildrequires)
```

works now quite well. Although `mmd.props.buildrequires['foo'] = 'bar'` doesn't have any effect (will debug tomorrow)...

@sgallagher @contyk I think we really need to stop using this `ModuleMetadata` thing.... `Module` should be enough. We can create compatibility wrapper for this if really want to. typing all this stuff made me cry this night.

modulemd/modulemd-modulemetadata.c
```diff
@@ -1,6 +1,7 @@
 /* modulemd-metadata.c
  *
  * Copyright (C) 2017 Stephen Gallagher
+ * Copyright (C) 2017 Igor Gnatenko <ignatenkobrain@fedoraproject.org>
  *
```
sgallagher: Every change in this patch is unnecessary and makes it difficult to follow the rest of the changes. You removed a large number of commented-out lines that were placeholders for future work that needed to be done, changed variable names for no clear reason and reordered code. Not a single line of this appears to have been done for any valid reason. Or if it was, it is unclear and should have been justified in the commit message. This patch is rejected.

modulemd/modulemd-modulemetadata.c
```diff
@@ -435,11 +409,12 @@ modulemd_modulemetadata_set_buildrequires (ModulemdModuleMetadata *self,
...
+  g_return_if_fail (buildrequires);
 
-  if (buildrequires != self->buildrequires)
+  if (self->buildrequires != buildrequires)
     {
-      g_hash_table_destroy (self->buildrequires);
-      modulemd_copy_htable_strings (buildrequires, &self->buildrequires);
+      g_hash_table_unref (self->buildrequires);
+      self->buildrequires = g_hash_table_ref (buildrequires);
       g_object_notify_by_pspec (G_OBJECT (self),
```
sgallagher: This significantly changes the desired behavior. Please justify it.
The reason for forcing a deep copy here was to ensure that the caller cannot modify the contents of the hash table without triggering a signal (which would happen when saving the property back).

By just bumping the refcount and returning the original value, the contents can be `insert()`ed or `remove()`ed without causing the signal to fire, so consumers could get out of sync.

Furthermore, the commit message is completely useless. The phrase "totally busted" does not explain in what way it didn't work properly or why these changes would improve the situation.

Patch is rejected.

modulemd/modulemd-modulemetadata.c
```diff
@@ -121,6 +121,7 @@ modulemd_modulemetadata_set_name (ModulemdModuleMetadata *self,
                                   const gchar            *name)
 {
   g_return_if_fail (MODULEMD_IS_MODULEMETADATA (self));
+  g_return_if_fail (name);
```
sgallagher: These were explicitly left out because it's the only way to cause the value to be cleared entirely. That was a desired behavior. I understand the desire to validate input, but this was intentional. I will add this to the gtk-doc so it is clear.

Patch is rejected.

modulemd/meson.build
```diff
@@ -1,42 +1,41 @@
...
-test('test_modulemd_modulemetadata', test_modulemd_modulemetadata)
+#test_modulemd_modulemetadata = executable(
+#  'test_modulemd_modulemetadata',
+#  'test-modulemd-modulemetadata.c',
+#  link_with : modulemd_lib,
+#  dependencies : [
+#    gobject,
+#  ],
+#)
+#test('test_modulemd_modulemetadata', test_modulemd_modulemetadata)
```
sgallagher: Why did you remove my tests?

modulemd/meson.build
```diff
@@ -1,42 +1,41 @@
...
-    version: '0.1.0',
-
+  'modulemd',
+  sources : modulemd_srcs,
+  dependencies : [
+      gobject,
+  ],
+  install : true,
+  soversion: '0',
+  version: '0.0.0',
```
sgallagher: Why did you downgrade the library version?

sgallagher:
I accepted several of the patches into my branch (fixing conflicts). The majority of this patch needs to be reworked.

modulemd/modulemd-modulemetadata.c
```diff
@@ -400,331 +375,303 @@ modulemd_modulemetadata_get_summary (ModulemdModuleMetadata *self)
...
+  properties[PROP_TRACKER] =
+    g_param_spec_string ("tracker",
+                         "Module Bug Tracker",
+                         "A string property representing a link to the "
+                         "upstream bug tracker for this module.",
+                         "",
+                         G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS);
+
+  /**
+   * ModulemdModuleMetadata:buildrequires: (type GLib.HashTable(utf8,utf8)) (transfer container)
```
sgallagher: After considerable inspection, it turns out that this is the only line in this entire patch that has a functional effect. Why was this not part of the patch that "fixed" buildrequires at least? (Since that's where you change the transfer type as well).

modulemd/modulemd-modulemetadata.c
ignatenkobrain: well, then we should create Object on top of GHashTable and make functions which wrap g_hash_table_*() functions and sending signals. Hard destory in finalize() and deepcopy everywhere is **worst thing ever**.

modulemd/modulemd-modulemetadata.c
ignatenkobrain: it's fine, it is in separate patch which you can drop. I was surprised to see default value as `""`, but accepting to set it to `NULL`.

modulemd/modulemd-modulemetadata.c
ignatenkobrain: The code was just unreadable ;)

* there is no point in _free() function, no one ever should use such functions in GObject-based things, this will just lead in memory corruption at some point
* properties were renamed to match function names, because DESC vs set_description() is not really useful, also it wrongly translates to python property
* also I ordered code not alphabetically, but by most common properties (mdversion, name, stream, version, summary, description, ...)
* indentation and code style was inconsistent
* Probably something else I forgot about (IIRC there were trivial typos I fixed)

P.S. this was WIP patch, so commit messages might be not describing everything

modulemd/meson.build
ignatenkobrain: since this PR is WIP, it just didn't compile ;) e.g. because I removed _free() function which no one should ever use ;)

modulemd/meson.build
ignatenkobrain: I planned to make separate commit, but looks like I sneaked it to big one =(

well, it's just cosmetic thing. Feel free to ignore it.

modulemd/modulemd-modulemetadata.c
ignatenkobrain: ah right, I also fixed memory leaks

PR #5, opened 2018-01-15
EOL: mark as deprecated and accept NULL values without throwing warnings

nphilipp:


modulemd/modulemd-module.c
```diff
@@ -387,7 +387,13 @@ modulemd_module_set_eol (ModulemdModule *self, const GDate *date)
 
   if (!date)
     {
-      g_date_clear (self->eol, 1);
+      if (g_date_valid(self->eol))
+        {
+          g_date_clear (self->eol, 1);
+          g_object_notify_by_pspec (G_OBJECT (self),
+                                    md_properties[MD_PROP_EOL]);
+        }
```
sgallagher: Just add an `else` block here and do `g_date_clear()` also. I'd like to make sure that it's always explicitly cleared if we pass NULL.

nphilipp:
I'll pull the `g_date_clear()` call out of the conditional block which has the same effect.

sgallagher:
@nphilipp 
Well, we don't want to call the notify until the clear has actually occurred.

PR #6, opened 2018-02-07
Bumping the specification to v2

contyk:
The new version includes two incompatible changes:

* The eol field is removed in favor of servicelevels.
* The dependencies section is now a list of dictionaries, supporting the
  Stream Expansion proposal.  More in the comments.

This commit also moves spec.yaml to spec.v1.yaml and adds a spec.yaml
symlink pointing to the latest specification implemented by libmodulemd.

libmodulemd should be able to read v1 documents but doesn't have to
write them (although detecting when we could do that and allowing it in
such cases would be awesome).

Suggested behavior for v1->v2 transition:

* eol dates are converted to servicelevel of type "rawhide".  If such
  a service level already exists, the eol value is ignored.
* Old dependencies blocks become part of the new dependencies block
  list; the stream names are converted into lists including just that
  one item.


spec.v2.yaml
```diff
@@ -0,0 +1,336 @@
...
+        # Build dependencies of this module, optional
+        # Keys are module names, values are the stream names
+        # These modules define the buildroot for this module
+        buildrequires:
+            platform: and-its-stream-name
+            extra-build-env: and-its-stream-name-too
+        # Run-time dependencies of this module, optional
+        # Keys are module names, values are their stream names
+        requires:
+            platform: and-its-stream-name
```
sgallagher: I suspect that this section was meant to be removed. As it exists right now, it is un-parseable because you are mixing a dictionary and a sequence at the same hierarchy level.

PR #10, opened 2018-02-19
Describe how libmodulemd can be used from Python

hanzz:
I'm using gobject-introspection for first time and I presume I'm not the only one, so might be useful to mention in README.md.


sgallagher:
Approved, signed and merged as https://github.com/fedora-modularity/libmodulemd/commit/4bfa7a5b30f6659af0d895fbd3169656f1a40e4d

PR #11, opened 2018-02-20
Do not check if date returned by modulemd_module_get_eol is valid, it is always valid when not NULL.

hanzz:
Fixes "Warning: g_date_valid: assertion 'd != NULL' failed" during `module_v1.upgrade()` which is spamming ODCS test logs.

### DIFF
--- a/modulemd/modulemd-module.c
+++ b/modulemd/modulemd-module.c
@@ -2267,7 +2267,7 @@ _modulemd_upgrade_v1_to_v2 (ModulemdModule *self)
 
   /* Upgrade the EOL field to a "rawhide" servicelevel*/
   eol = modulemd_module_get_eol (self);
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
NAK, this is just workaround instead of proper fix. I'm working on this.

hanzz:
Fix also this place then: https://github.com/hanzz/libmodulemd/blob/master/modulemd/modulemd-yaml-emitter.c#L453

sgallagher:
An issue describing the problem you're trying to solve would be desirable.

mprahl:
@sgallagher I filed issue #17 for this.

sgallagher:
Closing this pull request as it is irrelevant in modulemd v2 and addressed in the libmodulemd-1.1 release.

Issue #19, opened 2018-02-28
Add _dup() and _peek() functions for libmodulemd


sgallagher:
After conversations with GNOME folks, it was made clear that a "proper" GObject API should have both `object_peek()` and `object_dup()` methods. The `object_peek()` methods would be the same functionality as currently provided by `object_get()` in libmodulemd and should be implemented as deprecating the latter.

`object_dup()` methods must return a deep copy of the property.

PR #20, opened 2018-03-01
(SEMVER-MINOR) Add peek(), dup() and copy() functions to all libmodulemd objects

sgallagher:
Addresses:
https://github.com/fedora-modularity/libmodulemd/issues/19


sgallagher:
Withdrawing this one as I have more changes coming (and reworked a few bits of this as a result).

Issue #39, opened 2018-04-12
libmodulemd should provide a prioritized merging routine for modules and defaults


sgallagher:
When dealing with multiple repositories, we need to handle how to merge content. For example, if we have a release repository (`fedora-modular`) and an updates repository (`fedora-modular-updates`) and both provides differing profiles in a modulemd-defaults file, we need to prefer the `fedora-modular-updates` version.

sgallagher:
Addendum: in the event that there is a collision between two entries with the same priority value, it must be considered an error.

PR #40, opened 2018-04-13
Add merging functions for modulemd-defaults streams

sgallagher:
This set of patches adds routines that clients can use to manage defaults coming from multiple locations. The client application will need to assign priorities to these YAML streams in an appropriate manner (for example, DNF might add a new config option to the repo file called `module_priority`).

The client can then construct a `Modulemd.Prioritizer` object using its `Prioritizer.add()` function which can be fed the array output from `Modulemd.objects_from_file()` (or `_from_string()`) and the priority value above. Once all of the repos are added at their appropriate priorities, call `Prioritizer.resolve()` to get back an object array that has been appropriately merged.

See the associated tests for details.

ping @dmach
Resolves: #39


modulemd/modulemd-prioritizer.c
```diff
@@ -0,0 +1,209 @@
...
+  g_autoptr (GPtrArray) concat_objects = NULL;
+  g_autoptr (GPtrArray) merged_objects = NULL;
+  gint64 *prio = NULL;
+  gsize i;
+
+  g_return_val_if_fail (MODULEMD_IS_PRIORITIZER (self), FALSE);
+  g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+
+  prio = g_new0 (gint64, 1);
+  *prio = priority;
```
ignatenkobrain: why not simply `prio = &priotiry`?

sgallagher: I originally had that and it took me almost two hours to discover that it was the source of a memory-corruption issue I was dealing with. The safest way to keep a key valid in GHashTable is to allocate it and associate that memory with the table.

modulemd/modulemd-simpleset.c
```diff
@@ -374,3 +374,68 @@ modulemd_simpleset_new (void)
...
+        }
+    }
+
+  /* If we made it here, everything must have matched */
+  ret = TRUE;
+
+done:
+  if (set_a)
+    {
+      for (i = 0; set_a[i]; i++)
```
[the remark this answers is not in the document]
sgallagher: Mostly because I didn't know it existed. I'll use it.

sgallagher:
I made the following changes:
* Fixed two minor memory leaks
* Switched to `g_strvfree()` for `Modulemd.SimpleSet.is_equal()`
* Adjusted the merging tests to match the API expectation (the internal state is not guaranteed after a merge conflict, even though it's safe today).

modulemd/modulemd-simpleset.c
```diff
@@ -374,3 +374,54 @@ modulemd_simpleset_new (void)
...
+  if (modulemd_simpleset_size (self) != modulemd_simpleset_size (other))
+    {
+      /* If they have a different number of strings in the set, they can't
+       * be identical.
+       */
+      ret = FALSE;
+      goto done;
+    }
+
+  set_a = modulemd_simpleset_dup (self);
```
ignatenkobrain: why not simply `g_auto (GStrv) set_a ...`

sgallagher: Didn't think of it. Updated.

modulemd/modulemd-simpleset.c
```diff
@@ -374,3 +374,52 @@ modulemd_simpleset_new (void)
...
+          /* No match, so this simpleset is not equal */
+          ret = FALSE;
+          goto done;
+        }
+    }
+
+  /* If we made it here, everything must have matched */
+  ret = TRUE;
+
+done:
```
ignatenkobrain: now you don't need to have goto, just `return FALSE` from inner loop and return TRUE here ;)

modulemd/test-modulemd-defaults.c
```diff
@@ -393,6 +393,331 @@ modulemd_defaults_test_missing_default_from_profile (DefaultsFixture *fixture,
...
+
+  objects = modulemd_objects_from_file (yaml_path, &error);
+  g_clear_pointer (&yaml_path, g_free);
+  g_assert_nonnull (objects);
+  g_assert_cmpint (objects->len, ==, 6);
+
+  merged_base = modulemd_merge_defaults (objects, NULL, FALSE, &error);
+  if (!merged_base)
+    {
+      fprintf (stderr, "Error merging defaults: %s\n", error->message);
```
ignatenkobrain: g_warning? or you can actually even use g_error.

sgallagher: I intentionally didn't use those because they terminate immediately. I meant to go back and test that the error->code and quark match as well and forgot to do that. I'll fix that up later.

Issue #45, opened 2018-04-17
Prioritizer: add() should throw an error on priority that is out of range


dmach:
When priority is a too big number it either overflows or add() fails:
`module_defaults_prioritizer.add(defaults, 99999999999)`
The result is that the added defaults are not present in the prioritizer output.

sgallagher:
OK, I'll add a `MODULE_DEFAULT_PRIORITY_MAX` `#define` to the public header and validate that the priority doesn't exceed that.

PR #49, opened 2018-04-18
Prioritizer: Set minimum and maximum values

sgallagher:
The Prioritizer will now disallow values below 0 or above 1000 and
will return FALSE and an error code if this occurs.

Fixes: https://github.com/fedora-modularity/libmodulemd/issues/45


sgallagher:
@contyk The reason it's signed is because GHashTable has built-in capability to handle gint64 keys, but not guint64 keys. It would be more effort than it's worth to make it signed.

Issue #50, opened 2018-04-18
Add functions for reading from FILE *


ignatenkobrain:
Using path to file is fine, but not when you are dealing with something complex.

PR #51, opened 2018-04-18
Add support for reading YAML from FILE *

sgallagher:
Fixes: https://github.com/fedora-modularity/libmodulemd/issues/50

### DIFF
--- a/modulemd/modulemd-defaults.c
+++ b/modulemd/modulemd-defaults.c
@@ -525,6 +525,58 @@ modulemd_defaults_new_from_string (const gchar *yaml_string, GError **error)
 
 
 /**
+ * modulemd_defaults_new_from_stream:
+ * @stream: A YAML stream containing the module metadata and other related
+ * information such as default streams.
+ * @error: (out): A #GError containing additional information if this function
+ * fails.
+ *
+ * Constructs a new #ModulemdDefaults object from the first valid
+ * modulemd-defaults document in the given module stream. This will ignore any
+ * documents of other types, malformed documents and defaults that appear later
+ * in the stream.
+ *
+ * Returns: A #ModulemdDefaults object constructed the first valid
+ * modulemd-defaults document in the given module stream. This must be freed
+ * with g_object_unref() when no longer needed.
+ *
+ * Since: 1.4
+ */
+ModulemdDefaults *
+modulemd_defaults_new_from_stream (FILE *stream, GError **error)
+{
+  GObject *object = NULL;
+  g_autoptr (GPtrArray) data = NULL;
+  ModulemdDefaults *defaults = NULL;
+
+  if (!parse_yaml_stream (stream, &data, error))
+    {
+      return NULL;
+    }
+
+  for (gsize i = 0; i < data->len; i++)
+    {
+      object = g_ptr_array_index (data, i);
+      if (MODULEMD_IS_DEFAULTS (object))
+        {
+          defaults = MODULEMD_DEFAULTS (g_object_ref (object));
+          break;
+        }
+    }
+
+  if (!defaults)
+    {
+      g_set_error (error,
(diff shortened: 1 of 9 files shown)


modulemd/modulemd-defaults.c
```diff
@@ -524,6 +524,60 @@ modulemd_defaults_new_from_string (const gchar *yaml_string, GError **error)
...
+  GObject *object = NULL;
+  GPtrArray *data = NULL;
+  ModulemdDefaults *defaults = NULL;
+
+  if (!parse_yaml_stream (stream, &data, error))
+    {
+      return NULL;
+    }
+
+  for (gsize i = 0; i < data->len; i++)
```
ignatenkobrain: shouldn't it be `guint`? https://developer.gnome.org/glib/stable/glib-Pointer-Arrays.html#GPtrArray

modulemd/modulemd-defaults.c
```diff
@@ -524,6 +524,60 @@ modulemd_defaults_new_from_string (const gchar *yaml_string, GError **error)
...
+    {
+      object = g_ptr_array_index (data, i);
+      if (MODULEMD_IS_DEFAULTS (object))
+        {
+          defaults = MODULEMD_DEFAULTS (g_object_ref (object));
```
ignatenkobrain: I don't think you need `MODULEMD_DEFAULTS` cast here... but not sure.

modulemd/modulemd-defaults.c
```diff
@@ -524,6 +524,60 @@ modulemd_defaults_new_from_string (const gchar *yaml_string, GError **error)
...
+ * modulemd-defaults document in the given module stream. This must be freed
+ * with g_object_unref() when no longer needed.
+ *
+ * Since: 1.4
+ */
+ModulemdDefaults *
+modulemd_defaults_new_from_stream (FILE *stream, GError **error)
+{
+  GObject *object = NULL;
+  GPtrArray *data = NULL;
```
ignatenkobrain: `g_autoptr()`?

modulemd/modulemd-module.c
```diff
@@ -3625,6 +3625,53 @@ modulemd_module_new_all_from_string_ext (const gchar *yaml_string,
...
+ * Return value: a new #ModulemdModule. When no longer needed, free it with
+ * g_object_unref().
+ *
+ * Since: 1.4
+ */
+ModulemdModule *
+modulemd_module_new_from_stream (FILE *stream)
+{
+  GError *error = NULL;
+  ModulemdModule *module = NULL;
```
ignatenkobrain: `g_autoptr`?

modulemd/modulemd-defaults.c
sgallagher: It's irrelevant, really. gsize is essentially size_t which is pretty much designed for this purpose. It's also signed and always going to be at least as large as guint.

modulemd/modulemd-defaults.c
sgallagher: I do, the compiler warns about it.

modulemd/modulemd-defaults.c
sgallagher: Ah, I basically just copied this from the `modulemd_defaults_new_from_file()` routine. But you're right; I've been using `g_autoptr()` for all newly-written code; best not to break with that here. I will fix it up.

modulemd/modulemd-module.c
sgallagher: Same as above.

sgallagher:
I've updated this PR addressing the issues @ignatenkobrain saw, as well as the following:
* Fixed a couple error messages that were copied from the `_file()` versions to say "stream"
* Added a proper error return, even for functions whose file or string versions don't have it. That was a poor design that we're stuck with until 2.0, but there's no reason to keep it for new functions.

Issue #43, opened 2018-04-17
Do not warn about missing stream from profile defaults.


dmach:
Example data:
```
document: modulemd-defaults
version: 1
data:
  module: testmodule
  stream: 2.0
  profiles:
    master: [default]
```

This triggers following output:
```
** Message: 09:15:48.608: Default stream missing from profile defaults
** Message: 09:15:48.608: Invalid [ModulemdDefaults] document [Default stream missing from profile defaults].
** Message: 09:15:48.608: Invalid document [Default stream missing from profile defaults]. Skipping it.
```
I think this behavior is incorrect. Consider following use cases:

1. specify default stream only; use the 'default' profile in modulemd or the empty 'default' fallback
2. override installation profiles without touching the default stream (or eventually don't have a default at all)

sgallagher:
This is behaving according to the [modulemd-defaults v1 specification](https://github.com/fedora-modularity/libmodulemd/blob/master/mod-defaults/spec.v1.yaml#L13). It is therefore not a bug, it is functioning as designed. If you want to change the specification, we need to have a discussion with several others, including @contyk and @james-antill 

For the record, this change would be to lessen the strictness of the document, so it's probably not necessary to bump the document version number to do it.

contyk:
I think it's a good idea to be explicit in the defaults yet making the specification less strict in this regard would make sense to me, especially given that we have the `default` fallback and many modules don't provide much more than that.

james-antill:
I believe our assumption was that you'd do:

   profiles:
     master: [default]
     "2.0": []

...to make it obvious what was happening. But also that it would generally be bad to not have "install @testmodule" work. Saying that I can't think of a reason to not allow Daniel's form.

PR #52, opened 2018-04-18
Drop requirement for profile map to list default stream

sgallagher:
Fixes: https://github.com/fedora-modularity/libmodulemd/issues/43


james-antill:
Just deletes checks and tests for those checks, so fairly easy +1

Issue #54, opened 2018-04-19
build: add `modulemd_dep`


ignatenkobrain:
By using `declare_dependency()` for ease of bundling libmodulemd..

In theory, should be something like

```meson
modulemd_dep = declare_dependency(
    include_directories : include_directories('.'),
    link_with : modulemd_lib,
)
```

ignatenkobrain:
I would use libmodulemd as a subproject() from my project.

sgallagher:
You are talking about adding something like this to the meson.build file? Would this happen at the top-level or in the modulemd subdir?

I'll definitely do this; I just want to make sure I'm doing it correctly.

ignatenkobrain:
> or in the modulemd subdir?

This.

sgallagher:
@ignatenkobrain please try the PR I just submitted and see if it works for you.

PR #56, opened 2018-04-19
Make libmodulemd available as a subproject

sgallagher:
Fixes: https://github.com/fedora-modularity/libmodulemd/issues/54

### DIFF
--- a/modulemd/meson.build
+++ b/modulemd/meson.build
@@ -54,12 +54,19 @@ modulemd_lib = library(
     version: libmodulemd_version,
 )
 
+modulemd_dep = declare_dependency(
+    include_directories : include_directories('.'),
+    link_with : modulemd_lib,
+    dependencies : [
+        gobject,
+    ]
+)
+
 modulemd_validator = executable(
     'modulemd-validator',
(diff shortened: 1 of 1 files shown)


modulemd/meson.build
```diff
@@ -71,6 +71,11 @@ install_headers(
   subdir : header_path,
 )
 
+modulemd_dep = declare_dependency(
+    include_directories : include_directories('.'),
+    link_with : modulemd_lib,
+)
```
ignatenkobrain: it seems you need also add `dependencies: [gobject]`... Btw, also you should be able to change all references later to `dependencies:[modulemd_dep]` instead of `link_with:modulemd_lib,dependencies:[gobject]`

sgallagher: If it needs gobject, it probably needs yaml as well, yes?

ignatenkobrain: you don't expose libyaml in headers, so no.

sgallagher:
I made the changes requested by @ignatenkobrain as well as converting the other dependencies in the file for future maintainability.

Issue #46, opened 2018-04-17
Modulemd: enforce N-E:V-R.A format in artifacts/rpms


dmach:
The DNF team proposes a change in how libmodulemd handles RPM NEVRAs.
We'd like to always get records in N-E:V-R.A format.
libmodulemd SHOULD fail on reading data in invalid format.
libmodulemd MUST fail on writing data in invalid format.

As part of this, we'd like to enforce '0:' epoch when the epoch is zero or empty in RPM.
In some cases missing epoch can represented as "any epoch", which may result in bad UX.

ignatenkobrain:
technically, missing Epoch means epoch=0. I'm quite hesitant to make this change since it's pointless and just increases size of md.

sgallagher:
@dmach This ticket is asking only for the [data.artifacts.rpms](https://github.com/fedora-modularity/libmodulemd/blob/master/spec.v2.yaml#L319) list to be in this format and not for [data.api.rpms](https://github.com/fedora-modularity/libmodulemd/blob/master/spec.v2.yaml#L202) or [data.filter.rpms](https://github.com/fedora-modularity/libmodulemd/blob/master/spec.v2.yaml#L215), correct?

I *can* make such a change, but currently libmodulemd doesn't validate these strings at all. That information comes from the MBS or Pungi (I forget which) and I only validate that it is a string at the moment. If you want that change to happen, it needs to happen in the build-system first. After that, I can make the parser strict.

contyk:
The summary of the ticket only talks about `data.artifacts.rpms`.

The specification already says that this list includes NEVRAs but I'm also unsure whether such validation belongs in libmodulemd -- I'd much rather see a test for this (and other things) in a compose level test of some kind.

And for the record, it's Pungi that fills this in.

sgallagher:
Ah right, the spec *does* say that it's NEVRA and even has `0:` epoch in its examples. If that's required by the spec, libmodulemd probably should be validating it. That being said, I'd prefer not to land such validation until pungi is producing the correct output or else we're going to cause problems.

sgallagher:
@contyk I assume that it must not be, else I can't really see @dmach filing this ticket.

@dmach, can you show me an example of an artifact that isn't using the epoch notation?

PR #58, opened 2018-04-20
Add parsing functions that return failure information

sgallagher:
New Objects:
 * Modulemd.Subdocument

New functions:
 * Modulemd.objects_from_file_ext()
 * Modulemd.objects_from_string_ext()
 * Modulemd.objects_from_stream_ext()
 * Modulemd.Defaults.new_from_file_ext()
 * Modulemd.Defaults.new_from_string_ext()
 * Modulemd.Defaults.new_from_stream_ext()
 * Modulemd.Module.new_from_file_ext()
 * Modulemd.Module.new_from_string_ext()
 * Modulemd.Module.new_from_stream_ext()

These functions will return an array of Modulemd.Subdocument
objects that includes a GError explaining the failure as well as
the YAML string that failed (or at least as much of it as libyaml
was able to read).

Fixes: https://github.com/fedora-modularity/libmodulemd/issues/47

cc @dmach @hanzz 

Signed-off-by: Stephen Gallagher <sgallagh@redhat.com>

PR #59, opened 2018-04-24
Validate the data.artifacts.rpm contents

sgallagher:
These patches add validation that the contents of data.artifacts.rpm read from a modulemd YAML stream are in the prescribed NEVRA format.

In the future it might also validate that the "A" part of the NEVRA is a recognized architecture, but at the moment there is no reliable place to get the list of valid architectures against which to compare.

Fixes: https://github.com/fedora-modularity/libmodulemd/issues/46

### DIFF
--- a/modulemd/modulemd-util.c
+++ b/modulemd/modulemd-util.c
@@ -145,3 +145,100 @@ modulemd_variant_unref (void *ptr)
 {
   g_variant_unref ((GVariant *)ptr);
 }
+
+gboolean
+modulemd_validate_nevra (const gchar *nevra)
+{
+  g_autofree gchar *tmp = g_strdup (nevra);
+  gsize len = strlen (nevra);
+  gchar *i;
+  gchar *endptr;
+
+  /* Since the "name" portion of a NEVRA can have an infinite number of
+   * hyphens, we need to parse from the end backwards.
+   */
+  i = &tmp[len - 1];
+
+  /* Everything after the last '.' must be the architecture */
+  while (i >= tmp)
+    {
+      if (*i == '.')
+        {
+          break;
+        }
+      i--;
+    }
+
+  if (i < tmp)
+    {
+      /* We hit the start of the string without hitting '.' */
+      return FALSE;
+    }
+
+  /*
+   * TODO: Compare the architecture suffix with a list of known-valid ones
+   * This needs to come from an external source that's kept up to date or
+   * this will regularly break.
+   */
+
+  /* Process the "release" tag */
+  while (i > 0)
+    {
+      if (*i == '-')
+        {
+          break;
+        }
+      i--;
+    }
+
+  if (i <= 0)
+    {
+      /* We hit the start of the string without hitting '-' */
+      return FALSE;
+    }
+
+  /* No need to validate Release; it's fairly arbitrary */
+
+  /* Process the version */
+  while (i >= tmp)
+    {
+      if (*i == ':')
+        {
+          break;
+        }
+      i--;
+    }
+  if (i <= 0)
+    {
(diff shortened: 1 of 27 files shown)


ignatenkobrain:
will review showrtly

modulemd/modulemd-util.c
```diff
@@ -145,3 +145,100 @@ modulemd_variant_unref (void *ptr)
...
+  gsize len = strlen (nevra);
+  gchar *i;
+  gchar *endptr;
+
+  /* Since the "name" portion of a NEVRA can have an infinite number of
+   * hyphens, we need to parse from the end backwards.
+   */
+  i = &tmp[len - 1];
+
+  /* Everything after the last '.' must be the architecture */
```
ignatenkobrain: strrchr? or there probably some glib version

modulemd/modulemd-util.c
```diff
@@ -145,3 +145,100 @@ modulemd_variant_unref (void *ptr)
...
+    }
+
+  /*
+   * TODO: Compare the architecture suffix with a list of known-valid ones
+   * This needs to come from an external source that's kept up to date or
+   * this will regularly break.
+   */
+
+  /* Process the "release" tag */
+  while (i > 0)
```
ignatenkobrain: same

modulemd/modulemd-util.c
```diff
@@ -145,3 +145,100 @@ modulemd_variant_unref (void *ptr)
...
+        }
+      i--;
+    }
+  if (i <= 0)
+    {
+      /* We hit the start of the string without hitting ':' */
+      return FALSE;
+    }
+
+  /* Process the epoch */
```
ignatenkobrain: epoch is delemited by `:`

modulemd/test-modulemd-yaml.c
```diff
@@ -414,6 +415,55 @@ modulemd_yaml_test_v2_stream (YamlFixture *fixture, gconstpointer user_data)
...
 
+static void
+modulemd_yaml_test_validate_nevra (YamlFixture *fixture,
+                                   gconstpointer user_data)
+{
+  const gchar *good = "nodejs-devel-1:8.10.0-3.module_1572+d7ec111e.x86_64";
+  const gchar *garbage = "DEADBEEF";
+  const gchar *garbage2 = "DEAD.BEEF";
+  const gchar *garbage3 = "MORE-DEAD.BEEF";
+  const gchar *missing_epoch =
```
ignatenkobrain: why is it important? Epoch is entirely optional and `rpm` and `dnf` do not output it by default in UI.

sgallagher: We're being very specific in the format we expect here to simplify parsing expectations. Pungi always includes the 0:, so we're going to validate that if only to ensure consistency when third-parties start producing content.

sgallagher: See https://github.com/fedora-modularity/libmodulemd/issues/46

Consumers of libmodulemd want a simpler string to parse. Pungi already guarantees that the `0:` is included, so this is mostly just insurance for future third-parties to make sure they're producing the same content.

sgallagher:
Thanks to CI, I realized I actually have to have these patches atop the ones from https://github.com/fedora-modularity/libmodulemd/pull/58 so I rebased it. Only the last three in the set are unique to this PR.

modulemd/modulemd-util.c
sgallagher: Yes, and? Did you see some mistake here?

modulemd/modulemd-util.c
sgallagher: I don't really see how that's more advantageous.

PR #62, opened 2018-05-04
Validator: Better verbosity options

sgallagher:
Default operation will display error information. It can be
suppressed with --quiet. Debug verbosity will now display debugging
data from libmodulemd.


sgallagher:
@ignatenkobrain I fixed those two issues and pushed them as fixup commits. If you approve them, I'll squash and merge them.

PR #63, opened 2018-05-24
Fix license blurbs

nphilipp:
This clarifies the licenses of the files by replacing an old version of the MIT license text which refers to the X Consortium to a generic version of the text. Additionally, it adds a license blurb to the python test file.

### DIFF
--- a/modulemd/modulemd-common.c
+++ b/modulemd/modulemd-common.c
@@ -10,13 +10,13 @@
  * permit persons to whom the Software is furnished to do so, subject to
  * the following conditions:
  *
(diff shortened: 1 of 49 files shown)


ignatenkobrain:
Can we just put `/* SPDX-License-Identifier: MIT */` instead of those licenses?

ignatenkobrain:
Kernel and systemd has done this for instance.

sgallagher:
Sorry, @ignatenkobrain, I didn't see your reply until after I merged the PR. I'll look into that; I haven't seen that before.

PR #69, opened 2018-06-15
Specification: Allow overriding the default component whitelist

contyk:
Some packages rename the SRPM and therefore the package build name
during the SRPM build phase.  Module Build Service takes care of package
whitelisting in the tags in manages and therefore it needs to know the
names of the builds when it's setting up the build environment.

This list provides the necessary information to the build system if the
build names do not match the component names.

### DIFF
--- a/spec.v2.yaml
+++ b/spec.v2.yaml
@@ -229,6 +229,18 @@ data:
             macros: |
                 %demomacro 1
                 %demomacro2 %{demomacro}23
+            # Explicit list of package build names this module will produce.
+            # By default the build system only allows components listed under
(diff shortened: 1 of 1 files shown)


sgallagher:
Can you include a comment describing what is the acceptable format of the whitelist? Does it need to be NEVRA? ENVRA? Just N? Is it the SRPM name or the built RPM names?

contyk:
It's just name. I'll update the PR.

sgallagher:
This is failing the tests because the spec no longer validates successfully with the parser (which is expected). I'll pull this patch into my branch and get the parser updated, then resubmit them together.

PR #78, opened 2018-08-09
Fix API-breakage for Modulemd.Module.new_all_from_*_ext()

sgallagher:
These functions were not properly converting ModuleStream to Module for backwards-compatibility.
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1607083

PR #79, opened 2018-08-09
Don't return ModuleStream objects from modulemd_module_new_all_*_ext

owtaylor:
These older, deprecated functions should not have been changed to
return ModuleStream objects rather than Module objects.


sgallagher:
This is very similar to https://github.com/fedora-modularity/libmodulemd/pull/78

I think I should probably add the `if (data)` checks that you have to that PR, though. Okay to close this PR as a dupe, or is there another aspect of it you want me to include?

sgallagher:
Hmm, actually I do like that you renamed the conversion function when you moved it, so maybe I'll abandon #78 and merge this instead.

owtaylor:
Sorry for missing that you already had a PR for this (I wasn't expecting you to work on this on a day you had four talks! ;-) Feel free to merge whichever one you like better... the differences are entirely stylistic.

PR #81, opened 2018-08-27
Add support for module-level provides

contyk:
This should be a backwards-compatible change.

Define module-level provides for each of the `dependencies` branches.
Provides are used for module-level dependency resolution at both build
time and runtime.

### DIFF
--- a/spec.v2.yaml
+++ b/spec.v2.yaml
@@ -90,22 +90,28 @@ data:
     # the runtime block so that it matches the stream the module was built
     # against.  Multiple builds result in multiple output modulemd files.
     # See below for an example.
-    # TODO: Provides, conflicts, obsoletes, recommends, etc.
-    #       Do we even need those?
+    # Provides also support stream expansion and indicate what modules names
+    # and streams are provided.  Just like with dependencies, streams can
+    # be inclusive or exclusive lists, with an empty list meaning that any
+    # stream of that module name is provided.
     # The example below illustrates how to build the same module in four
     # different ways, with varying build time and runtime dependencies.
     dependencies:
         # Build on all available platforms except for f27, f28 and epel7
         # After build, the runtime dependency will match the one used for
         # the build.
+        # In this situation this module also provides "onestream:1".
         - buildrequires:
               platform: [-f27, -f28, -epel7]
           requires:
               platform: [-f27, -f28, -epel7]
+          provides:
+              onestream: [v1]
         # For platform:f27 perform two builds, one with buildtools:v1, another
         # with buildtools:v2 in the buildroot.  Both will also utilize
         # compatible:v3.  At runtime, buildtools isn't required and either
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
so what about conflicts/supplements/recommends/enhances/suggests? ;)

spec.v2.yaml
```diff
@@ -133,6 +146,8 @@ data:
               platform: [epel7]
               extras: []
               moreextras: [foo, bar]
+          provides:
+              exceptstreams: [~v3, ~v4]
```
ignatenkobrain: why not `-`?

spec.v2.yaml
```diff
@@ -90,41 +90,54 @@ data:
...
         # Build on all available platforms except for f27, f28 and epel7
         # After build, the runtime dependency will match the one used for
         # the build.
+        # In this situation this module also provides "onestream:1".
         - buildrequires:
               platform: [-f27, -f28, -epel7]
           requires:
               platform: [-f27, -f28, -epel7]
+          provides:
+              onestream: [v1]
```
ignatenkobrain: why do we need multiple keywords for same thing?

spec.v2.yaml
contyk: A bug! Thanks for catching this.

spec.v2.yaml
contyk: What do you mean?

contyk:
There was no need for provides until this point either; I'd rather keep it simple, so unless we have a geuine reason to add conflicts and weak deps, I'm not going to.

spec.v2.yaml
```diff
@@ -90,41 +90,54 @@ data:
     # the runtime block so that it matches the stream the module was built
     # against.  Multiple builds result in multiple output modulemd files.
     # See below for an example.
-    # TODO: Provides, conflicts, obsoletes, recommends, etc.
-    #       Do we even need those?
+    # Provides also support stream expansion and indicate what modules names
+    # and streams are provided.  Just like with dependencies, streams can
+    # be inclusive or exclusive lists, with an empty list meaning that any
+    # stream of that module name is provided.
```
sgallagher: Could you add some example of why exactly this is desirable? I have some guesses, but I'd really prefer that this was stated explicitly in the specification.

sgallagher:
Please notice that the automated tests have failed, because you changed the specification without changing the implementation to match it. Do not merge this until the tests pass.

spec.v2.yaml
sgallagher: Both, please.

From our conversation elsewhere, it seems that this is mainly going to be useful when there are multiple platform branches that should all be treated as Providing the same thing (such as if we decided to do a Fedora 29.1 and 29.2 that should both satisfy Requires: platform:29.

contyk:
We discussed this a couple of times last week and we think we don't really need this after all.
Closing in order to keep things simple.

PR #87, opened 2018-09-14
Proposal for modulemd-defaults v2

sgallagher:
While working on libmodulemd 2.0, it occurred to me that the implementation of intents in the modulemd-defaults document is extremely redundant (its subheading contains all the same information as the main document except for the module name). I think it would be more reasonable to simply make the intent name a field in the main portion and then just have separate modulemd-defaults documents for each of them.

What do you think, @contyk ?

### DIFF
--- a/mod-defaults/spec.v2.yaml
+++ b/mod-defaults/spec.v2.yaml
@@ -1,0 +1,24 @@
+# Document type identifier
+document: modulemd-defaults
+
+# Module metadata format version
+version: 2
+
+data:
+    # Module name that the defaults are for, required.
+    module: foo
+
+    # Module stream that is the default for the module, optional.
+    stream: x.y
+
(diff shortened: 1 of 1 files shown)


contyk:
I don't really have a problem with that.

However, before we move forward here, we need to update our defaults filesystem naming policy in `fedora-module-defaults` as well as any CI checks that enforce it.

Also, I don't suppose this affects defaults merging much if at all, but still. Is it somewhat compatible with the 1.0 libmodulemd API?

sgallagher:
I think I will scrap this idea. It's closer to how I'm representing the data inside the libmodulemd 2.0 API, but I can support either version of the YAML in that way, so changing it probably isn't worthwhile. Adding support to the 1.0 API for this would be more trouble than it's worth, since we want to get away from that implementation anyway.

PR #89, opened 2018-09-18
Deduplicate module streams when merging

sgallagher:
This is a fairly hacky way of deduplicating module streams with the same NSVC that come from different sources (repos). It makes a weak assumption that if the NSVC is the same, the whole stream is the same.

A better long-term fix will be to have a full module stream comparison function, but that will be added to the 2.0 API plans.

### DIFF
--- a/modulemd/v1/tests/test-modulemd-python.py
+++ b/modulemd/v1/tests/test-modulemd-python.py
@@ -175,7 +175,6 @@ class TestIssues(unittest.TestCase):
         assert mmd.peek_rpm_buildopts() != {}
         assert mmd.peek_rpm_buildopts()['macros'] == '%my_macro 1'
         dumped = mmd.dumps()
-        print("YAML:\n%s" % dumped, file=sys.stderr)
         mmd2 = Modulemd.Module.new_from_string(dumped)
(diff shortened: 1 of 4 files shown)


modulemd/v1/tests/test-modulemd-python.py
```diff
@@ -266,6 +265,36 @@ def test_issue85(self):
...
+            '%s/test_data/issue88.yaml' % os.getenv('MESON_SOURCE_ROOT'))
+
+        # Test at the same priority level
+        prioritizer = Modulemd.Prioritizer()
+        prioritizer.add(objects_from_repo_a, 0)
+        prioritizer.add(objects_from_repo_b, 0)
+        supposedly_merged_objects = prioritizer.resolve()
+
+        # I would expect all three of these to be the same length... but they
+        # are not.
```
ralphbean: Maybe adjust this comment and the one below it?

:+1: to including the reproducer almost verbatim as a test.  :)

sgallagher:
I'm pretty much the sole developer, so I'm going to go ahead and run with this since it fixes your issue and doesn't break any of my other merging tests.

PR #90, opened 2018-09-21
improvements for error messages

nphilipp:
As discussed (and long due), here are changes to the parser which:

1. allow printf-style messages in parsing macros and
2. report line/column numbers on errors

As previously discussed, I've implemented v2 linting in `fedmod`, so removed all changes I wrote for linting in `modulemd-validator`. Therefore the patch for printf-style messages is currently unused but I left it in anyway because it's userful on its own.

### DIFF
--- a/modulemd/include/modulemd-1.0/private/modulemd-yaml.h
+++ b/modulemd/include/modulemd-1.0/private/modulemd-yaml.h
@@ -84,6 +84,19 @@ typedef gboolean (*ModulemdParsingFunc) (yaml_parser_t *parser,
     }                                                                         \
   while (0)
 
+#define MMD_YAML_ERROR_EVENT_RETURN_RETHROW(_error, event, ...)               \
+  do                                                                          \
+    {                                                                         \
+      gchar *formatted = g_strdup_printf (__VA_ARGS__);                       \
+      g_debug ("%s [line %zu col %zu]",                                       \
+               formatted,                                                     \
+               event.start_mark.line,                                         \
+               event.start_mark.column);                                      \
+      g_free (formatted);                                                     \
+      goto error;                                                             \
+    }                                                                         \
+  while (0)
+
 #define MMD_ERROR_RETURN_FULL(_error, type, ...)                              \
   do                                                                          \
     {                                                                         \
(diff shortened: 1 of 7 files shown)


nphilipp:
It seems as if I need to update some tests...

modulemd/include/modulemd-1.0/private/modulemd-yaml.h
```diff
@@ -75,46 +75,50 @@ typedef gboolean (*ModulemdParsingFunc) (yaml_parser_t *parser,
...
     }                                                                         \
   while (0)
 
-#define YAML_EMITTER_EMIT_WITH_ERROR_RETURN(emitter, event, _error, msg)      \
+#define YAML_EMITTER_EMIT_WITH_ERROR_RETURN(emitter, event, _error, ...)      \
   do                                                                          \
     {                                                                         \
       if (!yaml_emitter_emit (emitter, event))                                \
         {                                                                     \
+          gchar *formatted = g_strdup_printf (__VA_ARGS__);                   \
```
sgallagher: Please use `g_autofree gchar *__formatted = g_strdup_printf (__VA_ARGS__);` here. I'm trying to move towards using the GLib auto pointers wherever possible, that way we don't end up leaking if we add an error return somewhere.

modulemd/include/modulemd-1.0/private/modulemd-yaml.h
```diff
@@ -75,46 +75,50 @@ typedef gboolean (*ModulemdParsingFunc) (yaml_parser_t *parser,
...
-                   msg,                                                       \
+                   formatted,                                                 \
                    mmd_yaml_get_event_name ((event)->type));                  \
-          g_set_error_literal (                                               \
-            _error, MODULEMD_YAML_ERROR, MODULEMD_YAML_ERROR_EMIT, msg);      \
+          g_set_error_literal (_error,                                        \
+                               MODULEMD_YAML_ERROR,                           \
+                               MODULEMD_YAML_ERROR_EMIT,                      \
+                               formatted);                                    \
+          g_free (formatted);                                                 \
```
sgallagher: If you make the `g_autofree` change above, this line becomes unnecessary.

nphilipp: I wasn't aware of `g_autofree()`, neat trick...

modulemd/include/modulemd-1.0/private/modulemd-yaml.h
nphilipp: The added commit 8cdf3ad2e27145079c07d61d0c3ca1b382975857 uses `g_autofree` rather than freeing the pointers manually.

nphilipp:
Hmm, formatting test fails in CI but I can't reproduce it here, hang on...

modulemd/include/modulemd-1.0/private/modulemd-yaml.h
nphilipp: And commit 644b944 should get the formatting right. :wink:

PR #91, opened 2018-09-27
offload linting documents to fedmod (if available)

nphilipp:
This patch just uses 'fedmod lint' if available and falls back to the old behavior (displaying the YAML document) if not. There's no way for the user to force the old behavior from the command line. Let me know if this needs to be an option.

### DIFF
--- a/modulemd/v1/modulemd-validator.c
+++ b/modulemd/v1/modulemd-validator.c
@@ -11,11 +11,15 @@
  * For more information on free software, see <https://www.gnu.org/philosophy/free-sw.en.html>.
  */
 
+#define _POSIX_SOURCE  /* for fdopen() */
+
 #include "modulemd.h"
 #include "private/modulemd-yaml.h"
 
 #include <glib.h>
+#include <glib/gstdio.h>
 #include <locale.h>
+#include <stdio.h>
 
 /* TODO: implement internationalization */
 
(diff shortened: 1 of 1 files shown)


modulemd/v1/modulemd-validator.c
```diff
@@ -178,10 +281,7 @@ main (int argc, char *argv[])
               for (gsize i = 0; i < failures->len; i++)
                 {
                   doc = (ModulemdSubdocument *)g_ptr_array_index (failures, i);
-                  fprintf (stdout,
-                           "\nFailed subdocument (%s): \n%s\n",
-                           modulemd_subdocument_get_gerror (doc)->message,
-                           modulemd_subdocument_get_yaml (doc));
+                  summarize_failures (doc, fedmod);
```
sgallagher: So, `fedmod lint` only runs on the set of YAML subdocuments that failed parsing? I'm not sure I see the value in adding it to this at all, then.

It seems to me that `fedmod lint` is more useful when run on its own against the complete YAML document, not just those that already failed the simple parser.

sgallagher:
Yeah, I’m not sure I see any value to putting this into libmodulemd. I could see an argument for dropping the modulemd-validator entirely in favor of the linter though.

nphilipp:
@sgallagher Heh, this PR implements things we discussed a couple of weeks ago, but replacing the validator with fedmod lint completely is fine with me. Need to check if it handles multi-document files well before that, though.

PR #98, opened 2018-10-26
Run tests against installed libraries, too

sgallagher:
Many changes to enable running the unit tests against a library installed on the local system (rather than the in-tree build).

Also extends the Travis test suite to install a set of built RPMs and run the tests against them.

This pull request is mainly to validate that the Travis suite works properly in production.


sgallagher:
It is a separate script. Mason just handles the template for that script.

Issue #99, opened 2018-10-31
No such file or directory: '/builddir/build/BUILD/modulemd-1.7.0/spec_tmpl.sh


tpgxyz:
latest version 1.7.0 does not build
More logs can be found here https://abf.openmandriva.org/build_lists/264186

```
ninja: Entering directory `build'
[1/64] Generating rpmsetup with a custom command.
[2/64] Compiling C object 'modulemd/v1/modulemd@v1@@modulemd@sha/modulemd-component-module.c.o'.
[3/64] Compiling C object 'modulemd/v1/modulemd@v1@@modulemd@sha/modulemd-buildopts.c.o'.
[4/64] Compiling C object 'modulemd/v1/modulemd@v1@@modulemd@sha/modulemd-common.c.o'.
[5/64] Compiling C object 'modulemd/v1/modulemd@v1@@modulemd@sha/modulemd-dependencies.c.o'.
../modulemd/v1/modulemd-dependencies.c:77:3: warning: passing 'typeof (*(&streams[0]))' (aka 'const char *') to parameter of type 'gpointer' (aka 'void *') discards qualifiers [-Wincompatible-pointer-types-discards-qualifiers]
  g_clear_pointer (&streams[0], g_free);
  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/glib-2.0/glib/gmem.h:121:18: note: expanded from macro 'g_clear_pointer'
      (destroy) (_ptr);                                                        \
                 ^~~~
/usr/include/glib-2.0/glib/gmem.h:71:35: note: passing argument to parameter 'mem' here
void     g_free           (gpointer      mem);
                                         ^
1 warning generated.
[6/64] Compiling C object 'modulemd/v1/modulemd@v1@@modulemd@sha/modulemd-component.c.o'.
[7/64] Generating specfile with a meson_exe.py custom command.
FAILED: libmodulemd.spec 
/usr/bin/meson --internal exe /builddir/build/BUILD/modulemd-1.7.0/build/meson-private/meson_exe_spec_tmpl.sh_bb026530aab33876f2f2aa904ab53932e74bbde4.dat
Traceback (most recent call last):
  File "/usr/bin/meson", line 11, in <module>
    load_entry_point('meson==0.48.1', 'console_scripts', 'meson')()
  File "/usr/lib/python3.7/site-packages/mesonbuild/mesonmain.py", line 392, in main
[...]

PR #100, opened 2018-10-31
Handle sh interpreter in other locations than /usr/bin/sh

sgallagher:
Fixes: https://github.com/fedora-modularity/libmodulemd/issues/99

@tpgxyz Can you verify that this patch fixes your issue, please?

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -51,6 +51,9 @@ gobject = dependency('gobject-2.0')
 yaml = dependency('yaml-0.1')
 gtkdoc = dependency('gtk-doc')
 
+sh = find_program('sh')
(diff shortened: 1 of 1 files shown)


Conan-Kudo:
@sgallagher It looked good to me, but the main thing I'd be concerned about is if your scripts are actually POSIX shell compatible. If they rely on so-called "bashisms", then you may want to locate bash specifically.

sgallagher:
@Conan-Kudo No bashisms as far as I know. It's basically just "set two variables, call one executable".

Conan-Kudo:
Then you're fine. :)

PR #103, opened 2018-11-06
modulemd-yaml-parser: error is not set though it's expected

ret2libc:
When parsing the YAML file in _parse_yaml, if an unexpected event is
received, no error is set by YAML_PARSER_PARSE_WITH_ERROR_RETURN (if it
was, it would have already goto "error" label) but
MMD_YAML_ERROR_EVENT_RETURN_RETHROW was used. This means in some cases
`error` could be NULL, though the caller expect it to be set when the
result of the function is false.

### DIFF
--- a/modulemd/v1/modulemd-yaml-parser.c
+++ b/modulemd/v1/modulemd-yaml-parser.c
@@ -434,8 +434,8 @@ _parse_yaml (yaml_parser_t *parser,
 
         default:
           /* We received a YAML event we shouldn't expect at this level */
-          MMD_YAML_ERROR_EVENT_RETURN_RETHROW (
-            error, event, "Unexpected YAML event during preprocessing");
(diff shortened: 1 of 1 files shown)


sgallagher:
Good catch! Only one small nitpick: would you mind running `ninja test` before submitting? Among other things, it runs the automatic code formatter `clang-format` on the sources so they remain in the same style. You'll notice that the CI build failed with the message "Autoformatter was not run before submitting. Please run `ninja test`, amend the commit and resubmit this pull request."

I'll merge it once that is done. Thanks!

ret2libc:
Ah ok, sorry for that :)

sgallagher:
It's kind of a sneaky way for me to see which pull-requests actually had the test suite run against them as well :)

ret2libc:
You caught me!

PR #105, opened 2018-11-06
yaml-parser: fix multiple uses of uninitialized variables

ret2libc:
when an error condition is met in parse_yaml_file, parse_yaml_string,
parse_yaml_stream at the very beginning of those functions, the parser
is deleted without having been initialized. This makes those functions
access uninitialized memory, possibly causing crashes or other effects.

### DIFF
--- a/modulemd/v1/modulemd-yaml-parser.c
+++ b/modulemd/v1/modulemd-yaml-parser.c
@@ -55,19 +55,10 @@ parse_yaml_file (const gchar *path,
   FILE *yaml_file = NULL;
   yaml_parser_t parser;
 
+  g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+  g_return_val_if_fail (path, FALSE);
+
   g_debug ("TRACE: entering parse_yaml_file");
-
-  if (error != NULL && *error != NULL)
-    {
-      MMD_ERROR_RETURN_FULL (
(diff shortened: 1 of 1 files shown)


ret2libc:
Maybe I didn't explain well. By error condition I meanly meant `path == NULL` or `yaml == NULL` or `stream == NULL`. Do you want to treat those checks as programming error as well?

sgallagher:
Looks like I introduced that mess back in March as https://github.com/fedora-modularity/libmodulemd/commit/211ba6b7792a1661bb9076fd599d717ac6efd942

That was not the right way to handle that case.

sgallagher:
Yeah. I don't remember why I created `MODULEMD_YAML_ERROR_PROGRAMMING`, but it was an incorrect decision. Programming mistakes (which these cases would be), should use `g_return_val_if_fail()` rather than providing data to the `error` variable.

It's been eight months and this was part of a huge rewrite of this section of code, so I was probably not thinking clearly by the end of it.

ret2libc:
So I will handle both things (error and path/stream/yaml checks) with `g_return_val_if_fail`. Does that sound ok?

sgallagher:
Yes, I think that's a more-correct approach to the problem. It will also be more immediately obvious if any other code in libmodulemd is calling those functions incorrectly.

PR #108, opened 2018-11-15
Implemented translation-entry for modulemdv2

puiterwijk:
Signed-off-by: Patrick Uiterwijk <patrick@puiterwijk.org>

### DIFF
--- a/modulemd/meson.build
+++ b/modulemd/meson.build
@@ -118,23 +118,27 @@ test_v1_srcs = files(
 
 modulemd_v2_srcs = files(
     'v2/modulemd-service-level.c',
+    'v2/modulemd-translation-entry.c',
     'v2/modulemd-util.c',
     'v2/modulemd-yaml-util.c',
 )
 
 modulemd_v2_hdrs = files(
     'v2/include/modulemd-2.0/modulemd-service-level.h',
+    'v2/include/modulemd-2.0/modulemd-translation-entry.h',
 )
 
 modulemd_v2_priv_hdrs = files(
     'v2/include/modulemd-2.0/private/glib-extensions.h',
     'v2/include/modulemd-2.0/private/modulemd-service-level-private.h',
+    'v2/include/modulemd-2.0/private/modulemd-translation-entry-private.h',
     'v2/include/modulemd-2.0/private/modulemd-util.h',
     'v2/include/modulemd-2.0/private/modulemd-yaml.h',
 )
 
 test_v2_srcs = files(
     'v2/tests/test-modulemd-service-level.c',
+    'v2/tests/test-modulemd-translation-entry.c',
     'v2/tests/test-utils.c',
 )
 
@@ -143,6 +147,8 @@ test_v2_priv_hdrs = files(
 )
 
 test_v2_python_scripts = files(
+    'v2/tests/ModulemdTests/base.py',
+    'v2/tests/ModulemdTests/translationentry.py',
     'v2/tests/ModulemdTests/servicelevel.py',
 )
 
@@ -218,6 +224,7 @@ if build_api_v2
           env : test_env,
           args : modulemd_v2_hdrs)
     valgrind_tests = valgrind_tests + [
+        'translation_entry_v2_debug',
         'service_level_v2_debug',
     ]
 endif
--- a/modulemd/v2/include/modulemd-2.0/private/modulemd-translation-entry-private.h
+++ b/modulemd/v2/include/modulemd-2.0/private/modulemd-translation-entry-private.h
@@ -1,0 +1,61 @@
+/*
+ * This file is part of libmodulemd
+ * Copyright (C) 2018 Red Hat, Inc.
+ *
+ * Fedora-License-Identifier: MIT
+ * SPDX-2.0-License-Identifier: MIT
+ * SPDX-3.0-License-Identifier: MIT
+ *
+ * This program is free software.
+ * For more information on the license, see COPYING.
+ * For more information on free software, see <https://www.gnu.org/philosophy/free-sw.en.html>.
+ */
+
+#pragma once
+
+#include <glib-object.h>
+#include <yaml.h>
+
+#include "modulemd-translation-entry.h"
+
+/**
+ * SECTION: modulemd-translation-entry-private
+ * @title: Modulemd.TranslationEntry (Private)
+ * @stability: Private
+ * @short_description: #Modulemd.TranslationEntry methods that should be used only
+ * by internal consumers.
+ */
+
+/**
+ * modulemd_translation_entry_parse_yaml:
+ * @parser: (inout): A libyaml parser object positioned at the beginning of a
+ * Translation Entry entry in the YAML document.
+ * @error: (out): A #GError that will return the reason for a parsing or
+ * validation error.
+ *
+ * Returns: (transfer full): A newly-allocated #ModulemdTranslationEntry object
+ * read from the YAML. NULL if a parse or validation error occurred and sets
+ * @error appropriately.
+ *
+ * Since: 2.0
+ */
+ModulemdTranslationEntry *
+modulemd_translation_entry_parse_yaml (yaml_parser_t *parser, GError **error);
+
+/**
+ * modulemd_translation_entry_emitter_yaml:
+ * @self: This #ModulemdTranslationEntry
+ * @emitter: (inout): A libyaml emitter object positioned where a Translation Entry
+ * belongs in the YAML document.
+ * @error: (out): A #GError that will return the reason for an emission or
+ * validation error.
+ *
+ * Returns: TRUE if the translation entry was emitted successfully. FALSE and sets
+ * @error appropriately if the YAML could not be emitted.
+ *
+ * Since: 2.0
+ */
+gboolean
+modulemd_translation_entry_emit_yaml (ModulemdTranslationEntry *self,
+                                      yaml_emitter_t *emitter,
+                                      GError **error);
--- a/modulemd/v2/include/modulemd-2.0/private/modulemd-util.h
+++ b/modulemd/v2/include/modulemd-2.0/private/modulemd-util.h
@@ -49,3 +49,16 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (modulemd_tracer, modulemd_trace_free);
   while (0)
 
 G_END_DECLS
+
+
+GHashTable *
+modulemd_hash_table_deep_str_copy (GHashTable *orig);
+
+gint
+modulemd_strcmp_sort (gconstpointer a, gconstpointer b);
+
+GPtrArray *
+modulemd_ordered_str_keys (GHashTable *htable, GCompareFunc compare_func);
+
+gchar **
+modulemd_ordered_str_keys_as_strv (GHashTable *htable);
--- a/modulemd/v2/include/modulemd-2.0/private/modulemd-yaml.h
+++ b/modulemd/v2/include/modulemd-2.0/private/modulemd-yaml.h
@@ -201,4 +201,18 @@ GDate *
 modulemd_yaml_parse_date (yaml_parser_t *parser, GError **error);
 
 
+/**
+ * modulemd_yaml_parse_string:
+ * @parser: (inout): A libyaml parser object positioned at the beginning of a string scalar entry.
+ * @error: (out): A #GError that will return the reason for a parsing or validation error.
+ *
+ * Returns: (transfer full): A newly-allocated gchar * representing the parsed value.
+ * NULL if a parse error occured and sets @error appropriately.
+ *
+ * Since: 2.0
+ */
+gchar *
+modulemd_yaml_parse_string (yaml_parser_t *parser, GError **error);
+
+
 G_END_DECLS
--- a/modulemd/v2/modulemd-translation-entry.c
+++ b/modulemd/v2/modulemd-translation-entry.c
@@ -1,0 +1,647 @@
+/*
+ * This file is part of libmodulemd
+ * Copyright (C) 2018 Red Hat, Inc.
+ *
+ * Fedora-License-Identifier: MIT
+ * SPDX-2.0-License-Identifier: MIT
+ * SPDX-3.0-License-Identifier: MIT
+ *
+ * This program is free software.
+ * For more information on the license, see COPYING.
+ * For more information on free software, see <https://www.gnu.org/philosophy/free-sw.en.html>.
+ */
+
+#include <glib.h>
+#include <yaml.h>
+
+#include "modulemd-translation-entry.h"
+#include "private/glib-extensions.h"
+#include "private/modulemd-translation-entry-private.h"
+#include "private/modulemd-util.h"
+#include "private/modulemd-yaml.h"
+
+#define TE_DEFAULT_STRING "__LOCALE_UNSET__"
+
+struct _ModulemdTranslationEntry
+{
+  GObject parent_instance;
+
+  gchar *locale;
+  gchar *summary;
+  gchar *description;
+
+  GHashTable *profile_descriptions;
+};
+
+G_DEFINE_TYPE (ModulemdTranslationEntry,
+               modulemd_translation_entry,
+               G_TYPE_OBJECT)
+
+enum
+{
+  PROP_0,
+
+  PROP_LOCALE,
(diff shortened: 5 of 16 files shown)


modulemd/v2/tests/ModulemdTests/servicelevel.py
```diff
@@ -68,6 +68,7 @@ def test_constructors(self):
         # Test that we fail if we call new() with a None name
         try:
             sl = Modulemd.ServiceLevel.new(None)
+            assert False
```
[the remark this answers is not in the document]
puiterwijk: Because otherwise this code was also proceeding an error was not thrown.

modulemd/v2/include/modulemd-2.0/private/modulemd-translation-entry-private.h
```diff
@@ -0,0 +1,71 @@
...
+ * read from the YAML. NULL if a parse or validation error occurred and sets
+ * @error appropriately.
+ *
+ * Since: 2.0
+ */
+ModulemdTranslationEntry *
+modulemd_translation_entry_parse_yaml (yaml_parser_t *parser, GError **error);
+
+static void *
+modulemd_translation_entry_parse_yaml_profiles (yaml_parser_t *parser,
```
sgallagher: This is an internal implementation detail of the parser for the TranslationEntry. Do you see any reason why another caller in libmodulemd would need to access this directly?

I think this should probably be dropped from the header and just implemented as a static function in the C file.

modulemd/v2/include/modulemd-2.0/private/modulemd-translation-entry-private.h
```diff
@@ -0,0 +1,71 @@
...
+ *
+ * Since: 2.0
+ */
+gboolean
+modulemd_translation_entry_emit_yaml (ModulemdTranslationEntry *self,
+                                      yaml_emitter_t *emitter,
+                                      GError **error);
+
+gboolean
+modulemd_translation_entry_emit_yaml_profiles (ModulemdTranslationEntry *self,
```
sgallagher: Same comment as I had for `modulemd_translation_entry_parse_yaml_profiles()`

modulemd/v2/include/modulemd-2.0/private/modulemd-util.h
```diff
@@ -49,3 +49,13 @@ G_DEFINE_AUTOPTR_CLEANUP_FUNC (modulemd_tracer, modulemd_trace_free);
   while (0)
 
 G_END_DECLS
+
+
+GHashTable *
+_modulemd_hash_table_deep_str_copy (GHashTable *orig);
```
sgallagher: Let's drop the leading underscores. I don't know why I thought that was a good idea in v1, but I don't anymore :)

modulemd/v2/include/modulemd-2.0/private/modulemd-yaml.h
```diff
@@ -201,4 +201,8 @@ GDate *
 modulemd_yaml_parse_date (yaml_parser_t *parser, GError **error);
 
 
+gchar *
+modulemd_yaml_parse_string (yaml_parser_t *parser, GError **error);
```
sgallagher: Please add gtkdoc annotations for this function.

modulemd/meson.build
```diff
@@ -118,23 +118,27 @@ test_v1_srcs = files(
...
     'v2/include/modulemd-2.0/private/glib-extensions.h',
     'v2/include/modulemd-2.0/private/modulemd-service-level-private.h',
+    'v2/include/modulemd-2.0/private/modulemd-translation-entry-private.h',
     'v2/include/modulemd-2.0/private/modulemd-util.h',
     'v2/include/modulemd-2.0/private/modulemd-yaml.h',
 )
 
 test_v2_srcs = files(
     'v2/tests/test-modulemd-service-level.c',
+    'v2/tests/test-modulemd-translation-entry.c',
```
sgallagher: You didn't add your python tests to the test_v2_python_scripts below.

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+
+static void
+modulemd_translation_entry_finalize (GObject *object)
+{
+  ModulemdTranslationEntry *self = (ModulemdTranslationEntry *)object;
+
+  g_clear_pointer (&self->locale, g_free);
+  g_clear_pointer (&self->summary, g_free);
+  g_clear_pointer (&self->description, g_free);
+
```
sgallagher: You aren't freeing the profiles here. That's a memory leak. The valgrind test should have caught that... I wonder why it did not.

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+modulemd_translation_entry_get_profiles_as_strv (
+  ModulemdTranslationEntry *self)
+{
+  GPtrArray *keys = _modulemd_ordered_str_keys (self->profile_descriptions,
+                                                _modulemd_strcmp_sort);
+  // Add the NULL sentinel
+  g_ptr_array_add (keys, NULL);
+  // Store the pdata for returning after we free the container
+  gchar **result = (gchar **)keys->pdata;
+  g_ptr_array_free (keys, FALSE);
```
sgallagher: Could you make this conversion into a utility function, please? It's probably going to be frequently handy.

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+}
+
+
+void
+modulemd_translation_entry_set_profile_description (
+  ModulemdTranslationEntry *self,
+  const gchar *profile_name,
+  const gchar *profile_description)
+{
+  g_hash_table_insert (self->profile_descriptions,
```
sgallagher: Please use `g_hash_table_replace()` here, it's clearer about how it handles the memory of the previous key if it already exists.

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+}
+
+ModulemdTranslationEntry *
+modulemd_translation_entry_parse_yaml (yaml_parser_t *parser, GError **error)
+{
+  MODULEMD_INIT_TRACE ();
+  MMD_INIT_YAML_EVENT (event);
+  gboolean done = FALSE;
+  gboolean in_map = FALSE;
+  gchar *value;
```
sgallagher:
```suggestion
  g_autofree gchar *value;
```

This way if you ever get an error before the manual `free()`ing it will still get cleaned up.

modulemd/v2/modulemd-yaml-util.c
```diff
@@ -327,3 +327,18 @@ modulemd_yaml_parse_date (yaml_parser_t *parser, GError **error)
...
+modulemd_yaml_parse_string (yaml_parser_t *parser, GError **error)
+{
+  MMD_INIT_YAML_EVENT (event);
+
+  YAML_PARSER_PARSE_WITH_EXIT (parser, &event, error);
+  if (event.type != YAML_SCALAR_EVENT)
+    {
+      MMD_YAML_ERROR_EVENT_EXIT (error, event, "String was not a scalar");
+    }
+  printf ("Encountered: %s\n", (const gchar *)event.data.scalar.value);
```
sgallagher: Please replace this with `g_debug()` if you want to keep it in the debug logs.

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+              MMD_YAML_ERROR_EVENT_EXIT (
+                error, event, "Missing mapping in translation entry");
+              break;
+            }
+          if (!g_strcmp0 ((const gchar *)event.data.scalar.value, "summary"))
+            {
+              value = modulemd_yaml_parse_string (parser, error);
+              modulemd_translation_entry_set_summary (te, value);
+              g_free (value);
+              g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
```
sgallagher: Do not use this for logic issues. This is a detection of programming errors. (It reads "the error is either NULL (caller doesn't care about it) or it's non-NULL when initially passed into this function, meaning an error occurred earlier and wasn't handled then".)

See `ServiceLevel.emit_yaml()`. It would be better to create a new nested_error and then pass that to 

You should instead be checking that `value` is non-null and calling ` g_propagate_prefixed_error()` to set `error`.

Then just call `return NULL` here. The auto-pointer cleanups should handle everything else.

(I do realize I forgot to go back and do this in ServiceLevel.parse_yaml(), which I will fix.)

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+            }
+          else if (!g_strcmp0 ((const gchar *)event.data.scalar.value,
+                               "description"))
+            {
+              value = modulemd_yaml_parse_string (parser, error);
+              modulemd_translation_entry_set_description (te, value);
+              g_free (value);
+              g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
```
sgallagher: See above

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+            }
+          else if (!g_strcmp0 ((const gchar *)event.data.scalar.value,
+                               "profiles"))
+            {
+              modulemd_translation_entry_parse_yaml_profiles (
+                parser, te, error);
+              g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
```
sgallagher: See above

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+                error, event, "Missing mapping in translation entry profiles");
+              break;
+            }
+
+          profile_name = g_strdup ((const gchar *)event.data.scalar.value);
+          value = modulemd_yaml_parse_string (parser, error);
+          g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+          modulemd_translation_entry_set_profile_description (
+            te, profile_name, value);
+          g_free (profile_name);
```
sgallagher:
```suggestion
          g_clear_pointer(&profile_name, g_free);
```

This is preferred to avoid double-free when the autoptr cleanup hits.

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+
+static void *
+modulemd_translation_entry_parse_yaml_profiles (yaml_parser_t *parser,
+                                                ModulemdTranslationEntry *te,
+                                                GError **error)
+{
+  MMD_INIT_YAML_EVENT (event);
+  gboolean done = FALSE;
+  gboolean in_map = FALSE;
+  gchar *profile_name = NULL;
```
sgallagher:
```suggestion
  g_autofree gchar *profile_name = NULL;
```

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+  gchar *value = NULL;
```
sgallagher:
```suggestion
  g_autofree gchar *value = NULL;
```

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+modulemd_translation_entry_init (ModulemdTranslationEntry *self)
+{
+  self->profile_descriptions =
+    g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_free);
+}
+
+
+/* === YAML Functions === */
+
+static void *
```
sgallagher: This return value makes no sense. It should return a GHashTable representing the parsed profiles, NULL if an error occurred (so the consumer can know to check GError).

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+            {
+              value = modulemd_yaml_parse_string (parser, error);
+              modulemd_translation_entry_set_description (te, value);
+              g_free (value);
+              g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
+            }
+          else if (!g_strcmp0 ((const gchar *)event.data.scalar.value,
+                               "profiles"))
+            {
+              modulemd_translation_entry_parse_yaml_profiles (
```
sgallagher: This should check for NULL as mentioned above and then otherwise iterate through the returned hash-table and add them to the TranslationEntry object.

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+
+
+/* === YAML Functions === */
+
+static void *
+modulemd_translation_entry_parse_yaml_profiles (yaml_parser_t *parser,
+                                                ModulemdTranslationEntry *te,
+                                                GError **error)
+{
+  MMD_INIT_YAML_EVENT (event);
```
sgallagher: Add `MMD_INIT_TRACE()` here as well. It's intended for tracing the parsing and emitting functions and all their nested calls.

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,623 @@
...
+            error, nested_error, "Failed to emit profile value: ");
+          return FALSE;
+        }
+    }
+
+  ret = mmd_emitter_end_mapping (emitter, &nested_error);
+  if (!ret)
+    {
+      g_propagate_prefixed_error (
+        error, nested_error, "Failed to emit profiles start: ");
```
sgallagher:
```suggestion
        error, nested_error, "Failed to emit profiles end: ");
```

modulemd/v2/tests/ModulemdTests/translationentry.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def test_get_set_summary(self):
+        te = Modulemd.TranslationEntry(locale="en_US")
+
+        assert te.props.locale == 'en_US'
+        assert te.get_locale() == 'en_US'
+        assert te.props.summary == None
+        assert te.get_summary() == None
+
+        te.set_summary("foobar")
```
sgallagher: Could you also add something in non-ASCII like `(┛ಠ_ಠ)┛彡┻━┻` so we can be sure it's not doing anything wrong with the unicode?

modulemd/v2/modulemd-translation-entry.c
puiterwijk: Oh, the way I read it, `g_hash_table_insert` just keeps the old key, and clears the one that you passed in: `If you supplied a key_destroy_func when creating the GHashTable, the passed key is freed using that function.`.
So the way I read it, that doesn't do anything with the old key, but just clears the new one.
But sure, will make it replace.

modulemd/v2/modulemd-yaml-util.c
puiterwijk: I did not, and I removed it.
This was me totally missing "event" was a `g_auto`.

modulemd/meson.build
```diff
@@ -143,6 +147,7 @@ test_v2_priv_hdrs = files(
 )
 
 test_v2_python_scripts = files(
+    'v2/tests/ModulemdTests/translationentry.py',
```
sgallagher: Please include `base.py` here as well. This is the list of files that gets run through autopep8 for formatting.

modulemd/v2/modulemd-translation-entry.c
```diff
@@ -0,0 +1,648 @@
...
+static void
+modulemd_translation_entry_finalize (GObject *object)
+{
+  ModulemdTranslationEntry *self = (ModulemdTranslationEntry *)object;
+
+  g_clear_pointer (&self->locale, g_free);
+  g_clear_pointer (&self->summary, g_free);
+  g_clear_pointer (&self->description, g_free);
+
+  g_hash_table_unref (self->profile_descriptions);
```
sgallagher:
```suggestion
  g_clear_pointer (&self->profile_descriptions, g_hash_table_unref);
```

modulemd/v2/tests/ModulemdTests/translationentry.py
sgallagher: Please test a non-en_US locale.

PR #110, opened 2018-11-16
Add Modulemd.Defaults objects

sgallagher:
This implements only the abstract parent class (plus just enough of the `Modulemd.DefaultsV1` class to be able to run the tests).

### DIFF
--- a/modulemd/v2/include/modulemd-2.0/private/test-utils.h
+++ b/modulemd/v2/include/modulemd-2.0/private/test-utils.h
@@ -15,8 +15,29 @@
 
 #include <glib.h>
 #include <locale.h>
+#include <yaml.h>
 
 G_BEGIN_DECLS
+
+typedef struct _CommonMmdTestFixture
+{
+} CommonMmdTestFixture;
+
+extern int modulemd_test_signal;
+
+
+/**
+ * modulemd_test_signal_handler:
+ * @sig_num: The signal received
+ *
+ * Sets the global variable modulemd_test_signal with the value of the signal
+ * that was received.
+ *
+ * Since: 2.0
+ */
+void
+modulemd_test_signal_handler (int sig_num);
+
 
 /**
  * SECTION: test-utils
--- a/modulemd/v2/modulemd-defaults.c
+++ b/modulemd/v2/modulemd-defaults.c
@@ -1,0 +1,306 @@
+/*
+ * This file is part of libmodulemd
+ * Copyright (C) 2017-2018 Stephen Gallagher
+ *
+ * Fedora-License-Identifier: MIT
+ * SPDX-2.0-License-Identifier: MIT
+ * SPDX-3.0-License-Identifier: MIT
+ *
+ * This program is free software.
+ * For more information on the license, see COPYING.
+ * For more information on free software, see <https://www.gnu.org/philosophy/free-sw.en.html>.
+ */
+
+
+#include <inttypes.h>
+#include "modulemd-defaults.h"
+#include "modulemd-defaults-v1.h"
+#include "private/modulemd-util.h"
+
+#define DEF_DEFAULT_NAME_STRING "__NAME_UNSET__"
+
+typedef struct
+{
+  gchar *module_name;
+} ModulemdDefaultsPrivate;
+
+G_DEFINE_ABSTRACT_TYPE_WITH_PRIVATE (ModulemdDefaults,
+                                     modulemd_defaults,
+                                     G_TYPE_OBJECT)
+
+enum
+{
+  PROP_0,
+
+  PROP_MDVERSION,
+  PROP_MODULE_NAME,
+
+  N_PROPS
+};
+
+static GParamSpec *properties[N_PROPS];
+
+
+ModulemdDefaults *
+modulemd_defaults_new (guint64 mdversion, const gchar *module_name)
+{
+  g_return_val_if_fail (mdversion && mdversion <= MD_DEFAULTS_VERSION_LATEST,
+                        NULL);
+
+  if (mdversion == MD_DEFAULTS_VERSION_ONE)
+    {
+      return MODULEMD_DEFAULTS (modulemd_defaults_v1_new (module_name));
+    }
+
+  /* This should be unreachable */
+  g_return_val_if_reached (NULL);
+}
+
+
+static void
+modulemd_defaults_finalize (GObject *object)
+{
+  ModulemdDefaults *self = (ModulemdDefaults *)object;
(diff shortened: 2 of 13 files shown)


sgallagher:
OK, I made the changes I mentioned in the comment above and pushed them as a fixup patch so it's easier to see the differences. If the combined set passes code-review, I'll squash them together prior to merging.

modulemd/v2/modulemd-defaults.c
```diff
@@ -0,0 +1,309 @@
...
+  N_PROPS
+};
+
+static GParamSpec *properties[N_PROPS];
+
+
+ModulemdDefaults *
+modulemd_defaults_new (guint64 mdversion, const gchar *module_name)
+{
+  g_return_val_if_fail (mdversion && mdversion <= MD_DEFAULTS_VERSION_LATEST,
```
puiterwijk: Shouldn't this be `MD_DEFAULTS_VERSION_ONE`, since this would seem to imply that when a Defaultsv2 is added, it is no longer possible to create a DefaultsV1.
While I guess your intention would be to modify this later, maybe it's an idea to make this intention clear by either using `VERSION_ONE`, or adding a `MD_DEFAULTS_VERSION_LOWEST_SUPPORTED`?

modulemd/v2/modulemd-defaults.c
```diff
@@ -0,0 +1,309 @@
...
+                   mdversion);
+      return NULL;
+    }
+
+  if (modulemd_defaults_get_mdversion (self) == mdversion)
+    {
+      /* Already at this version, just copy it and return that */
+      return modulemd_defaults_copy (self);
+    }
+
```
puiterwijk: In this case, an error isn't set.

modulemd/v2/modulemd-defaults.c
```diff
@@ -0,0 +1,309 @@
...
+{
+  ModulemdDefaults *self = MODULEMD_DEFAULTS (object);
+
+  switch (prop_id)
+    {
+    case PROP_MODULE_NAME:
+      modulemd_defaults_set_module_name (self, g_value_get_string (value));
+      break;
+
+    case PROP_MDVERSION:
```
puiterwijk: Maybe instead return an explicit error that while `MDVERSION` is valid, it can't be set?

modulemd/v2/modulemd-defaults.c
```diff
@@ -0,0 +1,309 @@
...
+
+  klass->copy = modulemd_defaults_default_copy;
+  klass->validate = modulemd_defaults_default_validate;
+
+  properties[PROP_MDVERSION] = g_param_spec_uint64 (
+    "mdversion",
+    "Metadata Version",
+    "The metadata version of this Defaults object. Defaults to the "
+    "highest-supported version if not specified at object construction. "
+    "Cannot be changed after object construction except through the upgrade() "
```
puiterwijk: Note that the `upgrade()` routine doesn't modify this on the object it receives, but creates a new one.
Thta makes this phrasing sound incorrect to me.

modulemd/v2/tests/ModulemdTests/defaults.py
```diff
@@ -0,0 +1,137 @@
...
+            assert False
+        except TypeError as e:
+            assert 'constructor returned NULL' in e.__str__()
+
+        # Test with an unknown mdversion
+        try:
+            with self.expect_signal():
+                defs = Modulemd.Defaults.new(
+                    Modulemd.DefaultsVersionEnum.LATEST + 1, 'foo')
+            assert False
```
puiterwijk: Perhaps instead of this `try:` block, use `self.assertRaises`?
For an example, see the `Profile` patch: https://github.com/fedora-modularity/libmodulemd/pull/113/files#diff-3a8af7909cf33247e7740e17b3bcfb58R60.

modulemd/v2/include/modulemd-2.0/private/test-utils.h
```diff
@@ -15,9 +15,30 @@
...
+ * modulemd_test_signal_handler:
+ * @sig_num: The signal received
+ *
+ * Sets the global variable modulemd_test_signal with the value of the signal
+ * that was received.
+ *
+ * Since: 2.0
+ */
+void
+modulemd_test_signal_handler (int sig_num);
```
puiterwijk: Given how often we try to run some code to check that it emits a signal, maybe it's an idea to instead make a macro that takes the signal and a function to run and sets up and asserts the signal handler?
Kind of like the `expect_signal()` in the python `base.py`?

puiterwijk:
CI test failure looks like just rawhide failure.

modulemd/v2/modulemd-defaults.c
sgallagher: That's exactly what this does. It falls through to the default clause which does `G_OBJECT_WARN_INVALID_PROPERTY_ID` which means "this property isn't valid for this call".

modulemd/v2/modulemd-defaults.c
sgallagher: You're right. I'll rephrase.

modulemd/v2/tests/ModulemdTests/defaults.py
sgallagher: Ah, I didn't see that in the other patch. Handy!

sgallagher: Replaced all of these cases with self.assertRaisesRegex() so I can also verify that the failure message is correct.

modulemd/v2/include/modulemd-2.0/private/test-utils.h
sgallagher: Yeah, I was thinking the same thing. I'll look into that in a follow-up patch.

sgallagher:
Modified the description of the `mdversion` property and replaced all of the `TypeError` checks in the python tests with `assertRaisesRegex()`. As before, I pushed this as a fixup commit so the differences are easier to spot, but if this is approved, I'll squash them together before merging.

sgallagher:
Also pushed another minor fix to have the `copy()` test verify that the `module_name` was also duplicated correctly.

modulemd/v2/modulemd-defaults.c
puiterwijk: Right, I meant to suggest not using the fallthrough and returning an explicit "This attribute is not supposed to be modified" error message.
But if you think the standard error suffices, fine with me.

sgallagher:
Rebased atop the Profile patch.

PR #113, opened 2018-11-16
Implement Profile v2 object

puiterwijk:

### DIFF
--- a/modulemd/v2/modulemd-profile.c
+++ b/modulemd/v2/modulemd-profile.c
@@ -1,0 +1,403 @@
+/*
+ * This file is part of libmodulemd
+ * Copyright (C) 2018 Red Hat, Inc.
+ *
+ * Fedora-License-Identifier: MIT
+ * SPDX-2.0-License-Identifier: MIT
+ * SPDX-3.0-License-Identifier: MIT
+ *
+ * This program is free software.
+ * For more information on the license, see COPYING.
+ * For more information on free software, see <https://www.gnu.org/philosophy/free-sw.en.html>.
+ */
+
+#include <glib.h>
+#include <yaml.h>
+
+#include "modulemd-profile.h"
+#include "private/glib-extensions.h"
+#include "private/modulemd-profile-private.h"
+#include "private/modulemd-util.h"
+#include "private/modulemd-yaml.h"
+
+#define P_DEFAULT_STRING "__PROFILE_NAME_UNSET__"
+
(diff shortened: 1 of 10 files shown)


modulemd/v2/modulemd-profile.c
```diff
@@ -0,0 +1,461 @@
...
+
+  return self->description;
+}
+
+
+void
+modulemd_profile_add_rpm (ModulemdProfile *self, const gchar *rpm)
+{
+  g_return_if_fail (MODULEMD_IS_PROFILE (self));
+  g_hash_table_insert (self->rpms, g_strdup (rpm), NULL);
```
sgallagher: Please use [g_hash_table_add()](https://developer.gnome.org/glib/stable/glib-Hash-Tables.html#g-hash-table-add) instead. It's optimized for this precise usage.

modulemd/v2/modulemd-profile.c
```diff
@@ -0,0 +1,461 @@
...
+          if (!in_rpms)
+            MMD_YAML_ERROR_EVENT_EXIT (error, event, "Unexpected end of list");
+          in_rpms = FALSE;
+          break;
+
+        case YAML_SCALAR_EVENT:
+          if (!in_map)
+            MMD_YAML_ERROR_EVENT_EXIT (
+              error, event, "Missing mapping in profile entry");
+          if (in_rpms)
```
sgallagher: Since this is going to come up frequently in other places (there are several such lists in ModuleStream), could you instead add a common routine to read in all strings in a YAML sequence into a passed GHashTable? That will also simplify the logic at this layer, rather than needing to maintain in_rpms as well.

modulemd/v2/modulemd-profile.c
```diff
@@ -0,0 +1,461 @@
...
+static gboolean
+modulemd_profile_emit_yaml_rpms (ModulemdProfile *self,
+                                 yaml_emitter_t *emitter,
+                                 GError **error)
+{
+  MODULEMD_INIT_TRACE ();
+  int ret;
+  g_autoptr (GError) nested_error = NULL;
+  MMD_INIT_YAML_EVENT (event);
+  g_auto (GStrv) rpms = modulemd_profile_get_rpms_as_strv (self);
```
sgallagher: Similarly to the above parser request, could you add a common emitter routine to take a GHashTable and emit its string keys as a YAML sequence?

PR #116, opened 2018-11-20
Assorted utility functions for libmodulemd

sgallagher:


modulemd/v2/include/modulemd-2.0/private/modulemd-yaml.h
```diff
@@ -244,4 +291,47 @@ GHashTable *
...
+
+
+/**
+ * modulemd_yaml_emit_document_headers:
+ * @emitter: (inout): A libyaml emitter object that is positioned where the
+ * YAML_DOCUMENT_START_EVENT should occur (so this must be after either a
+ * YAML_STREAM_START_EVENT or YAML_DOCUMENT_END_EVENT).
+ * @doctype: (in): The document type (see #ModulemdYamlDocumentType)
+ * @mdversion: (in): The metadata version for this document
+ *
```
puiterwijk: Missing the "@error: "

modulemd/v2/modulemd-yaml-util.c
```diff
@@ -382,6 +386,22 @@ modulemd_yaml_parse_string (yaml_parser_t *parser, GError **error)
...
+{
+  MMD_INIT_YAML_EVENT (event);
+
+  YAML_PARSER_PARSE_WITH_EXIT_INT (parser, &event, error);
+  if (event.type != YAML_SCALAR_EVENT)
+    {
+      MMD_YAML_ERROR_EVENT_EXIT_INT (error, event, "String was not a scalar");
+    }
+
+  return g_ascii_strtoull ((const gchar *)event.data.scalar.value, NULL, 10);
```
puiterwijk: This would not be according to the documentation for this function, I believe.
Specifically, with the error cases: this function might return `G_MAXUINT64` or zero in case of errors.
I am wondering whether we should be interpreting those values and the corresponding `errno` values and then set an error?
Or, alternatively, document this?

modulemd/v2/modulemd-yaml-util.c
```diff
@@ -421,3 +441,273 @@ modulemd_yaml_parse_string_set (yaml_parser_t *parser, GError **error)
...
+
+        case YAML_SEQUENCE_END_EVENT: depth--; break;
+
+        case YAML_MAPPING_END_EVENT:
+          depth--;
+          if (depth == 0)
+            done = TRUE;
+          break;
+
+        default: break;
```
puiterwijk: Perhaps in the `default` case, we should also have a `if (depth == 0) done = TRUE;`, the reason being that otherwise if one has a document with `data: gotcha` or `data: []`, we might be parsing up the rest of the document, and then try to proceed where we should have stopped.
Now, this would most likely be an invalid document further on, but we should probably make sure that we don't fail in unexpected ways.

Maybe do something like:
````
case YAML_SEQUENCE_END_EVENT:
case YAML_MAPPING_END_EVENT:
  depth--
default:
  if (depth == 0)
    done = TRUE;
  break;
````

File: modulemd/v2/include/modulemd-2.0/private/modulemd-yaml.h [hunk 4]
```diff
@@ -229,6 +260,22 @@ gchar *
 modulemd_yaml_parse_string (yaml_parser_t *parser, GError **error);
 
 
+/**
+ * modulemd_yaml_parse_uint64:
+ * @parser: (inout): A libyaml parser object positioned at the beginning of a
+ * string scalar entry.
```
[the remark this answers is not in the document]
sgallagher: Yeah, copy-paste error. I'll fix it.

modulemd/v2/include/modulemd-2.0/private/modulemd-yaml.h
```diff
@@ -244,4 +291,47 @@ GHashTable *
 modulemd_yaml_parse_string_set (yaml_parser_t *parser, GError **error);
 
 
+/**
+ * modulemd_yaml_parse_document_type:
+ * @parser: (inout): A libyaml parser object positioned at the beginning of a
+ * yaml subdocument immediately prior to a YAML_DOCUMENT_START_EVENT.
+ * @doctype: (out): The type of document that was encounters.
```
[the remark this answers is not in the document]
sgallagher: Yes

modulemd/v2/include/modulemd-2.0/private/modulemd-yaml.h
sgallagher: Good catch

sgallagher:
Thanks for the review. Fixes force-pushed. I also added a "Returns:" line for `modulemd_yaml_parse_document_type()` and `modulemd_yaml_emit_document_headers()`

modulemd/v2/modulemd-yaml-util.c
puiterwijk: Agreed

PR #118, opened 2018-11-21
Add Modulemd.ModuleStream

sgallagher:
Includes a skeleton for the two child classes to enable the tests.

Contains only Python-based tests at the moment, which means that valgrind isn't currently running against this code. We should add C-based tests in a follow-up.

Note in particular that this object has an unusual `copy()` routine. For a variety of reasons, the module stream name needs to remain constant for the life of the object. In the event that the stream name needs to change, it has to be done by copying the object entirely. (This is mainly so that if you want to add it back to the `ModuleIndex` later, it will hook into all the correct places such as defaults and translations.)


sgallagher:
@asamalik If you'd be willing to copy the Python tests into C so they can run with valgrind, that would be helpful.

asamalik:
@sgallagher On it!

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
```diff
@@ -0,0 +1,189 @@
...
+  MD_MODULESTREAM_VERSION_UNSET = 0,
+
+  MD_MODULESTREAM_VERSION_ONE = 1,
+  MD_MODULESTREAM_VERSION_TWO = 2,
+
+  MD_MODULESTREAM_VERSION_LATEST = MD_MODULESTREAM_VERSION_TWO
+} ModulemdModuleStreamVersionEnum;
+
+
+#define MODULEMD_TYPE_DEFAULTS (modulemd_defaults_get_type ())
```
puiterwijk: Is this a copy-paste error from defaults?

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
```diff
@@ -0,0 +1,189 @@
...
+   * breaking ABI. */
+  gpointer padding[10];
+};
+
+
+/**
+ * modulemd_module_stream_new:
+ * @mdversion: (in): The metadata version of ModuleStream to create.
+ * @module_name: (in): The name of the module.
+ * @stream_name: (in) (nullable): The name of this stream. Optional.
```
puiterwijk: Note that the spec calls this `module_stream`, and that's the name we used in other objects too.
Not a big deal, but just consistency note.

modulemd/v2/modulemd-module-stream.c
```diff
@@ -0,0 +1,346 @@
...
+
+ModulemdModuleStream *
+modulemd_module_stream_copy (ModulemdModuleStream *self,
+                             const gchar *stream_name)
+{
+  ModulemdModuleStreamClass *klass;
+
+  g_return_val_if_fail (MODULEMD_IS_MODULE_STREAM (self), NULL);
+
+  if (!self)
```
puiterwijk: Shouldn't this be caught by the `g_return_val_if_fail()` call?

modulemd/v2/modulemd-module-stream.c
```diff
@@ -0,0 +1,346 @@
...
+
+#include "modulemd-module-stream.h"
+#include "modulemd-module-stream-v1.h"
+#include "modulemd-module-stream-v2.h"
+#include "private/modulemd-module-stream-private.h"
+#include "private/modulemd-util.h"
+
+typedef struct
+{
+  guint64 mdversion;
```
puiterwijk: Is there a specific reason you're also storing this if the getter returns the mdversion of the child class?

modulemd/v2/modulemd-module-stream.c
```diff
@@ -0,0 +1,346 @@
...
+
+
+static gboolean
+modulemd_module_stream_default_validate (ModulemdModuleStream *self,
+                                         GError **error)
+{
+  ModulemdModuleStreamPrivate *priv =
+    modulemd_module_stream_get_instance_private (self);
+
+  if (!priv->mdversion)
```
puiterwijk: Maybe we should just use the getter to call it on the child class?

modulemd/v2/modulemd-v2-docs.xml
```diff
@@ -39,6 +39,8 @@
         <xi:include href="xml/modulemd-defaults.xml"/>
         <xi:include href="xml/modulemd-defaults-v1.xml"/>
         <xi:include href="xml/modulemd-dependencies.xml"/>
+        <xi:include href="xml/modulemd-module-stream.xml"/>
```
puiterwijk: I believe this may be missing the `v1`.

modulemd/v2/modulemd-v2-docs.xml
```diff
@@ -49,6 +51,7 @@
        <xi:include href="xml/modulemd-dependencies-private.xml"/>
        <xi:include href="xml/modulemd-defaults-private.xml"/>
        <xi:include href="xml/modulemd-defaults-v1-private.xml"/>
+       <xi:include href="xml/modulemd-module-stream-private.xml"/>
```
puiterwijk: I believe you're missing the `private` files for `v1` and `v2`.

modulemd/v2/modulemd-v2-docs.xml
```diff
@@ -60,4 +63,4 @@
         <xi:include href="xml/test-utils.xml"/>
     </chapter>
   </reference>
-</book>
+</book>
```
puiterwijk: Quite funny how we keep switching between trailing newline and not...
I believe your IDE might not like them, while mine keeps re-adding them.

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
sgallagher: Looks like it. I'll drop it.

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
sgallagher: Agreed, we should be consistent. Fixed.

modulemd/v2/modulemd-module-stream.c
sgallagher: Nope, relic of an earlier thought-process. Removed.

modulemd/v2/modulemd-module-stream.c
sgallagher: Incorrect ordering. The NULL check should return before `g_return_val_if_fail()` because that's a valid request (copy nothing, get nothing).

modulemd/v2/modulemd-module-stream.c
sgallagher: I reworked this so this check is properly done in the parent class.

modulemd/v2/modulemd-v2-docs.xml
sgallagher: Looks like it, yeah. Odd, I could have sworn I checked that...

sgallagher:
This has been reworked a bit. As @puiterwijk and I discussed on IRC, I've moved the "version" and "context" fields to the parent class as they're going to be needed by the `ModuleIndex` integration and are therefore required by any class implementing `ModuleStream`.

I also added the skeleton of a YAML parsing routine for `ModuleStream`. The actual implementation will be done as part of the implementations of `ModuleStreamV1` and `ModuleStreamV2`.

Please re-review.

sgallagher:
As a follow-up to moving the version and context properties to the parent class, I also just pushed an additional patch to this PR that adds the `get_nsvc_as_string()` call here, rather than in the derived classes.

PR #121, opened 2018-11-25
Run CI on all releases of Fedora

sgallagher:
Sometimes one release is broken by unrelated issues. This will
help us figure out both when libmodulemd has a CI failure and
also other issues in Fedora that can be reported.


sgallagher:
This should be ready to merge now. The Rawhide test is known to fail right now due to https://bugzilla.redhat.com/show_bug.cgi?id=1653023 but should be fixed in tomorrow's Rawhide compose.

There is some early efforts at supporting a CentOS 7 set of tests here as well, but as there are some issues specific to that platform, I'm deferring completing that work to a subsequent patch.

PR #127, opened 2018-11-28
SubdocumentInfo v2

puiterwijk:

### DIFF
--- a/modulemd/v2/modulemd-subdocument-info.c
+++ b/modulemd/v2/modulemd-subdocument-info.c
@@ -1,0 +1,300 @@
+/*
+ * This file is part of libmodulemd
+ * Copyright (C) 2018 Red Hat, Inc.
+ *
+ * Fedora-License-Identifier: MIT
+ * SPDX-2.0-License-Identifier: MIT
+ * SPDX-3.0-License-Identifier: MIT
+ *
+ * This program is free software.
+ * For more information on the license, see COPYING.
+ * For more information on free software, see <https://www.gnu.org/philosophy/free-sw.en.html>.
+ */
+
+#include <glib.h>
+#include <yaml.h>
+
+#include "modulemd-subdocument-info.h"
+#include "private/glib-extensions.h"
+#include "private/modulemd-subdocument-info-private.h"
+#include "private/modulemd-util.h"
+#include "private/modulemd-yaml.h"
+
+#define SD_DEFAULT_STRING "__SUBDOCUMENT_INFO_VALUE_UNSET__"
+
+struct _ModulemdSubdocumentInfo
+{
+  GObject parent_instance;
+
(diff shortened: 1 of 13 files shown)


modulemd/v2/modulemd-subdocument-info.c
```diff
@@ -0,0 +1,300 @@
...
+  return self->mdversion;
+}
+
+
+gboolean
+modulemd_subdocument_info_get_data_parser (ModulemdSubdocumentInfo *self,
+                                           yaml_parser_t *parser,
+                                           GError **error)
+{
+  g_return_val_if_fail (MODULEMD_IS_SUBDOCUMENT_INFO (self), FALSE);
```
sgallagher: We should have a check for the validity of the GError pointer here.
```suggestion
  g_return_val_if_fail (MODULEMD_IS_SUBDOCUMENT_INFO (self), FALSE);
  g_return_val_if_fail (error == NULL || *error == NULL, FALSE);
```

modulemd/v2/modulemd-yaml-util.c
```diff
@@ -458,33 +459,46 @@ modulemd_yaml_parse_string_set (yaml_parser_t *parser, GError **error)
...
+
   /* The first event must be the document start */
   YAML_PARSER_PARSE_WITH_EXIT_BOOL (parser, &event, error);
   if (event.type != YAML_DOCUMENT_START_EVENT)
     {
       MMD_YAML_ERROR_EVENT_EXIT_BOOL (
         error, event, "Parser is not at the start of a document");
     }
+  MMD_EMIT_WITH_EXIT_FULL (
+    emitter, FALSE, &event, error, "Error starting stream");
```
sgallagher:
```suggestion
    emitter, FALSE, &event, error, "Error starting document");
```

modulemd/v2/modulemd-yaml-util.c
```diff
@@ -555,15 +587,17 @@ modulemd_yaml_parse_document_type (yaml_parser_t *parser,
                   g_propagate_error (error, nested_error);
                   return FALSE;
                 }
+              /* TODO: Emit this scalar */
```
sgallagher: Drop this TODO message.

Also, there's a bug in the original code here; if mdversion is 0, it might be due to a parse error *or* it might be that there was a literal zero in the document. We can't just return the `nested_error` as-is, since it might be NULL. Instead of `if (!mdversion)`, we should do `if (nested_error)`. The case where `mdversion == 0` is already handled at line 636.

sgallagher:
Oh, and the docs weren't added to the modulemd-docs.xml. I'll do that as well.

sgallagher:
Merged along with a504b3a19eb6d0e96a5ef4de63e966294dd59900 fixes.

PR #132, opened 2018-11-30
Add locale option to summary and description getters

sgallagher:
Add locale option to summary and description getters

The ModuleStream.get_summary(), ModuleStream.get_description() and
Profile.get_description() fields need to take an optional locale in
order to return translated strings.

It also drops the properties associated with summary and description
as it is the safest course of action, rather than having users guess
which string they will get back, they must use the getter which will
handle the locale appropriately.

This patch updates the API but does not yet implement the
translation, which will be done later while implementing ModuleIndex
that handles associating the Translation documents with the streams.

Also reflows the gtkdoc comments in modulemd-profile.h.

### DIFF
--- a/modulemd/v2/modulemd-profile.c
+++ b/modulemd/v2/modulemd-profile.c
@@ -39,7 +39,6 @@ enum
   PROP_0,
 
   PROP_NAME,
-  PROP_DESCRIPTION,
 
   N_PROPS
 };
@@ -62,8 +61,8 @@ modulemd_profile_copy (ModulemdProfile *self)
 
   p = modulemd_profile_new (modulemd_profile_get_name (self));
 
-  modulemd_profile_set_description (p,
-                                    modulemd_profile_get_description (self));
+  modulemd_profile_set_description (
+    p, modulemd_profile_get_description (self, NULL));
 
   g_hash_table_unref (p->rpms);
   p->rpms = g_hash_table_ref (self->rpms);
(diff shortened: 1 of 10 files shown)


modulemd/v2/modulemd-profile.c
```diff
@@ -62,8 +62,8 @@ modulemd_profile_copy (ModulemdProfile *self)
 
   p = modulemd_profile_new (modulemd_profile_get_name (self));
 
-  modulemd_profile_set_description (p,
-                                    modulemd_profile_get_description (self));
```
puiterwijk: Using `NULL` in the tests would mean that their success/failure would be depending on the actual locale in tests where we build the object with a translation.
Maybe it's an idea to instead just pass an explicit `"C"`, just to make sure that any tests that test with translated entries, the primary test isn't ending up failing on the `robots_TABLES` locale?

puiterwijk: Or, alternatively, we could make sure to set a locale for the tests in the meson configuration?

sgallagher: Yeah, having meson set `LC_ALL=C` in the environment seems like the easiest approach.

sgallagher:
I just realized that the properties for summary and description disagree with the implementation as written in this patch.

After discussion with @puiterwijk, we're going to drop the object properties for summary and description so that we can avoid ambiguity about what they would return. Consumers will use the getters and setters instead.

sgallagher:
OK, I just pushed the changes discussed above, as well as adjusting the tests to use the "C" locale explicitly (in addition to setting the env var for safety).

sgallagher:
Also edited the description field for the PR to match reality.

sgallagher:
Rebased atop the rename-to patches. No other changes. so I will merge once the CI passes.

PR #134, opened 2018-11-30
Implement renames for _as_strv methods

puiterwijk:


puiterwijk:
I'm submitting this PR to give an indication of how the renames would look like in code.
Unfortunately, there must be a target definition, and it must be in a header that's ingested by gobject-introspection.

puiterwijk:
Added renames for all of the `_as_` methods except for `get_eol_as_string`.

PR #135, opened 2018-11-30
Python Overrides

sgallagher:
First patch adds overrides for the `ServiceLevel.set_eol()` and `ServiceLevel.get_eol()` functions. It also reworks slightly the method used to determine when to apply the overrides to base it on the API version of the Modulemd namespace, rather than on presence of the symbol in the namespace, since ServiceLevel existed in 1.0 and we don't want to break compatibility there if both APIs are available on the system.

The first patch also adds the overrides to the autopep8 formatter, which results in a few minor formatting changes as well.

The second patch drops an unneeded override function from ModuleStreamV1 and V2 that were leftovers from trial-and-error to get the overrides properly applied. They are entirely superfluous.

### DIFF
--- a/bindings/python/gi/overrides/Modulemd.py
+++ b/bindings/python/gi/overrides/Modulemd.py
@@ -36,9 +36,14 @@ Modulemd = get_introspection_module('Modulemd')
 from six import text_type
 from gi.repository import GLib
 
+import datetime
+
+
 __all__ = []
 
+
 class ModulemdUtil():
+
     def variant_str(s):
         """ Converts a string to a GLib.Variant
         """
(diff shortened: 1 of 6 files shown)


sgallagher:
Updated patch set. I reordered them so the first patch is removing the unnecessary ModuleStream overrides. The second patch drops "eol" as a property on ServiceLevel to make way for the overrides doing the automatic conversions. The third patch is the corrected override for `get_eol()` and `set_eol()` which also adds a new construct to the tests to validate the bindings both with and without overrides available.

sgallagher:
Pushed one additional patch to make the same improved check for whether we're running with overrides in the XMD tests.

bindings/python/gi/overrides/Modulemd.py
```diff
@@ -134,7 +136,28 @@ def get_xmd(self):
...
 
-
     ModuleStreamV1 = override(ModuleStreamV1)
     __all__.append(ModuleStreamV1)
 
+    class ServiceLevel(Modulemd.ServiceLevel):
+
+        def set_eol(self, eol):
+            if (isinstance(eol, datetime.date)):
+                return super().set_eol_ymd(eol.year, eol.month, eol.day)
```
puiterwijk: Note that `super().foo` only works on py3. On py2, this should be `super(ServiceLeve, self).foo`.
The latter works with both though, so using that might be better.

PR #141, opened 2018-12-02
Add Translation support to Module and ModuleStream

puiterwijk:

### DIFF
--- a/modulemd/v2/include/modulemd-2.0/private/modulemd-module-private.h
+++ b/modulemd/v2/include/modulemd-2.0/private/modulemd-module-private.h
@@ -17,6 +17,7 @@
 #include <yaml.h>
 
 #include "modulemd-module.h"
+#include "modulemd-translation.h"
 
 /**
  * SECTION: modulemd-module-private
@@ -51,6 +52,31 @@ modulemd_module_new (const gchar *module_name);
 void
 modulemd_module_set_defaults (ModulemdModule *self,
                               ModulemdDefaults *defaults);
+
+
+/**
+ * modulemd_module_associate_translation:
+ * @self: This #ModulemdModule object
+ * @translation: (in): A #ModulemdTranslation object which is copied into the
+ * #ModulemdModule object.
+ *
+ * Since: 2.0
+ */
+void
+modulemd_module_add_translation (ModulemdModule *self,
+                                 ModulemdTranslation *translation);
+
+
+/**
+ * modulemd_module_get_translations:
+ * @self: This #ModulemdModule object
+ *
+ * Returns: (transfer none): The set of translations attached to streams.
+ *
+ * Since: 2.0
+ */
+GHashTable *
+modulemd_module_get_translations (ModulemdModule *self);
 
 
 /**
--- a/modulemd/v2/modulemd-module-stream.c
+++ b/modulemd/v2/modulemd-module-stream.c
@@ -29,6 +29,7 @@ typedef struct
   gchar *stream_name;
   guint64 version;
   gchar *context;
+  ModulemdTranslation *translation;
 } ModulemdModuleStreamPrivate;
 
 G_DEFINE_ABSTRACT_TYPE_WITH_PRIVATE (ModulemdModuleStream,
@@ -281,6 +282,7 @@ modulemd_module_stream_finalize (GObject *object)
   g_clear_pointer (&priv->module_name, g_free);
   g_clear_pointer (&priv->stream_name, g_free);
   g_clear_pointer (&priv->context, g_free);
+  g_clear_pointer (&priv->translation, g_object_unref);
 
   G_OBJECT_CLASS (modulemd_module_stream_parent_class)->finalize (object);
 }
@@ -328,6 +330,8 @@ modulemd_module_stream_default_copy (ModulemdModuleStream *self,
     copy, modulemd_module_stream_get_version (self));
   modulemd_module_stream_set_context (
     copy, modulemd_module_stream_get_context (self));
+  modulemd_module_stream_associate_translation (
+    copy, modulemd_module_stream_get_translation (self));
 
   return g_steal_pointer (&copy);
 }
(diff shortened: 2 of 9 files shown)


puiterwijk:
Tests added

modulemd/v2/modulemd-module-stream.c
```diff
@@ -684,3 +688,53 @@ static void
...
+                                        ModulemdTranslation *translation)
+{
+  g_return_if_fail (MODULEMD_IS_MODULE_STREAM (self));
+
+  ModulemdModuleStreamPrivate *priv =
+    modulemd_module_stream_get_instance_private (self);
+
+  g_clear_pointer (&priv->translation, g_object_unref);
+  if (translation != NULL)
+    priv->translation = modulemd_translation_copy (translation);
```
sgallagher: I'm really not sure we want to be copying this data around. My original thought was that we'd make these into references. In some module-stream-expansion cases, we may have MANY instances of the same module/stream pair in the Index. (Multiple versions, each with multiple contexts can grow quickly). I'd much rather we maintained a single copy at the ModuleIndex level that we associate to the individual streams via `g_object_ref()`. The summaries and descriptions can be arbitrarily long, and I'd rather not be duplicating hundreds of kilobytes of memory unnecessarily.

Also, if we move to referencing them, I think we should change the function name from `modulemd_module_stream_set_translation()` to `modulemd_module_stream_associate_translation()`, which will make it more clear that it's taking a reference and not making a copy.

modulemd/v2/modulemd-module.c
```diff
@@ -205,10 +211,20 @@ modulemd_module_get_defaults (ModulemdModule *self)
 void
 modulemd_module_add_stream (ModulemdModule *self, ModulemdModuleStream *stream)
 {
+  ModulemdTranslation *translation = NULL;
+  ModulemdModuleStream *newstream = NULL;
   g_return_if_fail (MODULEMD_IS_MODULE (self));
 
-  g_ptr_array_add (self->streams,
-                   modulemd_module_stream_copy (stream, NULL, NULL));
+  //newstream = modulemd_module_stream_copy (stream, NULL, NULL);
```
sgallagher: This should probably be removed.

modulemd/v2/modulemd-module.c
```diff
@@ -205,10 +211,20 @@ modulemd_module_get_defaults (ModulemdModule *self)
...
+  newstream = g_object_ref (stream);
+  g_ptr_array_add (self->streams, newstream);
```
sgallagher: Why did you move from a copy to a reference here?

Also, it just occurred to me that this function really ought to have
```
g_return_if_fail (g_str_equal(modulemd_module_stream_get_name(stream), self->module_name);
```
since it should really be a coding error if we ever try to add a stream whose name doesn't match the module.

modulemd/v2/modulemd-module.c
```diff
@@ -293,3 +309,39 @@ modulemd_module_get_stream_by_NSVC (ModulemdModule *self,
...
+
+void
+modulemd_module_add_translation (ModulemdModule *self,
+                                 ModulemdTranslation *translation)
+{
+  gsize i;
+  ModulemdModuleStream *stream = NULL;
+
+  if (!g_str_equal (modulemd_translation_get_module_name (translation),
+                    modulemd_module_get_module_name (self)))
```
sgallagher: This should be `g_return_if_fail()` so that it generates a log message and calls `abort()` when `G_DEBUG=fatal-warnings`. We should be able to assume that the caller should already have validated this and isn't just trying to assign translations at random. Or if they are, they should rightly get called out on it.

modulemd/v2/modulemd-module.c
```diff
@@ -293,3 +309,39 @@ modulemd_module_get_stream_by_NSVC (ModulemdModule *self,
...
+    return;
+
+  g_hash_table_replace (
+    self->translations,
+    g_strdup (modulemd_translation_get_module_stream (translation)),
+    g_object_ref (translation));
```
sgallagher: You're taking a reference to the translation here, but making a copy of it for each individual stream that is added. That really doesn't make any sense to me. We should make a copy onto the highest-level location that will need to reference it, then take a reference on it at each lower level. So in this case, I think we need to take the copy *here* and then have `modulemd_module_stream_associate_translation()` take a reference on each stream that matches.

Also, when implementing the ModuleIndex we need to have the addition of a Translation or Defaults object cause the creation of the appropriate Module object, if a Stream has not already done so.

modulemd/v2/modulemd-module.c
puiterwijk: The move from copy to ref is a result of the fact that the stream v1 and v2 objects don't yet have their own `copy` implementations, so that internal things (like the summary and description) don't get copied when you copy a stream.
Because of that I moved to a ref during development so I could at least verify that this specific patchset was working.
This patch probably should wait until we've added that, so that the tests can pass with a copy.

sgallagher: Aargh, I didn't implement the `copy()`, did I? OK, I'll get that done by the end of Monday...

puiterwijk: I actually have a fair bit of the `copy()` working for stream v1, so will finish that up and submit by Monday morning.

modulemd/v2/modulemd-module.c
puiterwijk: The creation of the Module object I had already done: `get_or_create_module`

puiterwijk:
I updated this to fix the requested changes. It is now based on top of #143 to have a working `copy()`.

modulemd/v2/include/modulemd-2.0/private/modulemd-module-private.h
```diff
@@ -53,6 +54,31 @@ modulemd_module_set_defaults (ModulemdModule *self,
                               ModulemdDefaults *defaults);
 
 
+/**
+ * modulemd_module_associate_translation:
+ * @self: This #ModulemdModule object
+ * @translation: A #ModulemdTranslation object to associate with this
+ * #ModulemdModule
```
sgallagher: GObject treats it as implicit here, but I'd prefer the `(in)` annotation for `@translation` as well as having the comment expressly state that it takes a reference and not a copy of the translation.

modulemd/v2/modulemd-module.c
sgallagher: You marked this "resolved" but didn't change it to `g_return_if_fail()` here. I'm guessing this was an oversight.

modulemd/v2/modulemd-module.c
sgallagher: We need to take a copy of the translation here (and use references to that copy within the associated streams). It's possible at the Index layer that we would get a Translation document passed in rather than built from YAML and we don't want to make the caller responsible for the memory (or for not changing it after it's referenced).

puiterwijk:
Okay, I think I finally might have the memory ownership correct now...

PR #144, opened 2018-12-03
Module Index v2

puiterwijk:
ModuleIndex v2 implementation

Dependencies (these are in this patchset for now):
- https://github.com/fedora-modularity/libmodulemd/pull/141
- https://github.com/fedora-modularity/libmodulemd/pull/143

Todo:
- Stream (v1/v2) emitting: not implemented at Stream layer yet


modulemd/v2/modulemd-module-index.c
```diff
@@ -0,0 +1,536 @@
...
+{
+  g_return_val_if_fail (MODULEMD_IS_MODULE_INDEX (self), NULL);
+
+  MMD_INIT_YAML_EMITTER (emitter);
+  MMD_INIT_YAML_STRING (&emitter, yaml_string);
+
+  if (!modulemd_module_index_dump_to_emitter (self, &emitter, error))
+    return NULL;
+
+  return g_strdup (yaml_string->str);
```
sgallagher: Use `g_steal_pointer()` here instead. It's likely to be a very large string and there's no need to copy it as it's about to be freed anyway.

modulemd/v2/modulemd-module-index.c
```diff
@@ -0,0 +1,536 @@
...
+
+  MMD_INIT_YAML_EMITTER (emitter);
+  yaml_emitter_set_output_file (&emitter, yaml_stream);
+
+  return modulemd_module_index_dump_to_emitter (self, &emitter, error);
+}
+
+
+gboolean
+modulemd_module_index_validate (ModulemdModuleIndex *self, GError **error)
```
sgallagher: Let's drop the `validate()` function for now (or at least make it non-public). I'm not sure we want this to be a simple boolean or if we want to report any and all failures on contained objects in some other manner.

Rather than design it right now, let's just avoid making it public API.

modulemd/v2/modulemd-module-index.c
```diff
@@ -0,0 +1,536 @@
...
+        return FALSE;
+      if (!modulemd_module_index_add_translation (self, translation, error))
+        return FALSE;
+      break;
+
+    default:
+      g_set_error (error,
+                   MODULEMD_YAML_ERROR,
+                   MODULEMD_YAML_ERROR_PARSE,
+                   "Invalid doctype encountered");
```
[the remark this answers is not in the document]
puiterwijk: Note that Translation doesn't have a versioned parse function, so this case is when it's not any of the known doctypes.

puiterwijk:
Okay, I think I have resolved the issues.

PR #150, opened 2018-12-04
ModuleIndex only supports ModuleStream with name and stream

sgallagher:
Also replaces calls to `g_str_equal()` with `g_strcmp0()` which is
NULL-safe, since the values may be unset.

### DIFF
--- a/modulemd/v2/include/modulemd-2.0/private/modulemd-module-private.h
+++ b/modulemd/v2/include/modulemd-2.0/private/modulemd-module-private.h
@@ -96,7 +96,10 @@ modulemd_module_get_translation (ModulemdModule *self, const gchar *stream);
 /**
  * modulemd_module_add_stream:
  * @self: This #ModulemdModule object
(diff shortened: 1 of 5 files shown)


modulemd/v2/include/modulemd-2.0/private/modulemd-module-private.h
```diff
@@ -96,7 +96,10 @@ modulemd_module_get_translation (ModulemdModule *self, const gchar *stream);
 /**
  * modulemd_module_add_stream:
  * @self: This #ModulemdModule object
- * @stream: A #ModulemdModuleStream object to associate with this #ModulemdModule.
+ * @stream: A #ModulemdModuleStream object to associate with this
+ * #ModulemdModule. A stream added to a #ModulemdModule must have a module
```
puiterwijk: Maybe here mention that it will be ignored? Since there's no error feedback other than the logged output.

PR #153, opened 2018-12-06
Handle upgrading streams and defaults from the Index

sgallagher:
This is a significant number of patches all related to being able to upgrade all the contents of a ModuleIndex to the same version.

We want the result to be that a ModuleIndex will always have all of its contents at the same version, matching the highest metadata version of each object type that has been seen. This will make it simpler to do merges between repositories, as we'll always be comparing the same versions.

I made the upgrade functions public as well, since I expect that applications like DNF will want to ensure that the content is always at the highest version they know how to process. This way, they can always travel a single codepath, even if the repodata is older (such as when we eventually release v3, we want DNF to be handling that interface well before we allow v3 actually into the repository).


sgallagher:
It looks like I introduced a really subtle memory leak somewhere in patch "Upgrade Streams when adding them to the ModuleIndex". I spotted this leak when writing the test that I included in https://github.com/fedora-modularity/libmodulemd/pull/154

I've been staring at the problem for a couple hours now; I'm going to sleep on it and see if it comes to me in the morning.

sgallagher:
To expand on the above slightly, the reason we didn't spot it before is that the memory leak only appears when we are reading in a YAML document containing more than one kind of document. If the YAML is all streams, all defaults or all translations, no memory is lost. As soon as there are any two kinds of documents in the YAML, the leak appears.

sgallagher:
Finally located the memory leak. In `modulemd_module_index_upgrade_streams()`, it called `continue;` in the loop if it encountered a Module with zero streams attached, but it didn't release the reference it had taken on that module. Added a `g_object_clear()` there and now all is right with the world.

modulemd/v2/include/modulemd-2.0/modulemd-module-index.h
```diff
@@ -234,3 +271,66 @@ gboolean
...
+
+
+/**
+ * modulemd_module_index_upgrade_defaults:
+ * @self: This #ModulemdModuleIndex
+ * @mdversion: The #ModulemdDefaults metadata version to upgrade to.
+ * @error: (out): A #GError that contains information on why the index could
+ * not be upgraded in the event of an error.
+ *
+ * Upgrades all #ModulemdDefaults objects in this index to @dmversion if they
```
puiterwijk: `dmversion` -> `mdversion`

modulemd/v2/modulemd-module.c
```diff
@@ -181,23 +181,68 @@ modulemd_module_init (ModulemdModule *self)
...
 
   g_clear_object (&self->defaults);
-
   if (defaults == NULL)
-    return;
+    {
+      /* If we are empty here, return the passed-in index_mdversion so the
+       * function reports success and does not influence further upgrades.
+       */
+      return index_mdversion;
```
puiterwijk: This does not match the documentation in the header file: `Returns MD_DEFAULTS_VERSION_UNSET if @defaults was NULL`.

I think I would suggest to go with explicitly returning `VERSION_UNSET` and handling that in the index functions, so that in the future it's easier to handle this case specifically if needed.

modulemd/v2/modulemd-module.c
```diff
@@ -181,23 +181,68 @@ modulemd_module_init (ModulemdModule *self)
...
+                   modulemd_defaults_get_module_name (defaults),
+                   modulemd_module_get_module_name (self));
+      return MD_DEFAULTS_VERSION_ERROR;
+    }
+
+  /* For Modulemd.ModuleIndex, we always want all entries to have the same
+   * version, so that merges can be performed. If this Defaults object has a
+   * lower mdversion than the Index, upgrade it to that version.
+   */
+  if (modulemd_defaults_get_mdversion (defaults) < index_mdversion)
```
puiterwijk: Note that the `modulemd_defaults_upgrade` implementation already does the copy if the mdversion is equal to the requested version, so you could simplify here by just always calling the upgrade function.

modulemd/v2/modulemd-module.c
```diff
@@ -235,7 +324,24 @@ modulemd_module_add_stream (ModulemdModule *self, ModulemdModuleStream *stream)
       old = NULL;
     }
 
-  newstream = modulemd_module_stream_copy (stream, NULL, NULL);
+  if (modulemd_module_stream_get_mdversion (stream) < index_mdversion)
```
puiterwijk: Same as `Defaults`: the `modulemd_module_stream_upgrade` function already returns a copy if `current_mdversion == mdversion`, so maybe just use that fact here?

modulemd/v2/modulemd-module.c
sgallagher: No, I can't. I'll update the comment to explain why, but it's basically this: if the version we have here is already *higher* than the index_mdversion, the upgrade routine would fail because it would be asked to downgrade.

modulemd/v2/modulemd-module.c
sgallagher: Same as above.

modulemd/v2/modulemd-module.c
sgallagher: Thanks for catching that. One too many times back-and-forth on how to handle that case. Initially, I had been setting `MD_DEFAULT_VERSION_ERROR` to be `MD_DEFAULT_VERSION_LATEST + 1`, but then I reworked it so that it would be -1, updated the documentation and then forgot to edit that location in the code.

PR #163, opened 2018-12-13
Add modulemd-validator for v2 API

sgallagher:
Relegate the v1 implementation to the name modulemd-validator-v1

### DIFF
--- a/modulemd/v2/modulemd-module-index.c
+++ b/modulemd/v2/modulemd-module-index.c
@@ -23,6 +23,7 @@
 #include "private/modulemd-defaults-v1-private.h"
 #include "private/modulemd-subdocument-info-private.h"
 #include "private/modulemd-module-index-private.h"
+#include "private/modulemd-module-stream-private.h"
 #include "private/modulemd-module-stream-v1-private.h"
 #include "private/modulemd-module-stream-v2-private.h"
 #include "private/modulemd-translation-private.h"
@@ -121,11 +122,13 @@ static gboolean
 add_subdoc (ModulemdModuleIndex *self,
             ModulemdSubdocumentInfo *subdoc,
             gboolean strict,
+            gboolean autogen_module_name,
             GError **error)
 {
   g_autoptr (ModulemdModuleStream) stream = NULL;
   g_autoptr (ModulemdTranslation) translation = NULL;
   g_autoptr (ModulemdDefaults) defaults = NULL;
+  g_autofree gchar *name = NULL;
 
   switch (modulemd_subdocument_info_get_doctype (subdoc))
     {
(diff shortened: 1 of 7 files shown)


modulemd/v2/modulemd-module-index.c
```diff
@@ -158,6 +153,32 @@ add_subdoc (ModulemdModuleIndex *self,
...
           return FALSE;
         }
+
+      if (stream == NULL)
+        return FALSE;
+
+      if (autogen_module_name &&
+          !modulemd_module_stream_get_module_name (stream))
+        {
+          name = g_strdup_printf ("__unknown_module_%d",
```
puiterwijk: Maybe instead change this to `__unnamed_module_%d` or `__nameless_module_%d`, to indicate it's actually just a module without a name?
This might be less confusing to people who are trying to figure out why their unnamed stream is invalid.

modulemd/v2/modulemd-module-index.c
```diff
@@ -201,10 +222,11 @@ add_subdoc (ModulemdModuleIndex *self,
 }
 
 
-static gboolean
+gboolean
```
puiterwijk: If we're making this a function that's callable from outside of this file (even if private), I think that it should have the `failures` initialization as well that are in the public functions.

modulemd/v2/modulemd-validator.c
```diff
@@ -0,0 +1,213 @@
...
+    {
+      g_fprintf (stderr,
+                 "At least one file must be specified on the command-line\n");
+      return EXIT_FAILURE;
+    }
+
+  for (gsize i = 0; options.filenames[i]; i++)
+    {
+      filename = options.filenames[i];
+      failures = g_ptr_array_new_with_free_func (g_object_unref);
```
puiterwijk: As mentioned in the index code, I think that if we make that function available from "outside", this should also probably be initialized by the `update_from_parser` function.

modulemd/v2/modulemd-validator.c
```diff
@@ -0,0 +1,213 @@
...
+          num_invalid++;
+          if (options.verbosity >= MMD_DEFAULT)
+            {
+              g_fprintf (stderr, "%s failed to validate\n", filename);
+
+              if (error != NULL)
+                {
+                  /* Unparseable content */
+                  g_fprintf (
+                    stderr, "%s contained unparseable YAML\n", filename);
```
puiterwijk: Maybe also print the `error`?

modulemd/v2/modulemd-module-index.c
sgallagher: Sure, I can paint the shed that color.

modulemd/v2/modulemd-module-index.c
sgallagher: I avoided doing that because it would involve non-trivial changes to the other places it's called (and for little benefit). Everywhere else this function is called, it's appending to a failures list that may already have content. To change it would mean adding code to all the callers to iterate it and append its contents to another list. That's not impossible, but I'm not sure it's a valid use of time.

I documented in the header that @failures needed to be pre-initialized before calling this function.

modulemd/v2/modulemd-validator.c
sgallagher: I probably should, but the error is going to be "Parse Error" pretty much 100% of the time. I suppose I should be future-compatible in case it becomes possible for a different error to come back.

modulemd/v2/modulemd-module-index.c
puiterwijk: As discussed on IRC, I seem to have been vague in my comment.

I meant that you might just add what e.g. `update_from_stream` has:
````
if (*failures == NULL)                                 
  *failures = g_ptr_array_new_full (0, g_object_unref);
````

That should make it not change anything for other callers.

sgallagher: OK, that seems reasonable. I'll do that.

sgallagher:
I pushed several fixes:
* I made it so that `failures` is allocated by `modulemd_module_index_update_from_parser()` if it's NULL
* I provided the error message if the file could not be parsed.
* I fixed several memory leaks, including initializing the parser twice.

sgallagher:
I made the changes as a fixup commit for ease of review which I will squash before merging to `2.0-dev`.

PR #164, opened 2018-12-13
[DO_NOT_MERGE] Test that each individual commit compiles

sgallagher:
In order to ensure that future attempts to use `git bisect` will
work, we need to ensure that every commit to the repository can
pass `ninja test` individually.


sgallagher:
Well, this experiment failed.

PR #167, opened 2019-01-07
Always sort the dependencies

sgallagher:
This will make sure that the output is always consistent. This
patch fixes both the ordering of the module names as well as the
stream names.


sgallagher:
The remaining test failure on Rawhide appears to be due to a new leak in glib2 2.59.

sgallagher:
> The remaining test failure on Rawhide appears to be due to a new leak in glib2 2.59.

Reported as https://bugzilla.redhat.com/show_bug.cgi?id=1664102

PR #172, opened 2019-01-16
Don't fail merges when default streams differ

sgallagher:
Instead of failing the merge on a default stream conflict, we will
instead treat it as having no default set. This is safe because if
the module is not yet installed, then this just means its packages
aren't visible and it needs to be selected explicitly.

If some packages from it are already installed, then the module
stream is already set, and thus changing the default won't matter.

Resolves: rhbz#[1666871](https://bugzilla.redhat.com/show_bug.cgi?id=1666871)

### DIFF
--- a/modulemd/v1/include/modulemd-1.0/private/modulemd-private.h
+++ b/modulemd/v1/include/modulemd-1.0/private/modulemd-private.h
@@ -26,6 +26,8 @@ enum
 
 #define MD_VERSION_LATEST MD_VERSION_2
 
+#define DEFAULT_MERGE_CONFLICT "__merge_conflict__"
+
 ModulemdModule *
 modulemd_module_new_from_modulestream (ModulemdModuleStream *stream);
 
--- a/modulemd/v1/modulemd-defaults.c
+++ b/modulemd/v1/modulemd-defaults.c
@@ -14,6 +14,7 @@
 #include "modulemd.h"
 #include "modulemd-defaults.h"
 #include "modulemd-simpleset.h"
+#include "private/modulemd-private.h"
 #include "private/modulemd-yaml.h"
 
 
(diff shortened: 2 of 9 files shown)


contyk:
Could this emit a warning somehow? Unless it already does and I just don't see it.

sgallagher:
Hmm, tests are passing locally and failing in Travis CI... time to do a little digging.

> Resolves: rhbz#1666871 <https://bugzilla.redhat.com/show_bug.cgi?id=1666871>
> I would prefer clickable link.

It's a clickable link on the web UI. Maybe not in the email?

sgallagher:
> Could this emit a warning somehow? Unless it already does and I just don't see it.

See line https://github.com/fedora-modularity/libmodulemd/pull/172/commits/c57161dad794a1fa14f9b2871902d2711234c81b#diff-631d0fbdef06e80016ba77f8c7b32305R787 (which I just realized I didn't replicate in the 1.0 API, so I'll fix that while I figure out what's up with the CI).

sgallagher:
Ah, forgot to include a file. That explains it.

> > Could this emit a warning somehow? Unless it already does and I just don't see it.
> 
> See line [c57161d#diff-631d0fbdef06e80016ba77f8c7b32305R787](https://github.com/fedora-modularity/libmodulemd/commit/c57161dad794a1fa14f9b2871902d2711234c81b#diff-631d0fbdef06e80016ba77f8c7b32305R787) (which I just realized I didn't replicate in the 1.0 API, so I'll fix that while I figure out what's up with the CI).

No, I was looking at just one of the two commits. It's in both.

The test failure is due to a missing file, which I'm pushing now.

sgallagher:
@contyk Regarding a warning: the best I can do is print a message to the log if the logging is enabled for libmodulemd. I can't send a warning back to the consumer (e.g. DNF) without changing the semantics, because the API contract right now says you only get a GError if the function fails. I'd have to change API and ABI to allow that.

sgallagher:
Okay, tests passed this time. I'll wait for a review from @contyk before pushing.

ignatenkobrain:
You should not send warnings back through API, just send to logging.

sgallagher:
Ping @contyk or @puiterwijk for a review

Issue #187, opened 2019-02-14
Provide method which compares two Defaults objects


sgallagher:
This should be a fairly easy (but not quick) addition to the 2.0 API. What we'll need:

Modulemd.Defaults (the parent class) will add a .equals() virtual function that will be implemented by Modulemd.DefaultsV1. Its responsibility will be to recurse down into the various objects and variables making up the Defaults and perform an equality comparison on them. If any of them are different, it should return false.

As an implementation recommendation, the child objects should "chain up" to the parent class to compare the pieces common to all.

nfatma:
Hello!
I am an Outreachy applicant, I would like to work on this issue.

sgallagher:
Please go right ahead!

I've also just added a new section to the README.md to help you get started setting up your development environment. Please let me know if you have any difficulty with getting started.

OrionStar25:
> As an implementation recommendation, the child objects should "chain up" to the parent class to compare the pieces common to all.

Can you please explain what this means?

sgallagher:
@OrionStar25 Sure! What it means is that for objects that are derived from other objects (such as the `Modulemd.DefaultsV1` object being derived from `Modulemd.Defaults`), it's best to have common code stored with the parent object.

In this case, it means that for attributes like "module_name", "mdversion" and "modified", the comparison should be done in a routine in `Modulemd.Defaults`, while comparisons of "default_stream" and profile data should be done in the `Modulemd.DefaultsV1` object.

For example, see https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/modulemd-module-stream-v2.c#L764 for how we "chain up" in the `Modulemd.ModuleStreamV2.validate()` routine. It calls up to `Modulemd.ModuleStream.validate()` to verify the bits that are common to all `Modulemd.ModuleStream` objects and then proceeds further to validate the bits that are specific to `Modulemd.ModuleStreamV2`.

Let me know if that helps or if you need more guidance here.

OrionStar25:
Thats enough to help me start! I'll open a PR soon

PR #200, opened 2019-02-27
[WIP] Equals function for ModuleStream objects

indigoOmega021:


modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
```diff
@@ -78,6 +78,8 @@ struct _ModulemdModuleStreamClass
                                        const gchar *module_name,
                                        const gchar *stream_name);
 
+  gboolean (*equals) (ModulemdModuleStream *self, ModulemdModuleStream *other);
+
   /* Padding to allow adding up to 8 new virtual functions without
    * breaking ABI. */
   gpointer padding[8];
```
sgallagher: When adding a new virtual function, you need to decrement the padding size. The reason for this is so that we don't break ABI compatibility; the size of the struct doesn't change. If you just add a new function without reducing this padding value, the struct would increase in size by one pointer and any application compiled against the previous version would be forced to rebuild to understand the new size.

The padding is there to avoid the rebuild when adding new functions.

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
```diff
@@ -240,8 +242,6 @@ modulemd_module_stream_upgrade (ModulemdModuleStream *self,
 
 
 /**
- * modulemd_module_stream_validate:
- * @self: (in): This #ModulemdModuleStream.
```
sgallagher: Why are you deleting these lines? I'm guessing this was a mistake. It will break the documentation-generation for this function.

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
```diff
@@ -395,4 +395,16 @@ gboolean
 modulemd_module_stream_build_depends_on_stream (ModulemdModuleStream *self,
                                                 const gchar *module_name,
                                                 const gchar *stream_name);
+
+/**
+ *modulemd_module_stream_equals:
```
sgallagher: You need to leave a space between the asterisk (*) and the content of the comment or the documentation parser will have problems.

modulemd/v2/modulemd-module-stream.c
```diff
@@ -990,3 +990,15 @@ modulemd_module_stream_build_depends_on_stream (ModulemdModuleStream *self,
...
+
+    
+static gboolean
+modulemd_module_stream_default_equals (ModulemdModuleStream *self, ModulemdModuleStream *other)
+{
+  g_return_val_if_fail (MODULEMD_IS_MODULE_STREAM (self), FALSE);
+
+  guint64 mdversion_self = modulemd_module_stream_get_module_name (self);
+  guint64 mdversion_other = modulemd_module_stream_get_module_name (other);
+  g_return_val_if_fail ((mdversion_self == mdversion_other), FALSE);
```
sgallagher: Don't use `g_return_val_if_fail()` for proper comparisons. This is actually an `assert()` under the hood. It's used in places like
```
g_return_val_if_fail (MODULEMD_IS_MODULE_STREAM (self), FALSE);
```
in order to catch programming mistakes. When compiled in debug mode, this will halt execution rather than returning FALSE. When compiled in release mode, it will log a loud error to the system log and return FALSE.

You want to do:
```
  if (mdversion_self != mdversion_other) return FALSE;
```

indigoOmega021: Should I replace line 998 with an if statement that returns FALSE if self or other is NULL?

sgallagher: No, that one is correct. That should be an assertion that a programmer did not accidentally pass in an incorrect variable type (like accidentally passing a `Modulemd.Module` instead of a `Modulemd.ModuleStream`).

The tests for equality are giving you the actual requested answer for this function: if they're not equal, the objects cannot be equal.

Oh, one other tip: include a test at the top of each equality function for whether `self == other`. If, for any reason, you've been passed the exact same pointer address, you can short-circuit all of the other comparisons and return TRUE, since it's not just equivalent: it's the same location in memory.

sgallagher: I'm not sure if I was clear enough above, so let me try to explain. The purpose of `g_return_val_if_fail()` is that it is intended for non-standard reasons to exit. Essentially, it's a side-effect. We call this when we're making sure that the person who wrote the code did not make a mistake with what they passed in.

We want to use standard return values for any situation where the return value is the correct answer to whatever question the function is supposed to answer.

In this case, the function is supposed to answer: "Do these two `Modulemd.ModuleStream` instances have equivalent values for all of their attributes?". If we're comparing their mdversions or their stream name, that's answering that question. If, however, we're just making sure that we were passed real `Modulemd.ModuleStream` objects, that's not answering the same question. So we treat that as a non-standard error and use `g_return_val_if_fail()`

sgallagher:
@indigoOmega021 You appear to have added a huge number of unrelated (and large!) files to your commit. These need to be excluded. They don't belong in git and also make it impossible to review your patch.

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
```diff
@@ -395,4 +397,17 @@ gboolean
 modulemd_module_stream_build_depends_on_stream (ModulemdModuleStream *self,
                                                 const gchar *module_name,
                                                 const gchar *stream_name);
+
+
+/**
+ * modulemd_module_stream_equals:
+ * @self: (not nullable): This #ModulemdModuleStream.
+ *
+ * Returns: TRUE if all of the objects and varibles composing the two
```
sgallagher:
```suggestion
 * Returns: TRUE if all of the objects and variables composing the two 
```

modulemd/v2/modulemd-module-stream.c
```diff
@@ -990,3 +990,43 @@ modulemd_module_stream_build_depends_on_stream (ModulemdModuleStream *self,
 
   return klass->build_depends_on_stream (self, module_name, stream_name);
 }
+
+    
+static gboolean
+modulemd_module_stream_default_equals (ModulemdModuleStream *self, ModulemdModuleStream *other)
+{
+  g_return_val_if_fail (MODULEMD_IS_MODULE_STREAM (self), FALSE);
```
sgallagher: Test that both inputs are of the correct object type.

```suggestion
  g_return_val_if_fail (MODULEMD_IS_MODULE_STREAM (self), FALSE);
  g_return_val_if_fail (MODULEMD_IS_MODULE_STREAM (other), FALSE);
```

sgallagher:
> @sgallagher Would you mind reviewing the changes that I made before I move on to writing equals functions for v1 and v2.

I've already done so. See my replies above. This code looks quite good, but see my comment about the next steps for this object before you move on to V1 and V2.

sgallagher:
@indigoOmega021 You don't need to credit me for my code-review comments. You're doing the work, you get the credit (and blame!) in git.

Also, please be aware that before we merge this upstream, I'm going to ask you to squash your patches together. This is probably a good place to do that for the first time, since we've pretty much approved this piece of it.

One policy we have on the project that I may not have noted: every patch committed upstream must pass the tests and we don't want to carry around a lot of work-in-progress patches. In particular, this is to ensure that tools like `git bisect` can help us track down when a bug was added. If any of the commits do not build, we can't use that technique.

So, do a `git rebase -i` and merge your patches together.

indigoOmega021:
@sgallagher I am trying to implement the equals function in V2 and I am confused about how to chain up the function to the original one I wrote by calling it in the v2 version of the equals function. Could you provide any guidance on how to properly do this?

sgallagher:
> @sgallagher I am trying to implement the equals function in V2 and I am confused about how to chain up the function to the original one I wrote by calling it in the v2 version of the equals function. Could you provide any guidance on how to properly do this? Does what I have so far seem like a good start?

Have a look at https://github.com/fedora-modularity/libmodulemd/issues/187#issuecomment-468705134 which was my response to the student attempting to do the same for the Defaults objects. It includes an example of chaining up for the `validate()` function which you can use as a guideline.

indigoOmega021:
@sgallagher I am having trouble comparing two gobjects from the ModulemdBuildopts I receive from the getter in Modulemd.ModuleStreamV2. The gobject library doesn't provide an equals function and I do not know what fields the gobjects may have or how to compare them. Any advice would be appreciated.

sgallagher:
@indigoOmega021 You're going to need to create the equality function for ModulemdBuildopts first. In hindsight, I probably should have gone through and created individual issue tickets for each of the sub-objects and then marked this one as depending on those.

I'll get those created in a few minutes.

sgallagher:
Please edit the title of this PR so it's possible to identify what it's intended to accomplish. There are now multiple PRs titled "[WIP]" in the list and I can't tell them apart at a glance.

sgallagher:
@indigoOmega021 Are you still working on this PR?

indigoOmega021:
@sgallagher I am still working on it. I haven't been as active because of Spring Break and the pile up of school work after. I will make some progress this week.

Issue #192, opened 2019-02-15
Extend tests for ModuleStream.depends_on_stream() to cover empty stream lists


sgallagher:
The tests for `depends_on_stream()` and `build_depends_on_stream()` should validate that it handles the case of an empty list of streams properly.

As per the specification, the behavior here must be that an empty list means "all possible values".

athira-selvam:
Hi,
I am an Outreachy applicant, I would like to work on this project. Shall work on this issue?
 - Thank you.

sgallagher:
Please go right ahead!

I've also just added a new section to the README.md to help you get started setting up your development environment. Please let me know if you have any difficulty with getting started.

athira-selvam:
Hi,
I did set up the development environment. I need some help. Is modulestream.py -  the file that I have to add the tests to?  I found some tests for `depends_on_stream()` and `build_depends_on_stream()` in that file. Just wanted confirmation before I proceed.

sgallagher:
> Hi,
> I did set up the development environment. I need some help. Is modulestream.py - the file that I have to add the tests to? I found some tests for `depends_on_stream()` and `build_depends_on_stream()` in that file. Just wanted confirmation before I proceed.

Yes, those are the tests I meant. You can choose to write them in either the Python tests or into new C tests in [modulemd/v2/tests/test-modulemd-modulestream.c](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/tests/test-modulemd-modulestream.c) (related: #199)

Issue #199, opened 2019-02-26
Copy python tests for Modulemd.ModuleStream into C


sgallagher:
Currently, most of the tests for the ModuleStream objects are run under Python. The valgrind tests for identifying memory leaks and errors can only run against tests in the C code. We need to replicate the python tests as C tests.

This would be a good task for a newcomer to the project. Look at the tests being run in [`modulemd/v2/tests/ModulemdTests/modulestream.py`](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/tests/ModulemdTests/modulestream.py) and copy them into [`modulemd/v2/tests/test-modulemd-modulestream.c`](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/tests/test-modulemd-modulestream.c)

This will ensure that these code-paths are scanned with valgrind for leaks when `ninja test` is run (and when the CI test suite is executed).

SunitiJain:
hi ! I want to start contributing in this project. Please guide me how to approach.
Thanks in advance.

swagatikapanda123:
hello!! my name is swagatika . I am facing trouble in how to find an issue in the project. could you please guide me how to find an issue..

athira-selvam:
Is @swagatikapanda123 working on this? 
@sgallagher  Shall I try to copy some more functions from python to C?

athira-selvam:
Hi,
Is anyone working on this Issue?

swagatikapanda123:
yes !!

OrionStar25:
@sgallagher Currently, these tests are left to copy:

1. [test_buildopts](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/tests/ModulemdTests/modulestream.py#L274)
2. [v1_eol](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/tests/ModulemdTests/modulestream.py#L508)
3. [v1_yaml](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/tests/ModulemdTests/modulestream.py#L899)

Since `v1` is deprecated, are 2 and 3 required to copy?

sgallagher:
@OrionStar25 I'd prefer if they were, because it allows us to run memory checks and static analysis scans against those codepaths. Thank you (and your mentees) very much for your work on this!

PR #203, opened 2019-02-28
Extends tests for ModuleStream.depends_on_stream()

athira-selvam:
The tests for depends_on_stream() and build_depends_on_stream() is modified to handle empty lists
Replicates the existing python tests, and newly added tests for depends_on_stream to C.
Fixes : https://github.com/fedora-modularity/libmodulemd/issues/192
related:  https://github.com/fedora-modularity/libmodulemd/issues/199

### DIFF
--- a/modulemd/v2/tests/ModulemdTests/modulestream.py
+++ b/modulemd/v2/tests/ModulemdTests/modulestream.py
@@ -682,7 +682,6 @@ data:
 
         assert 'rawhide' in stream.get_servicelevel_names()
         assert 'production' in stream.get_servicelevel_names()
-
         sl = stream.get_servicelevel('rawhide')
         assert sl is not None
         assert sl.props.name == 'rawhide'
@@ -1035,8 +1034,16 @@ data:
                 stream.build_depends_on_stream(
                     'platform', 'f28'), False)
 
-            self.assertEqual(stream.depends_on_stream('base', 'f30'), False)
-            self.assertEqual(stream.depends_on_stream('base', 'f30'), False)
+            self.assertEqual(
+                stream.depends_on_stream(
+                    'base', ''), False)
+            self.assertEqual(
+                stream.build_depends_on_stream(
+                    'base', ''), False)
+
+            ret = stream.build_depends_on_stream(
+                'platform', '')
+            assert (ret is True or ret is False)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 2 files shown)


modulemd/v2/tests/ModulemdTests/modulestream.py
```diff
@@ -1035,8 +1034,16 @@ def test_depends_on_stream(self):
                 stream.build_depends_on_stream(
                     'platform', 'f28'), False)
 
-            self.assertEqual(stream.depends_on_stream('base', 'f30'), False)
-            self.assertEqual(stream.depends_on_stream('base', 'f30'), False)
```
sgallagher: Why are you removing these two tests? Though, the second one should have been `stream.build_depends_on_stream()`, That's a bug in the original code here.

modulemd/v2/tests/ModulemdTests/modulestream.py
```diff
@@ -682,7 +682,6 @@ def test_v2_yaml(self):
 
         assert 'rawhide' in stream.get_servicelevel_names()
         assert 'production' in stream.get_servicelevel_names()
-
```
sgallagher: Why delete this line?

modulemd/v2/tests/ModulemdTests/modulestream.py
```diff
@@ -1035,8 +1034,16 @@ def test_depends_on_stream(self):
                 stream.build_depends_on_stream(
                     'platform', 'f28'), False)
 
-            self.assertEqual(stream.depends_on_stream('base', 'f30'), False)
-            self.assertEqual(stream.depends_on_stream('base', 'f30'), False)
+            self.assertEqual(
+                stream.depends_on_stream(
+                    'base', ''), False)
```
sgallagher: I think you misunderstood the requirement here. We don't need tests for a stream name being the empty string. We need a new requires/buildrequires in the v2 YAML document. Something like `perl` in:

```
  dependencies:
  - buildrequires:
      platform: [ f30 ]
      perl: []
    requires:
      platform: [ -f28 ]
      perl: []
```

This test is ONLY for ModuleStreamV2, since there is no way to represent this in the ModuleStreamV1 YAML.

sgallagher:
Tracking this in #210 

Please do not file new PRs each time. Just update the current one by pushing to this branch.

PR #210, opened 2019-03-06
Extends tests for ModuleStream.depends_on_stream()

athira-selvam:
The tests for depends_on_stream() and build_depends_on_stream() is modified to handle empty lists

issue: https://github.com/fedora-modularity/libmodulemd/issues/192

### DIFF
--- a/modulemd/v2/tests/ModulemdTests/modulestream.py
+++ b/modulemd/v2/tests/ModulemdTests/modulestream.py
@@ -1085,7 +1085,17 @@ data:
                     'platform', 'f28'), False)
 
             self.assertEqual(stream.depends_on_stream('base', 'f30'), False)
-            self.assertEqual(stream.depends_on_stream('base', 'f30'), False)
+            self.assertEqual(
+                stream.build_depends_on_stream(
+                    'base', 'f30'), False)
+
+            if version >= Modulemd.ModuleStreamVersionEnum.TWO:
+                self.assertEqual(
+                    stream.depends_on_stream(
+                        'streamname', 'f30'), True)
+                self.assertEqual(
+                    stream.build_depends_on_stream(
+                        'streamname', 'f30'), True)
 
 
 if __name__ == '__main__':
--- a/modulemd/v2/tests/test_data/dependson_v2.yaml
+++ b/modulemd/v2/tests/test_data/dependson_v2.yaml
@@ -14,8 +14,10 @@ data:
   dependencies:
   - buildrequires:
       platform: [ f30 ]
+      streamname: [ ] 
     requires:
       platform: [ -f28 ]
+      streamname: [ ]     
   references:
     community: http://nodejs.org
     documentation: http://nodejs.org/en/docs


modulemd/v2/tests/test_data/dependson_v2.yaml
```diff
@@ -14,8 +14,10 @@ data:
   dependencies:
   - buildrequires:
       platform: [ f30 ]
+      streamname: []
```
sgallagher: You added this field but didn't add a test for `stream.build_depends_on_stream('streamname', 'f30', True)

modulemd/v2/tests/test_data/dependson_v2.yaml
```diff
@@ -14,8 +14,10 @@ data:
   dependencies:
   - buildrequires:
       platform: [ f30 ]
+      streamname: [] 
     requires:
       platform: [ -f28 ]
+      streamname: []
```
sgallagher: Same with this one, except `stream.depends_on_stream()`.

modulemd/v2/tests/test_data/dependson_v2.yaml
athira-selvam: self.assertEqual(
                stream.build_depends_on_stream(
                    'streamname', 'f30'), False)
ninja test is failing when I give this test.
It fails when I give True also. What should be the condition to check with? Is it like it could be Either true or false?

sgallagher: The empty set ( `[]` ) should mean that any stream name should be accepted. It's possible that you have found a bug in the code (which is the purpose of writing new tests!).

The expected behavior is that `stream.build_depends_on_stream('streamname', 'f30'), True)` should be successful here. If it's not, we should look into the code this is exercising and try to figure out where it's going wrong.

athira-selvam: Is this the file to be modified to fix that bug:
[link](https://github.com/fedora-modularity/libmodulemd/blob/4e60ff2b1194b7da8648517411d48a4f4b7ae471/modulemd/v2/modulemd-module-stream-v2.c) 
Can you give me a lead to fix it.

sgallagher: @athira-selvam That is indeed the correct location, but I don't see anything obviously wrong with the code. Please push your work-in-progress to this PR and I'll see if there's a mistake in your test.

athira-selvam: I  have pushed my work-in-progress.

sgallagher: I found the problem: you are running this test on both the `Modulemd.ModuleStreamV1` and `Modulemd.ModuleStreamV2` objects.

There is no `[ ]` implementation for V1, so you can't use this test there. Wrap this test in a check to make sure it only runs on V2 or later:

```
if (version >= Modulemd.ModuleStreamVersionEnum.TWO):
```

athira-selvam:
I accidentally did git add --all instead of git add {filename}. I pushed that code. Is it possible to revert my push?

sgallagher:
@athira-selvam You have to do a new push. Probably the easiest way is to do `git rm -r api2 && git commit --amend && git push --force`

modulemd/v2/tests/ModulemdTests/modulestream.py
```diff
@@ -1036,7 +1036,26 @@ def test_depends_on_stream(self):
                     'platform', 'f28'), False)
 
             self.assertEqual(stream.depends_on_stream('base', 'f30'), False)
-            self.assertEqual(stream.depends_on_stream('base', 'f30'), False)
+            self.assertEqual(
+                stream.build_depends_on_stream(
+                    'base', 'f30'), False)
+
+<<<<<<< HEAD
```
sgallagher: Looks like you have a merge conflict that you didn't fix here.

athira-selvam: Fixed it : 
[commit_link](https://github.com/fedora-modularity/libmodulemd/pull/210/commits/91b17b6c7662fc1d1ba84c041c971b7af1479ece)

modulemd/v2/tests/test_data/dependson_v2.yaml
```diff
@@ -13,9 +13,12 @@ data:
     - MIT
   dependencies:
   - buildrequires:
-      platform: [ f30 ]
+      platform: [f30]
```
sgallagher: Please don't remove these spaces. They make it more readable.
```suggestion
      platform: [ f30 ]
```

athira-selvam:
I have merged all the commits.

athira-selvam:
@sgallagher  Shall I work on the file to include:
1. stream.build_depends_on_stream('platform', '-f28') - true
2. Some tests for modulemd_dependencies_validate_deps()
Can you guide me on how to identify issues to work on.
What I did now was to compare the function implementation with the test file and see if all cases are included. Is it right way to do it?

PR #213, opened 2019-03-06
Fixes #187:  Provide method which compares two Defaults objects 

### LINKED ISSUE #187: Provide method which compares two Defaults objects

OrionStar25:
Fixes #187 

I am getting confused as to what is defined, and what all needs to be checked. How is `validate` different from `equals`? Can you please help me with this?

### DIFF
--- a/modulemd/v2/modulemd-defaults-v1.c
+++ b/modulemd/v2/modulemd-defaults-v1.c
@@ -26,7 +26,8 @@ struct _ModulemdDefaultsV1
 
   gchar *default_stream;
 
-  /* @key: stream name @value: GHashTable set of profile names
+  /* @key: stream name
+   * @value: GHashTable set of profile names
    */
   GHashTable *profile_defaults;
 
@@ -44,6 +45,57 @@ struct _ModulemdDefaultsV1
 G_DEFINE_TYPE (ModulemdDefaultsV1,
                modulemd_defaults_v1,
                MODULEMD_TYPE_DEFAULTS)
+
+
+static gboolean
+modulemd_defaults_v1_equals (ModulemdDefaults *self_1,
+                             ModulemdDefaults *self_2)
+{
+  ModulemdDefaultsV1 *v1_self_1 = NULL;
+  ModulemdDefaultsV1 *v1_self_2 = NULL;
+
+  g_return_val_if_fail (MODULEMD_IS_DEFAULTS_V1 (self_1), FALSE);
+  v1_self_1 = MODULEMD_DEFAULTS_V1 (self_1);
+  g_return_val_if_fail (MODULEMD_IS_DEFAULTS_V1 (self_2), FALSE);
+  v1_self_2 = MODULEMD_DEFAULTS_V1 (self_2);
+
+  if (!MODULEMD_DEFAULTS_CLASS (modulemd_defaults_v1_parent_class)
+         ->equals (self_1, self_2))
+    {
+      return FALSE;
+    }
+
+  if (g_strcmp0 (v1_self_1->default_stream, v1_self_2->default_stream) != 0)
+    {
+      return FALSE;
+    }
+
+  /*Check profile_defaults: size, keys, values*/
+  if (!modulemd_hash_table_equals (v1_self_1->profile_defaults,
+                                   v1_self_2->profile_defaults,
+                                   modulemd_hash_table_sets_are_equal))
+    {
+      return FALSE;
+    }
+
+  /*Check intent_default_streams: size, keys, values*/
+  if (!modulemd_hash_table_equals (v1_self_1->intent_default_streams,
+                                   v1_self_2->intent_default_streams,
+                                   g_str_equal))
+    {
+      return FALSE;
+    }
+
+  /*Check intent_default_profiles: size, keys, values*/
+  if (!modulemd_hash_table_equals (v1_self_1->intent_default_profiles,
+                                   v1_self_2->intent_default_profiles,
+                                   modulemd_hash_table_sets_are_equal))
+    {
+      return FALSE;
+    }
+
+  return TRUE;
+}
 
 
 ModulemdDefaultsV1 *
@@ -440,6 +492,7 @@ modulemd_defaults_v1_class_init (ModulemdDefaultsV1Class *klass)
   defaults_class->copy = modulemd_defaults_v1_copy;
   defaults_class->get_mdversion = modulemd_defaults_v1_get_mdversion;
   defaults_class->validate = modulemd_defaults_v1_validate;
+  defaults_class->equals = modulemd_defaults_v1_equals;
 }
 
 
--- a/modulemd/v2/modulemd-defaults.c
+++ b/modulemd/v2/modulemd-defaults.c
@@ -44,6 +44,47 @@ enum
 static GParamSpec *properties[N_PROPS];
 
 
+gboolean
+modulemd_defaults_equals (ModulemdDefaults *self_1, ModulemdDefaults *self_2)
+{
+  ModulemdDefaultsClass *klass;
+
+  if (!self_1 && !self_2)
+    return TRUE;
+
+  if (!self_1 || !self_2)
+    return FALSE;
+
+  g_return_val_if_fail (MODULEMD_IS_DEFAULTS (self_1), FALSE);
+  g_return_val_if_fail (MODULEMD_IS_DEFAULTS (self_2), FALSE);
+
+  klass = MODULEMD_DEFAULTS_GET_CLASS (self_1);
+  g_return_val_if_fail (klass->equals, FALSE);
+
+  return klass->equals (self_1, self_2);
+}
+
+
+static gboolean
+modulemd_defaults_default_equals (ModulemdDefaults *self_1,
+                                  ModulemdDefaults *self_2)
+{
+  if (g_strcmp0 (modulemd_defaults_get_module_name (self_1),
+                 modulemd_defaults_get_module_name (self_2)) != 0)
+    return FALSE;
+
+  if (modulemd_defaults_get_modified (self_1) !=
+      modulemd_defaults_get_modified (self_2))
+    return FALSE;
+
+  if (modulemd_defaults_get_mdversion (self_1) !=
+      modulemd_defaults_get_mdversion (self_2))
+    return FALSE;
+
+  return TRUE;
+}
+
+
 ModulemdDefaults *
 modulemd_defaults_new (guint64 mdversion, const gchar *module_name)
 {
@@ -320,6 +361,7 @@ modulemd_defaults_class_init (ModulemdDefaultsClass *klass)
 
   klass->copy = modulemd_defaults_default_copy;
   klass->validate = modulemd_defaults_default_validate;
+  klass->equals = modulemd_defaults_default_equals;
 
   properties[PROP_MDVERSION] = g_param_spec_uint64 (
     "mdversion",
--- a/modulemd/v2/modulemd-util.c
+++ b/modulemd/v2/modulemd-util.c
@@ -171,6 +171,54 @@ modulemd_hash_table_sets_are_equal (GHashTable *a, GHashTable *b)
   return TRUE;
 }
 
+/* Processes the set of keys first, then calls a unique compare function to handle any kind of value*/
+gboolean
+modulemd_hash_table_equals (GHashTable *a,
+                            GHashTable *b,
+                            GEqualFunc compare_func)
+{
+  g_autoptr (GPtrArray) set_a = NULL;
+  g_autoptr (GPtrArray) set_b = NULL;
+  int i = 0;
+  gchar *key = NULL;
+  gchar *value_a = NULL;
+  gchar *value_b = NULL;
+
+  /*Check size*/
+  if (g_hash_table_size (a) != g_hash_table_size (b))
+    {
+      return FALSE;
+    }
+
+  /*Equality check on keys of hashtable*/
+  set_a = modulemd_ordered_str_keys (a, modulemd_strcmp_sort);
+  set_b = modulemd_ordered_str_keys (b, modulemd_strcmp_sort);
+
+  for (i = 0; i < set_a->len; i++)
+    {
+      if (!g_str_equal (g_ptr_array_index (set_a, i),
+                        g_ptr_array_index (set_b, i)))
+        {
+          return FALSE;
+        }
+    }
+
+  /*Equality check for each value of all keys of hashtable*/
+  for (i = 0; i < set_a->len; i++)
+    {
+      key = g_ptr_array_index (set_a, i);
+      value_a = g_hash_table_lookup (a, key);
+      value_b = g_hash_table_lookup (b, key);
+
+      if (!compare_func (value_a, value_b))
+        {
+          return FALSE;
+        }
+    }
+
+  return TRUE;
+}
+
 
 gint
 modulemd_strcmp_sort (gconstpointer a, gconstpointer b)
--- a/modulemd/v2/tests/test-modulemd-defaults.c
+++ b/modulemd/v2/tests/test-modulemd-defaults.c
@@ -134,6 +134,36 @@ defaults_test_validate (CommonMmdTestFixture *fixture, gconstpointer user_data)
 
 
 static void
+defaults_test_equals (CommonMmdTestFixture *fixture, gconstpointer user_data)
+{
+  g_autoptr (ModulemdDefaults) defaults_1 = NULL;
+  g_autoptr (ModulemdDefaults) defaults_2 = NULL;
+
+  /*Check equality for 2 default objects with same module name and mdversion*/
+  defaults_1 = modulemd_defaults_new (MD_DEFAULTS_VERSION_ONE, "foo");
+  g_assert_true (MODULEMD_IS_DEFAULTS (defaults_1));
+  defaults_2 = modulemd_defaults_new (MD_DEFAULTS_VERSION_ONE, "foo");
+  g_assert_true (MODULEMD_IS_DEFAULTS (defaults_2));
+
+  g_assert_true (modulemd_defaults_equals (defaults_1, defaults_2));
+
+  g_clear_object (&defaults_1);
+  g_clear_object (&defaults_2);
+
+  /*Check equality for 2 default objects with different module name and same mdversion*/
+  defaults_1 = modulemd_defaults_new (MD_DEFAULTS_VERSION_ONE, "foo");
+  g_assert_true (MODULEMD_IS_DEFAULTS (defaults_1));
+  defaults_2 = modulemd_defaults_new (MD_DEFAULTS_VERSION_ONE, "bar");
+  g_assert_true (MODULEMD_IS_DEFAULTS (defaults_2));
+
+  g_assert_false (modulemd_defaults_equals (defaults_1, defaults_2));
+
+  g_clear_object (&defaults_1);
+  g_clear_object (&defaults_2);
+}
+
+
+static void
 defaults_test_upgrade (CommonMmdTestFixture *fixture, gconstpointer user_data)
 {
   g_autoptr (ModulemdDefaults) defaults = NULL;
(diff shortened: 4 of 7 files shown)


sgallagher:
`validate()` guarantees that the contents of a Defaults object is internally consistent. It is a set of tests that are run before attempting to export the object to make sure it's complete and valid.

`equals()` must take two Defaults objects, this one and another to compare it against. It then must verify that every field, recursively, in each of these objects contains equivalent information. If any of them are not the same, the function must return false.

sgallagher:
This looks like a good start. Ping me to review it again when there's a little more to it, please.

OrionStar25:
@sgallagher I would need some predefined functions too for the equality checks, same as module stream objects, right? Right now  I can only check the string values. Or are there any predefined getter functions that I haven't come across for default objects?

sgallagher:
@OrionStar25 No, the Defaults objects contain only integers, strings and hash tables of strings. So you should be able to interact with them without needing other object equality functions.

Comparing two hash tables can be tricky; you need to iterate through each one and determine if they have the same number of keys, then whether all of their keys are identical, then lastly whether each key's associated value is identical.

In the Modulemd.Default object, the profile defaults and intent_default_profiles are hash tables with a string key and a hash table as the value, so you will need to compare downwards as well. That said, be aware that any of these hash tables that are described as "sets" can be compared with the utility function `modulemd_hash_table_sets_are_equal()` in modulemd-util.h

OrionStar25:
@sgallagher I tried to iterate over the hashtables. Not entirely sure if this is on the right track, though!

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -47,6 +48,95 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+  GHashTable *intent_default_profiles_1, *intent_default_profiles_2;
+
+
+  g_return_val_if_fail (MODULEMD_IS_DEFAULTS_V1 (self_1), NULL);
+  v1_self_1 = MODULEMD_DEFAULTS_V1 (self_1);  
+  g_return_val_if_fail (MODULEMD_IS_DEFAULTS_V1 (self_2), NULL);
+  v1_self_2 = MODULEMD_DEFAULTS_V1 (self_2);  
+
+  if (!MODULEMD_DEFAULTS_CLASS (modulemd_defaults_v1_parent_class)
+             ->equals (self_1, error);)
```
sgallagher: This function takes two arguments, which should be `self_1` and `self_2`. This code won't even compile, so I'm a bit surprised you pushed it.

You don't need to call the `parent->equals()` function twice; just once with both values.

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -47,6 +48,95 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+    {
+      return FALSE;
+    }
+
+  if (!MODULEMD_DEFAULTS_CLASS (modulemd_defaults_v1_parent_class)
+             ->equals (self_2, error);)
```
sgallagher: This should not be here at all.

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -47,6 +48,95 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+
+  g_return_val_if_fail (MODULEMD_IS_DEFAULTS_V1 (self_1), NULL);
+  v1_self_1 = MODULEMD_DEFAULTS_V1 (self_1);  
+  g_return_val_if_fail (MODULEMD_IS_DEFAULTS_V1 (self_2), NULL);
+  v1_self_2 = MODULEMD_DEFAULTS_V1 (self_2);  
+
+  if (!MODULEMD_DEFAULTS_CLASS (modulemd_defaults_v1_parent_class)
+             ->equals (self_1, error);)
+    {
+      return FALSE;
```
sgallagher: Please propagate the error rather than just passing it in and back up.

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -47,6 +48,95 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+    }
+
+  if (!MODULEMD_DEFAULTS_CLASS (modulemd_defaults_v1_parent_class)
+             ->equals (self_2, error);)
+    {
+      return FALSE;
+    }
+
+  if(g_strcmp0 (v1_self_1->intent, v1_self_2->intent))
```
sgallagher: Please use `!= 0` explicitly here, just for readability.

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -47,6 +48,95 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+    {
+      return FALSE;
+    }
+
+  const gchar *default_stream_1 =
+    modulemd_defaults_v1_get_default_stream(v1_self_1, v1_self_1->intent);
```
sgallagher: Don't use `v1_self_1->intent` or expect it to exist. I didn't realize it was still in there, but it's nonfunctional and always NULL. I need to remove it. (It was part of an earlier implementation that I scrapped and must have missed that piece).

You should just compare `v1_self_1->default_stream` and `v2_self->default_stream` directly. Don't bother using the getters; this is an internal function.

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -47,6 +48,95 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+  if (g_hash_table_size (profile_defaults_1) != g_hash_table_size(profile_defaults_2))
+    {
+      return FALSE;
+    }
+  g_hash_table_iter_init (&iter_1, profile_defaults_1);
+  g_hash_table_iter_init (&iter_2, profile_defaults_2);
+
+  while (g_hash_table_iter_next (&iter_1, &key_1, &value_1) && g_hash_table_iter_next (&iter_2, &key_2, &value_2))
+    {
+      if(g_strcmp0((gchar *)key_1, (gchar *)key_1))
```
sgallagher: You're comparing `key_1` to `key_1` here. However, you can't actually do what you were trying to do because there is actually no guarantee that you will get the keys from the two hash tables in the same order.

What I'd recommend doing is first getting the ordered set of keys using `modulemd_ordered_str_keys()` and compare those. If all of the keys are identical, then you can iterate through one set of them and do `g_hash_table_lookup(table1, key)` and `g_hash_table_lookup(table2, key)` and compare the values with `modulemd_hash_table_sets_are_equal()`.

OrionStar25:
@sgallagher Please review.

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -46,6 +47,105 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
                MODULEMD_TYPE_DEFAULTS)
 
 
+static gboolean
+modulemd_defaults_v1_equals (ModulemdDefaults *self_1, ModulemdDefaults *self_2, GError **error)
```
sgallagher: This function doesn't need a GError return. Either they are identical or they are not, but it shouldn't be possible for any of the comparisons to actually fail (barring programming errors like dereferencing NULL). In the unlikely event that an error is detected, it should just return FALSE.

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -46,6 +47,105 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+{
+  ModulemdDefaultsV1 *v1_self_1 = NULL;
+  ModulemdDefaultsV1 *v1_self_2 = NULL;
+  g_autoptr (GError) nested_error = FALSE;
```
sgallagher: Remove

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -46,6 +47,105 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+  GHashTable * table_2 = NULL;
+  int i = 0;
+
+  g_return_val_if_fail (MODULEMD_IS_DEFAULTS_V1 (self_1), FALSE);
+  v1_self_1 = MODULEMD_DEFAULTS_V1 (self_1);  
+  g_return_val_if_fail (MODULEMD_IS_DEFAULTS_V1 (self_2), FALSE);
+  v1_self_2 = MODULEMD_DEFAULTS_V1 (self_2);  
+
+  if (!MODULEMD_DEFAULTS_CLASS (modulemd_defaults_v1_parent_class)
+             ->equals (self_1, self_2, error))
```
sgallagher: Drop the error

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -46,6 +47,105 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+    {
+      g_propagate_error (error, g_steal_pointer (&nested_error));
```
sgallagher: We don't need to propagate this if we drop the error return.

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -46,6 +47,105 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+
+  //Check profile_defaults: size, keys, values
+  if (g_hash_table_size (v1_self_1->profile_defaults) != g_hash_table_size(v1_self_2->profile_defaults))
+    {
+      return FALSE;
+    }
+  profile_defaults_1 = modulemd_ordered_str_keys (v1_self_1->profile_defaults, modulemd_strcmp_sort);
+  profile_defaults_2 = modulemd_ordered_str_keys (v1_self_2->profile_defaults, modulemd_strcmp_sort);
+
+  for (i = 0; i < profile_defaults_1->len; i++)
```
sgallagher: Hmm, I just realized that the implementation of `modulemd_hash_table_sets_are_equal()` is identical to this (because we only need to compare the keys for that). Probably best if we just add a `#define modulemd_hash_table_keys_are_equal modulemd_hash_table_sets_are_equal()` in private/modulemd-util.h and use that here instead. No sense duplicating the code.

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -46,6 +47,105 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+      key_1 = g_ptr_array_index (profile_defaults_1, i);
+      key_2 = g_ptr_array_index (profile_defaults_2, i);
+
+      if(g_strcmp0 (key_1, key_2) != 0)
+        {
+          return FALSE;
+        } 
+      }
+
+  for (i = 0; i < profile_defaults_1->len; i++)
```
sgallagher: This implementation is correct, but I'm wondering if we should make a generic hash-table comparison function that just takes a function pointer for a comparison function to handle the comparison of the values. This would be similar to how `modulemd_ordered_str_keys()` works; we'd want to take a `GEqualFunc ()` pointer and then implement it for each value type.

This could then be reused for any hash table equality check we might need to write, such as the intent_default_profiles just below here, as well as probably for the ModuleStream.equals() later.

modulemd/v2/modulemd-defaults.c
```diff
@@ -44,6 +51,28 @@ enum
 static GParamSpec *properties[N_PROPS];
 
 
+gboolean 
+modulemd_defaults_equals(ModulemdDefaults *self_1, ModulemdDefaults *self_2, GError **error)
+{
+  if (!self_2 || !self_1)
```
sgallagher: If they're both NULL, we should probably return TRUE, since they do match. If only one or the other is NULL, then this is correct.

modulemd/v2/modulemd-defaults-v1.c
OrionStar25: So I should make a function that first checks the size, then keys, then its corresponding values for equality. Basically wrap all of this in 1 function like:

```
if (!modulemd_hash_table_are_equal(self_1->profile_default, self_2->profile_default))
{
      return FALSE;
}

```

Also, what file will that change be in? I see only function declarations in `private/modulemd-util.h`

sgallagher: Yes, almost exactly. I'd probably call the function `modulemd_hash_table_equals()` and it will need to take three arguments (not two): `modulemd_hash_table_equals (self_1->profile_default, self_2->profile-default, profile_default_equals)`

And `profile_default_equals()` will be a function of type [`GEqualFunc()`](https://developer.gnome.org/glib/stable/glib-Hash-Tables.html#GEqualFunc), which you will use after comparing the keys to compare the values for each key. You will need to implement a unique equality function for any type of value you need to deal with.

OrionStar25:
@sgallagher I've attempted to make the asked function. I'm not sure as to where to write the function definitions so I've written it in the same file for the time being. 
Also, I'm aware that the parameters for `GEqualfunc()` should be` const void*` but then I was getting some struct error.

OrionStar25:
Shall update the other 2 PRs when the function is finalized :+1:

modulemd/v2/modulemd-util.c
```diff
@@ -172,6 +172,44 @@ modulemd_hash_table_sets_are_equal (GHashTable *a, GHashTable *b)
...
+      key_1 = g_ptr_array_index (a_1, i);
+      key_2 = g_ptr_array_index (b_1, i);
+
+      if (g_strcmp0 (key_1, key_2) != 0)
+        {
+          return FALSE;
+        }
+    }
+
+  if (!compare_func (a_1, b_1))
```
sgallagher: You can shorten the rest of this function to:
```suggestion
  return compare_func (a_1, b_1);
```

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -46,6 +47,100 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
                MODULEMD_TYPE_DEFAULTS)
 
 
+GEqualFunc profile_defaults_equals (g_autoptr (GPtrArray) profile_defaults_1,
```
sgallagher: You don't need this function at all. You're calling this equivalence within `modulemd_defaults_equals()`, which means you've already processed each of the keys. This function then processes all of the keys of the set of profile defaults and then calls a `modulemd_hash_table_sets_are_equal()` on each of the values, except those values are strings, not hash tables.

Delete this function and just pass `modulemd_hash_table_sets_are_equal()` to `modulemd_defaults_equals()` below.

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -46,6 +47,100 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+
+  if (g_strcmp0 (v1_self_1->default_stream, v1_self_2->default_stream) != 0)
+    {
+      return FALSE;
+    }
+
+  //Check profile_defaults: size, keys, values
+  if (!modulemd_hash_table_equals (v1_self_1->profile_defaults,
+                                   v1_self_2->profile_defaults,
+                                   profile_defaults_equals))
```
sgallagher:
```suggestion
                                   modulemd_hash_table_sets_are_equal))
```

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -46,6 +47,100 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+        {
+          return FALSE;
+        }
+    }
+
+  return TRUE;
+}
+
+
+GEqualFunc intent_default_profiles_equals (g_autoptr (GPtrArray)
```
sgallagher: Same for this one.

modulemd/v2/modulemd-defaults-v1.c
```diff
@@ -46,6 +47,100 @@ G_DEFINE_TYPE (ModulemdDefaultsV1,
...
+
+  //Check intent_default_profiles: size, keys, values
+  if (!modulemd_hash_table_equals (v1_self_1->intent_default_profiles,
+                                   v1_self_2->intent_default_profiles,
+                                   intent_default_profiles_equals))
+    {
+      return FALSE;
+    }
+
+  return TRUE;
```
sgallagher: You haven't done a test for `intent_default_streams`. Please add that.

modulemd/v2/modulemd-defaults.c
```diff
@@ -44,6 +45,31 @@ enum
 static GParamSpec *properties[N_PROPS];
 
 
+gboolean
```
sgallagher:
```suggestion
static gboolean
```

modulemd/v2/modulemd-defaults.c
```diff
@@ -44,6 +45,31 @@ enum
...
+modulemd_defaults_equals (ModulemdDefaults *self_1, ModulemdDefaults *self_2)
```
sgallagher: This is the parent implementation, but if you name it as the public name, you'll never call the `equals()` routine on the child classes.

```suggestion
modulemd_defaults_default_equals (ModulemdDefaults *self_1, ModulemdDefaults *self_2)
```

You need to implement a proper `modulemd_defaults_equals()` that calls this function via `klass->equals()`.

modulemd/v2/modulemd-defaults.c
```diff
@@ -320,6 +346,7 @@ modulemd_defaults_class_init (ModulemdDefaultsClass *klass)
 
   klass->copy = modulemd_defaults_default_copy;
   klass->validate = modulemd_defaults_default_validate;
+  klass->equals = modulemd_defaults_equals;
```
sgallagher:
```suggestion
  klass->equals = modulemd_defaults_default_equals;
```

OrionStar25:
I misunderstood a GHashtable set as a Ghashtable when it's just a string
array (kind of). Shall make the changes soon! In a flight rn.

OrionStar25:
So, as I have understood it this far, its this:

`profile_defaults` is a GHashtable of the form: `key:<string> -> value:<GHashtable set>`
and
`intent_default_streams` is a GHashtable of the form: `key:<string> -> value:<string>`

So, checking keys is same for all, but values for corresponding keys can be of different types, and hence we have a custom function to handle different value types for each hashtable. 

I've added some very basic test cases, but will need your feedback to move further on that.

OrionStar25:
Added some more tests. `ninja test` is failing, but I don't know how to fix it.

sgallagher:
> Added some more tests. `ninja test` is failing, but I don't know how to fix it.

As you can see from the CI test results (and would also get when running `ninja test` locally), the error is:

```
libmodulemd:ERROR:../modulemd/v2/tests/test-modulemd-defaults.c:170:defaults_test_equals: 'modulemd_defaults_equals (defaults_1, defaults_2)' should be FALSE
```

So, looking at `test-modulemd-defaults.c`, I see that you've made `defaults_1` from `MD_DEFAULTS_VERSION_ONE` and `defaults_2` from `MD_DEFAULTS_VERSION_LATEST`. However, since there's currently only a single version of the Defaults format (`1`), these are synonymous. So you are creating two identical input values and asserting that they are different.

modulemd/v2/tests/test-modulemd-defaults.c
```diff
@@ -133,6 +133,47 @@ defaults_test_validate (CommonMmdTestFixture *fixture, gconstpointer user_data)
...
+  g_assert_true (MODULEMD_IS_DEFAULTS (defaults_1));
+  defaults_2 = modulemd_defaults_new (MD_DEFAULTS_VERSION_ONE, "bar");
+  g_assert_true (MODULEMD_IS_DEFAULTS (defaults_2));
+
+  g_assert_false (modulemd_defaults_equals (defaults_1, defaults_2));
+
+  g_clear_object (&defaults_1);
+  g_clear_object (&defaults_2);
+
+  /*Check equality for 2 default objects with same module name and different mdversion*/
```
sgallagher: There are no Defaults objects with a version other than `ONE` at this time, so remove this test entirely. Trying to create one with version `0` or something `>1` will just result in a failure to create the object at all.

PR #226, opened 2019-03-09
Fixes #217: Provide Modulemd.Profile.equals() method

### LINKED ISSUE #217: Provide Modulemd.Profile.equals() method
In order to implement #186, we need to have an equality function for `Modulemd.Profile` objects.

OrionStar25:
Fixes #217 

It seemed similar to `buildopts` so decided to give it a try.

### DIFF
--- a/modulemd/v2/modulemd-profile.c
+++ b/modulemd/v2/modulemd-profile.c
@@ -46,6 +46,28 @@ enum
 static GParamSpec *properties[N_PROPS];
 
 
+gboolean
+modulemd_profile_equals (ModulemdProfile *self_1, ModulemdProfile *self_2)
+{
+  g_return_val_if_fail (MODULEMD_IS_PROFILE (self_1), FALSE);
+  g_return_val_if_fail (MODULEMD_IS_PROFILE (self_2), FALSE);
+
+  if (g_strcmp0 (modulemd_profile_get_name (self_1),
+                 modulemd_profile_get_name (self_2)) != 0)
+    return FALSE;
+
+  if (g_strcmp0 (modulemd_profile_get_description (self_1, NULL),
+                 modulemd_profile_get_description (self_2, NULL)))
+    return FALSE;
+
+  //Check rpms: size, set values
+  if (!modulemd_hash_table_sets_are_equal (self_1->rpms, self_2->rpms))
+    return FALSE;
+
+  return TRUE;
+}
+
+
 ModulemdProfile *
 modulemd_profile_new (const gchar *name)
 {
--- a/modulemd/v2/tests/test-modulemd-profile.c
+++ b/modulemd/v2/tests/test-modulemd-profile.c
@@ -81,6 +81,151 @@ profile_test_construct (ProfileFixture *fixture, gconstpointer user_data)
 
 
 static void
+profile_test_equals (ProfileFixture *fixture, gconstpointer user_data)
+{
+  g_autoptr (ModulemdProfile) p_1 = NULL;
+  g_autoptr (ModulemdProfile) p_2 = NULL;
+
+  /*Test 2 objects with same name*/
+  p_1 = modulemd_profile_new ("testprofile");
+  g_assert_nonnull (p_1);
+  g_assert_true (MODULEMD_IS_PROFILE (p_1));
+
+  p_2 = modulemd_profile_new ("testprofile");
+  g_assert_nonnull (p_2);
+  g_assert_true (MODULEMD_IS_PROFILE (p_2));
+
+  g_assert_true (modulemd_profile_equals (p_1, p_2));
+
+  g_clear_object (&p_1);
+  g_clear_object (&p_2);
+
+  /*Test 2 objects with different name*/
+  p_1 = modulemd_profile_new ("testing");
+  g_assert_nonnull (p_1);
+  g_assert_true (MODULEMD_IS_PROFILE (p_1));
+
+  p_2 = modulemd_profile_new ("testprofile");
+  g_assert_nonnull (p_2);
+  g_assert_true (MODULEMD_IS_PROFILE (p_2));
+
+  g_assert_false (modulemd_profile_equals (p_1, p_2));
+
+  g_clear_object (&p_1);
+  g_clear_object (&p_2);
+
+  /* Test 2 profile objects with same name and description */
+  p_1 = modulemd_profile_new ("testprofile");
+  modulemd_profile_set_description (p_1, "a test");
+  g_assert_nonnull (p_1);
+  g_assert_true (MODULEMD_IS_PROFILE (p_1));
+
+  p_2 = modulemd_profile_new ("testprofile");
+  modulemd_profile_set_description (p_2, "a test");
+  g_assert_nonnull (p_2);
+  g_assert_true (MODULEMD_IS_PROFILE (p_2));
+
+  g_assert_true (modulemd_profile_equals (p_1, p_2));
+
+  g_clear_object (&p_1);
+  g_clear_object (&p_2);
+
+  /* Test 2 profile objects with same name and different description */
+  p_1 = modulemd_profile_new ("testprofile");
+  modulemd_profile_set_description (p_1, "a test");
+  g_assert_nonnull (p_1);
+  g_assert_true (MODULEMD_IS_PROFILE (p_1));
+
+  p_2 = modulemd_profile_new ("testprofile");
+  modulemd_profile_set_description (p_2, "b test");
+  g_assert_nonnull (p_2);
+  g_assert_true (MODULEMD_IS_PROFILE (p_2));
+
+  g_assert_false (modulemd_profile_equals (p_1, p_2));
+
+  g_clear_object (&p_1);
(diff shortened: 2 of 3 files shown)


modulemd/v2/modulemd-profile.c
```diff
@@ -46,6 +46,50 @@ enum
...
+
+  if (g_hash_table_size (rpms_self) != g_hash_table_size(rpms_other))
+    {
+      return FALSE;
+    }
+
+  g_hash_table_iter_init (&iter_1, rpms_self);
+  g_hash_table_iter_init (&iter_2, rpms_other);
+
+  while (g_hash_table_iter_next (&iter_1, &key_1, &value_1) && g_hash_table_iter_next (&iter_2, &key_2, &value_2))
```
sgallagher: This iteration isn't going to work, in part because there is no guarantee that the keys are in the same order in each hash table. You probably need to first get the lists of keys (there's a utility function for this in modulemd-util.h) and compare those. Then, if they are identical, look up each value one at a time by the keys and compare them.

OrionStar25:
@sgallagher Please review.

modulemd/v2/modulemd-profile.c
```diff
@@ -46,6 +46,29 @@ enum
...
+  if (g_strcmp0 (modulemd_profile_get_name (self_1),
+                 modulemd_profile_get_name (self_2)) != 0)
+    return FALSE;
+
+  if (g_strcmp0 (modulemd_profile_get_description (self_1, NULL),
+                 modulemd_profile_get_description (self_2, NULL)))
+    return FALSE;
+
+  //Check rpms: size, keys, values
+  if (!modulemd_hash_table_equals (
```
sgallagher: You don't need to call `modulemd_hash_table_equals()` here, because `self->rpms` are already just sets. You can call `modulemd_hash_table_sets_are_equal (self_1->rpms, self_2->rpms)` directly here.

modulemd/v2/tests/test-modulemd-profile.c
```diff
@@ -80,6 +80,111 @@ profile_test_construct (ProfileFixture *fixture, gconstpointer user_data)
...
+  modulemd_profile_add_rpm (p_2, "testrpm");
+  g_assert_nonnull (p_2);
+  g_assert_true (MODULEMD_IS_PROFILE (p_2));
+
+  g_assert_true (modulemd_profile_equals (p_1, p_2));
+
+  g_clear_object (&p_1);
+  g_clear_object (&p_2);
+
+  /* Test 2 profile objects with same name, description, and different rpms */
```
sgallagher: Please add two more tests:

* Compare two RPM sets where the first sorted value matches and the second does not. e.g. `[ "a", "b" ]` vs `[ "a", "c" ]`.
* Compare two RPM sets where the first sorted value matches, but one has more entries than the other. e.g. `[ "a", "b" ]` vs `[  "a", "b", "c" ]`.

The first one will make sure the code doesn't accidentally return TRUE if the first one seen matches. The second one makes sure we return TRUE if they are different size sets.

OrionStar25: > The second one makes sure we return TRUE if they are different size sets.

You mean FALSE, right?

sgallagher: Oops, yes I definitely meant FALSE.

OrionStar25:
@sgallagher  I'm getting this error:

```
(/home/orionstar25/libmodulemd/api2/modulemd/v2/profile_v2:15905): libmodulemd-CRITICAL **: 00:02:40.957: modulemd_profile_set_name: assertion 'name' failed

(/home/orionstar25/libmodulemd/api2/modulemd/v2/profile_v2:15905): libmodulemd-CRITICAL **: 00:02:40.957: modulemd_profile_set_name: assertion 'g_strcmp0 (name, P_DEFAULT_STRING)' failed

(/home/orionstar25/libmodulemd/api2/modulemd/v2/profile_v2:15905): libmodulemd-CRITICAL **: 00:02:40.957: modulemd_profile_set_name: assertion 'name' failed
```
I don't understand how that's part of the code I changed

sgallagher:
> @sgallagher I'm getting this error:
> 
> ```
> (/home/orionstar25/libmodulemd/api2/modulemd/v2/profile_v2:15905): libmodulemd-CRITICAL **: 00:02:40.957: modulemd_profile_set_name: assertion 'name' failed
> 
> (/home/orionstar25/libmodulemd/api2/modulemd/v2/profile_v2:15905): libmodulemd-CRITICAL **: 00:02:40.957: modulemd_profile_set_name: assertion 'g_strcmp0 (name, P_DEFAULT_STRING)' failed
> 
> (/home/orionstar25/libmodulemd/api2/modulemd/v2/profile_v2:15905): libmodulemd-CRITICAL **: 00:02:40.957: modulemd_profile_set_name: assertion 'name' failed
> ```
> I don't understand how that's part of the code I changed

Those aren't related to your tests. The relevant issue is `(killed by signal 11 SIGSEGV)`. That means that your code *crashed* somewhere. You'll want to do `meson test --gdb profile_v2_debug` and catch the source of the segmentation fault in the debugger.

The most common reason for this is a NULL-dereference; passing a pointer to NULL and calling a function that tries to access the pointer.

sgallagher:
> > @sgallagher I'm getting this error:
> > ```
> > (/home/orionstar25/libmodulemd/api2/modulemd/v2/profile_v2:15905): libmodulemd-CRITICAL **: 00:02:40.957: modulemd_profile_set_name: assertion 'name' failed
> > 
> > (/home/orionstar25/libmodulemd/api2/modulemd/v2/profile_v2:15905): libmodulemd-CRITICAL **: 00:02:40.957: modulemd_profile_set_name: assertion 'g_strcmp0 (name, P_DEFAULT_STRING)' failed
> > 
> > (/home/orionstar25/libmodulemd/api2/modulemd/v2/profile_v2:15905): libmodulemd-CRITICAL **: 00:02:40.957: modulemd_profile_set_name: assertion 'name' failed
> > ```
> > I don't understand how that's part of the code I changed
> 
> Those aren't related to your tests. The relevant issue is `(killed by signal 11 SIGSEGV)`. That means that your code _crashed_ somewhere. You'll want to do `meson test --gdb profile_v2_debug` and catch the source of the segmentation fault in the debugger.
> 
> The most common reason for this is a NULL-dereference; passing a pointer to NULL and calling a function that tries to access the pointer.

When running `gdb`, launch the test with `run -p /modulemd/v2/profile/equals` to limit it to running the test you just added. (Otherwise, you're going to have it hit some expected SIGABRT cases and it'll confuse you.)

I just ran it myself and got:
```
(gdb) run -p /modulemd/v2/profile/equals
Starting program: /home/sgallagh/workspace/libmodulemd/api2/modulemd/v2/profile_v2 -p /modulemd/v2/profile/equals
Missing separate debuginfos, use: dnf debuginfo-install glibc-2.29-8.fc30.x86_64
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
/modulemd/v2/profile/equals: 
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff7db98a2 in g_strfreev () from /lib64/libglib-2.0.so.0
Missing separate debuginfos, use: dnf debuginfo-install glib2-2.60.0-3.fc30.x86_64 libffi-3.1-19.fc30.x86_64 libyaml-0.2.1-5.fc30.x86_64 pcre-8.43-1.fc30.x86_64
(gdb) bt
#0  0x00007ffff7db98a2 in g_strfreev () at /lib64/libglib-2.0.so.0
#1  0x0000000000402745 in glib_auto_cleanup_GStrv (_ptr=0x7fffffffc8f8) at /usr/include/glib-2.0/glib/glib-autocleanups.h:90
#2  0x0000000000404eee in profile_test_equals (fixture=0x0, user_data=0x0) at ../modulemd/v2/tests/test-modulemd-profile.c:88
#3  0x00007ffff7dc018e in  () at /lib64/libglib-2.0.so.0
#4  0x00007ffff7dbff34 in  () at /lib64/libglib-2.0.so.0
#5  0x00007ffff7dbff34 in  () at /lib64/libglib-2.0.so.0
#6  0x00007ffff7dbff34 in  () at /lib64/libglib-2.0.so.0
#7  0x00007ffff7dc063b in g_test_run_suite () at /lib64/libglib-2.0.so.0
#8  0x00007ffff7dc0695 in g_test_run () at /lib64/libglib-2.0.so.0
#9  0x0000000000408def in main (argc=1, argv=0x7fffffffcfc8) at ../modulemd/v2/tests/test-modulemd-profile.c:551
```

What this means is the autopointer for the `GStrv rpms` attempted to get freed, but it was an invalid pointer address. You didn't initialize it to NULL at line 88. If you use `g_auto()` or `g_autoptr()`, the variable must always be assigned to NULL initially or this may happen.

That said, you don't actually *use* this `rpms` variable anywhere, so you should probably just remove it entirely.

sgallagher:
Looks good to me. Please squash this into a single patch and I'll merge it. Great work!

OrionStar25:
@sgallagher Done.

Issue #186, opened 2019-02-14
Provide method which compares two ModuleStream objects


hanzz:
In MBS unit-test, we want to be able to compare two modulemd objects to ensure they are the same. We currently have to do `yaml.load(mmd1.dumps()) == yaml.load(mmd2.dumps())` to achieve that. It would be nice if there is some libmodulemd method to do the same.

sgallagher:
This should be a fairly easy (but not quick) addition to the 2.0 API. What we'll need:

`Modulemd.ModuleStream` (the parent class) will add a `.equals()` virtual function that will be implemented by `Modulemd.ModuleStreamV1` and `Modulemd.ModuleStreamV2`. Its responsibility will be to recurse down into the various objects and variables making up the ModuleStream and perform an equality comparison on them. If any of them are different, it should return false.

As an implementation recommendation, the child objects should "chain up" to the parent class to compare the pieces common to all.

ayushi1998:
Hey I am an Outreachy applicant, would like to work on this issue!

indigoOmega021:
@sgallagher virtual functions are not native to C. Do you mind further explaining what you mean by "virtual function" and possibly point to some resources on how to implement it?

sgallagher:
@indigoOmega021 The libmodulemd project is built atop the GObject toolkit. This offers us two advantages: the first is a rough approximation of object-orientation for the C language. The second is automatic introspection of these objects to create bindings for other languages such as python.

A pretty good tutorial on GObjects (including how to create virtual and pure-virtual functions) can be found at https://developer.gnome.org/gobject/stable/howto-gobject.html

In the case of libmodulemd, the parent class is located at https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/modulemd-module-stream.c and the two existing child-classes are https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/modulemd-module-stream-v1.c and https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/modulemd-module-stream-v2.c

indigoOmega021:
@sgallagher I am trying to wrap my head around what "various objects and variables" to check in the equals function. Should I just use all of the getter methods for the Modulemd.ModuleStream, Modulemd.ModuleStreamV1, and Modulemd.ModuleStreamV2 classes and check that every word that is returned is equal for both objects? Any advice would be appreciated.

sgallagher:
[...]

PR #227, opened 2019-03-11
Fix lookups of streams to include architecture

### LINKED ISSUE #212: libmodulemd 2.x doesn't handle repodata where there are multiple stream documents for separate arches of the same NSVC
This may require a major-version soname bump. Try to avoid this if at all possible.

sgallagher:
Fixes #212 

The first two patches in the series are minor bugfixes noticed while working on the other bits.

The third patch moves the getter and setter for "arch" to `Modulemd.ModuleStream` rather than `Modulemd.ModuleStreamV1` and Modulemd.ModuleStreamV2`, because that information will be needed at this layer to support the `Modulemd.ModuleStream.get_NSVCA()` addition.

The fourth patcch introduces the new method `Modulemd.ModuleStream.get_NSVCA()`. It deprecates the previous function `Modulemd.ModuleStream.get_nsvc()`, generating non-fatal warnings during compilation and Python execution. It also modifies the `get_nsvc()` function to be a wrapper around  `Modulemd.ModuleStream.get_NSVCA()`.

The fifth patch adds a new search method: `Modulemd.Module.get_stream_by_NSVCA()`, which deprecates ` Modulemd.Module.get_stream_by_nsvc()` in favor of the version supporting architecture.

The sixth and final patch adds support to `Modulemd.Module.add_stream()` to consider architecture in addition to name, stream, version and context when determining if a particular stream is already present in the `Modulemd.Module`. Originally, I had planned to make it an error when the same NSVCA is encountered, but while writing this, I realized that the `Modulemd.ModuleIndexMerger` relies on the current implementation of deduplicating them by removing the older one. I added a TODO to the code for throwing an error if we determine that they are not in fact exactly identical, but that is currently blocked by #186.


sgallagher:
While filing this, I realized I don't actually have the correct deprecations around `Modulemd.ModuleStream.get_nsvc()`, so I'll fix that quickly and then force-push.

sgallagher:
The missing deprecations are now present.

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
```diff
@@ -352,11 +380,28 @@ modulemd_module_stream_get_context (ModulemdModuleStream *self);
...
 modulemd_module_stream_get_nsvc_as_string (ModulemdModuleStream *self);
 
 
+/**
+ * modulemd_module_stream_get_NSVCA_as_string: (rename-to modulemd_module_stream_get_NSVCA)
+ * @self: (in): This #ModulemdModuleStream.
+ *
+ * Returns: (transfer full): The NSVCA of this module stream. The returned
+ * format is described
+ * [here](https://docs.fedoraproject.org/en-US/modularity/architecture/consuming/naming-policy/#_forms).
```
Conan-Kudo: Will this reference work when generating the documentation? I don't recall if `gtk-doc` actually handles Markdown.

modulemd/v2/modulemd-module-index.c
```diff
@@ -353,9 +353,9 @@ compare_stream_SVC (gconstpointer a, gconstpointer b)
 {
   ModulemdModuleStream *a_ = *(ModulemdModuleStream **)a;
   ModulemdModuleStream *b_ = *(ModulemdModuleStream **)b;
-  g_autofree gchar *a_nsvc = modulemd_module_stream_get_nsvc_as_string (a_);
-  g_autofree gchar *b_nsvc = modulemd_module_stream_get_nsvc_as_string (b_);
-  return g_strcmp0 (a_nsvc, b_nsvc);
+  g_autofree gchar *a_nsvca = modulemd_module_stream_get_NSVCA_as_string (a_);
+  g_autofree gchar *b_nsvca = modulemd_module_stream_get_NSVCA_as_string (b_);
+  return g_strcmp0 (a_nsvca, b_nsvca);
```
Conan-Kudo: Is it a good idea to do it this way? Shouldn't we have this deprecated for an explicitly SVCA variant?

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
sgallagher: https://developer.gnome.org/gtk-doc-manual/stable/documenting_syntax.html.en

> As of version 1.20, the preferred way is to use a subset of the basic text formatting syntax called Markdown. On older GTK-Doc versions any documentation that includes Markdown will be rendered as is.

So this should work fine on any platform we're likely to care about.

modulemd/v2/modulemd-module-index.c
sgallagher: This is a static, internal function. I can rename it to `compare_stream_SVCA()` if you prefer. It's called in exactly one place, as a comparison function for a sort.

Conan-Kudo: I'd prefer us to be explicit, even with the function definition for an internal function. :)

sgallagher: Fixed and pushed.

PR #233, opened 2019-03-12
Update test-modulemd-merger.c

AarushiSingh09:
Modify some basic statements

### DIFF
--- a/modulemd/v2/tests/test-modulemd-merger.c
+++ b/modulemd/v2/tests/test-modulemd-merger.c
@@ -58,6 +58,8 @@ merger_test_deduplicate (CommonMmdTestFixture *fixture,
   index = modulemd_module_index_new ();
 
   g_assert_nonnull (yaml_path);
+  g_assert_nonnull (index);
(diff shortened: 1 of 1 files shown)


modulemd/v2/tests/test-modulemd-merger.c
```diff
@@ -58,6 +58,8 @@ merger_test_deduplicate (CommonMmdTestFixture *fixture,
   index = modulemd_module_index_new ();
 
   g_assert_nonnull (yaml_path);
+  g_assert_nonnull (index);
+  g_assert_null (error);
```
sgallagher: Checking the `error` here is meaningless, as the `error` was initialized to NULL and no function has been called that might modify it.

AarushiSingh09:
Okay No problem

modulemd/v2/tests/test-modulemd-merger.c
AarushiSingh09: Okay. I shall contribute accordingly. Thank you for the guidance.

PR #235, opened 2019-03-13
Update test-modulemd-merger.c

AarushiSingh09:
Added test merger_test_merger_with_real_world_data.

### DIFF
--- a/modulemd/v2/tests/test-modulemd-merger.c
+++ b/modulemd/v2/tests/test-modulemd-merger.c
@@ -106,7 +106,7 @@ merger_test_deduplicate (CommonMmdTestFixture *fixture,
 
 static void
 merger_test_merger_with_real_world_data (CommonMmdTestFixture *fixture,
-                         gconstpointer user_data)
+                                         gconstpointer user_data)
 {
   g_autoptr (ModulemdModuleIndex) fedora_index = NULL;
   g_autoptr (ModulemdModuleIndex) updates_index = NULL;
--- a/modulemd/v2/tests/test-modulemd-merger.c
+++ b/modulemd/v2/tests/test-modulemd-merger.c
@@ -208,7 +208,7 @@ main (int argc, char *argv[])
               merger_test_deduplicate,
               NULL);
   
-  g_test_add ("/modulemd/v2/module/index/merger/real/world/data",
+  g_test_add ("/modulemd/v2/module/index/merger/real_world_data",
               CommonMmdTestFixture,
               NULL,
               NULL,
--- a/modulemd/v2/tests/test-modulemd-merger.c
+++ b/modulemd/v2/tests/test-modulemd-merger.c
@@ -115,21 +115,40 @@ merger_test_merger_with_real_world_data (CommonMmdTestFixture *fixture,
   g_autofree gchar *yaml_path = NULL;
   g_autoptr (GPtrArray) failures = NULL;
   g_autoptr (GError) error = NULL;
+  g_autofree gchar *baseline = NULL;
+  g_autofree gchar *newline = NULL;
   
   yaml_path =
-    g_strdup_printf ("%s/modulemd/v2/tests/test_data/long-valid.yaml",
+    g_strdup_printf ("%s/modulemd/v2/tests/test_data/f29-updates.yaml",
                      g_getenv ("MESON_SOURCE_ROOT"));
   
   
   fedora_index = modulemd_module_index_new ();
   
-  yaml_path =
-    g_strdup_printf ("%s/modulemd/v2/tests/test_data/long-valid.yaml",
-                     g_getenv ("MESON_SOURCE_ROOT"));
+  g_assert_nonnull (yaml_path);
+
+  g_assert_true (modulemd_module_index_update_from_file (
+    fedora_index, yaml_path, TRUE, &failures, &error));
+  g_assert_cmpint (failures->len, ==, 0);
+  g_clear_pointer (&failures, g_ptr_array_unref);
+  
+  /* Save the baseline output for later comparison */
+  baseline = modulemd_module_index_dump_to_string (index, &error);
+  g_assert_nonnull (baseline);
+  g_assert_null (error);
+
 
   updates_index = modulemd_module_index_new ();
   
   g_assert_nonnull (yaml_path);
+  
+  g_assert_true (modulemd_module_index_update_from_file (
+    updates_index, yaml_path, TRUE, &failures, &error));
+  g_assert_cmpint (failures->len, ==, 0);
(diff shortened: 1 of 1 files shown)


modulemd/v2/tests/test-modulemd-merger.c
```diff
@@ -100,6 +104,47 @@ merger_test_deduplicate (CommonMmdTestFixture *fixture,
...
+{
+  g_autoptr (ModulemdModuleIndex) fedora_index = NULL;
+  g_autoptr (ModulemdModuleIndex) updates_index = NULL;
+  g_autoptr (ModulemdModuleIndex) merged_index = NULL;
+  g_autoptr (ModulemdModuleIndexMerger) merger = NULL;
+  g_autofree gchar *yaml_path = NULL;
+  g_autoptr (GPtrArray) failures = NULL;
+  g_autoptr (GError) error = NULL;
+  
+  yaml_path =
```
sgallagher: You allocate this, then don't free it anywhere. It gets masked by the other allocation at line 126, resulting in leaked memory.

modulemd/v2/tests/test-modulemd-merger.c
```diff
@@ -100,6 +104,47 @@ merger_test_deduplicate (CommonMmdTestFixture *fixture,
...
+    g_strdup_printf ("%s/modulemd/v2/tests/test_data/long-valid.yaml",
+                     g_getenv ("MESON_SOURCE_ROOT"));
+  
+  
+  fedora_index = modulemd_module_index_new ();
+  
+  yaml_path =
+    g_strdup_printf ("%s/modulemd/v2/tests/test_data/long-valid.yaml",
```
sgallagher: This is the same data, making this test effectively identical to the deduplication test. It would be better to use the `f29.yaml` and `f29-updates.yaml` data files instead of `long-valid.yaml`.

modulemd/v2/tests/test-modulemd-merger.c
```diff
@@ -100,6 +104,47 @@ merger_test_deduplicate (CommonMmdTestFixture *fixture,
...
+  merger = modulemd_module_index_merger_new ();
+  g_assert_nonnull (merger);
+  
+  modulemd_module_index_merger_associate_index (merger, fedora_index, 0);
+  modulemd_module_index_merger_associate_index (merger, updates_index, 0);
+  
+  merged_index = modulemd_module_index_merger_resolve (merger, &error);
+  g_assert_nonnull (merged_index);
+  g_assert_null (error);
+  g_clear_object(&merged_index);
```
sgallagher: It's not enough here to verify that the `merged_index` is not NULL. You need to actually validate that the merge results match what you expect.

You can either dump the YAML to a string and comparing it to a known plaintext or pick a few objects that you know should be updated during the merge and do a deep-comparison of them to see if they match the expected results.

sgallagher:
Also, when making changes, please update the same branch. Do not create new pull-requests, as that means that the history of changes won't be tracked in the same place.

modulemd/v2/tests/test-modulemd-merger.c
AarushiSingh09: I have added cmpstr with a baseline and newline (sumped yaml string). Is it enough. It is in the same way the deduplicated test was written.

modulemd/v2/tests/test-modulemd-merger.c
AarushiSingh09: I have removed the line 126 and changed  the first yaml_path to f29-updates.yaml file to make it non identical with the deduplicated test as suggested by you.

modulemd/v2/tests/test-modulemd-merger.c
sgallagher: That only worked for the deduplication test because the two indices were in fact exactly the same. If you are doing a proper merge here with `f29` and `f29-updates`, the result is expected to be entirely different from either of the inputs.

AarushiSingh09: Can you please elaborate. I am getting confused on what actually I need to do. I have added f29 as well.

sgallagher: It may not be a good idea to start with such large data sets for this test. It may be a better idea to start small: maybe create two different ModuleStream YAML files and verify that a merge stores both of them to the output.

We can work up to replacements and complicated interactions.

AarushiSingh09: When I searched about ModuleStream, this is what I found :" STREAMS enables you to create modules to provide standard data communications services and then manipulate the modules on a stream. From the application level, modules can be dynamically selected and interconnected. No kernel programming, compiling, and link editing are required to create the interconnection". But I don't know exactly how to write statements to create a ModuleStream YAML files. Can you please guide me?

AarushiSingh09: Also can you please tell me if I am going in the right direction or not?

sgallagher: You are not. I'm not sure where you got that message from, but it's entirely unrelated to Fedora Modularity. (It looks like it may have to do with Linux kernel modules).

ModuleStream YAML is described by https://github.com/fedora-modularity/libmodulemd/blob/master/spec.v2.yaml and many examples are provided in the [`modulemd/v2/tests/test_data](https://github.com/fedora-modularity/libmodulemd/tree/master/modulemd/v2/tests/test_data) directory in the codebase.

The `f29.yaml` contains MANY such documents inside a single YAML file. An example of a single ModuleStream document would be [`dependson_v2.yaml`](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/tests/test_data/dependson_v2.yaml)

What I'm asking is for you to create another example module stream document similar to [`dependson_v2.yaml`](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/tests/test_data/dependson_v2.yaml), load each of them into an index, merge the two indices and then verify that the output is as expected.

sgallagher:
Please make sure to run `ninja test` before committing and pushing a commit. The test suite also runs a code-formatter that will help ensure that the style guidelines are followed. The CI detects whether the formatter makes changes and fails if the code in the patch is not properly formatted.

AarushiSingh09:
I am very thankful for your patience.I am afraid that the fedora on my
virtual machine is not working properly, I am trying to figure out the
solution otherwise I'll do a dual boot on my system hence I won't be run
the ninja test right away. Untill then Can you please point out the mistake
I don't want to delay my contributions sir.

AarushiSingh09:
There is still some problem with my fedora system. I am trying to contribute in some other issue until that error fixes and I ma able to run the ninja test.

PR #236, opened 2019-03-13
Adding Mageia 7 test

Lashuk1729:
As, Mageia 7 allows the user to use `dnf` as a system package manager in Mageia, so it did not require many changes compared to Fedora configuration.


.travis/mageia/travis-tasks.sh
```diff
@@ -0,0 +1,124 @@
...
+meson --buildtype=debug \
+      -Dbuild_api_v1=true \
+      -Dbuild_api_v2=true \
+      $COMMON_MESON_ARGS \
+      travis
+
+set +e
+ninja -C travis test
+if [ $? != 0 ]; then
+    cat travis_v1/meson-logs/testlog.txt
```
sgallagher: This should be `travis`, not `travis_v1`. There was a bug in the centos version originally, but that should have been fixed a few days ago. Make sure you are rebased atop the latest master branch.

et --hard HEAD
```diff
@@ -0,0 +1,6679 @@
+[33mcommit 7cd57e6ecdd1fb494c4b5b51112dae803d82ee8b[m[33m ([m[1;36mHEAD -> [m[1;32mmaster[m[33m)[m
```
sgallagher: This entire file looks like it was an accidental commit. Remove it.

sgallagher: It also appears to be the cause of the CI build failure, as `git ls-files` is choking on this file.

Issue #96, opened 2018-10-22
implement "buildafter" or something similar for components/rpms


ignatenkobrain:
Only one of `buildorder` or `buildafter` should be used. Then MBS could take advantage of this submitting all builds at once and let koji handle build queue and priorities.

I picked first name which came to my mind, you can choose different one.

contyk:
Do you think just the notion of "after" would be sufficient or should there be other keywords, similar to the dependency language of systemd unit files, for instance?

By the way, this is tracked as a high level goal here: https://tree.taiga.io/project/modularity-wg/epic/33

ignatenkobrain:
hmmm, I think `after` should be enough.

sgallagher:
@contyk, @ignatenkobrain 
I gave this some thought this morning and I have a suggestion for how this might work. Here's an example of the structure (underneath the `data` section of modulemd and excluding the rest of the modulemd)

```yaml
components:
    rpms:
        myrpm:
            rationale: My RPM
            ref: master
            after: somedep
            bootstrap:
                first_pass:
                    buildopts:
                        rpms:
                            macros: <stuff>
                second_pass:
                    buildopts:
                        rpms:
                            macros: <more stuff>
                    after:
                        myrpm: first_pass
                        somedep: ""
                        someotherdep: ""                    
        somedep:
            rationale: Dep for myrpm
            ref: master
            after:
                myrpm: first_pass
        someotherdep:
            rationale: Needed to bootstrap rationale
            ref: master
```

To break it down, we'd add two new directives: `after` and `bootstrap`. We would replace `buildorder` with `after` (and I'd recommend we make this a metadata version bump so we can process upgrades from v2 rather than trying to have `buildorder` and `after` coexist. `after` is a YAML mapping whose keys are the names of other RPM components in this module and whose value is the name of the bootstrapping section (see below) they need to occur after. If they need to be ordered after the main RPM, the empty string should be used.

The second addition is the new `bootstrap` section. We can bikeshed on the name if we must, but the main reason one would want to build more than once is to handle bootstrapping situations, so it seemed like the right name to me. The idea is that if the bootstrap section is present, its builds must be performed before the main build is run. If there are multiple bootstrap steps, they can be named whatever makes sense and should be ordered using the same `after` directive.

[...]

PR #237, opened 2019-03-13
Add override_name for components

sgallagher:
Related to https://github.com/fedora-modularity/libmodulemd/issues/96


sgallagher:
Pushed a second patch that also updates documentation and function prototypes for `new()` and `copy()` to make it clear that those take the `key` and not the real name.

PR #239, opened 2019-03-14
[WIP] Update test-modulemd-merger.c

AarushiSingh09:
Addition of merger_test_merger_with_modified EMPTY function for Pass statement


sgallagher:
I've updated the title to include [WIP]. When this is close to being ready to merge, please drop that prefix.

AarushiSingh09:
I am working on the other function but this is a new function which I added. What is the issue with this function sir?

sgallagher:
> I am working on the other function but this is a new function which I added. What is the issue with this function sir?

An empty function is not a useful change. If the original Python has a TODO for that test, copying it doesn't make sense.

PR #241, opened 2019-03-15
Fixes #218: Provide Modulemd.ServiceLevel.equals() method

### LINKED ISSUE #218: Provide Modulemd.ServiceLevel.equals() method
In order to implement #186, we need to have an equality function for `Modulemd.ServiceLevel` objects.

OrionStar25:
Fixes #218

### DIFF
--- a/modulemd/v2/modulemd-service-level.c
+++ b/modulemd/v2/modulemd-service-level.c
@@ -55,6 +55,37 @@ modulemd_service_level_new (const gchar *name)
 }
 
 
+gboolean
+modulemd_service_level_equals (ModulemdServiceLevel *self_1,
+                               ModulemdServiceLevel *self_2)
+{
+  if (!self_1 && !self_2)
+    return TRUE;
+
+  if (!self_1 || !self_2)
+    return FALSE;
+
+  g_return_val_if_fail (MODULEMD_IS_SERVICE_LEVEL (self_1), FALSE);
+  g_return_val_if_fail (MODULEMD_IS_SERVICE_LEVEL (self_2), FALSE);
+
+  if (g_strcmp0 (modulemd_service_level_get_name (self_1),
+                 modulemd_service_level_get_name (self_2)) != 0)
+    return FALSE;
+
+  /*if both eols are invalid, its equivalent*/
+  if (!g_date_valid (self_1->eol) && !g_date_valid (self_2->eol))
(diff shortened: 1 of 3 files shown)


modulemd/v2/modulemd-service-level.c
```diff
@@ -55,6 +55,30 @@ modulemd_service_level_new (const gchar *name)
...
+  g_return_val_if_fail (MODULEMD_IS_SERVICE_LEVEL (self_2), FALSE);
+
+  if (g_strcmp0 (modulemd_service_level_get_name (self_1),
+                 modulemd_service_level_get_name (self_2)) != 0)
+    return FALSE;
+
+  if ((!self_1->eol && self_2->eol) || (!self_2->eol && self_1->eol))
+    return FALSE;
+
+  if (!g_date_valid (self_1->eol) || !g_date_valid (self_2->eol))
```
sgallagher: `g_date_valid()` will crash if passed a `NULL` value, which is possible if both `self_1->eol` and `self_2->eol` are `NULL` above. I think it's best if we just document that this function will always return false if either service level passed in contains a NULL or invalid date, since they cannot be legitimately compared. If they are *both* NULL, we can treat that as a match and return that immediately (and skip the `g_date_valid()` and `g_date_compare()` calls)

modulemd/v2/modulemd-service-level.c
```diff
@@ -55,6 +55,30 @@ modulemd_service_level_new (const gchar *name)
 }
 
 
+gboolean
+modulemd_service_level_equals (ModulemdServiceLevel *self_1,
```
sgallagher: You need to include a prototype for this function in the `modulemd-service-level.h` header file. Follow the examples of the other functions in there for what the GTKDoc comments should be (and mark it as `Since: 2.2`).

OrionStar25:
@sgallagher added tests, but `ninja test` failing.

sgallagher:
@OrionStar25 I’ll look at the patch when I get in, but if `ninja test` is failing, the log output should show you why.

sgallagher:
Please squash this into a single patch and I'll merge it.

OrionStar25:
@sgallagher  Done.

sgallagher:
Approved. I'll merge it as soon as the CI run completes. Excellent work!

PR #242, opened 2019-03-16
Adding Mageia 7 Tests

Lashuk1729:
I had made the changes suggested in the previous Pull Request. 
Mageia 7 allows the user to use `dnf` as a system package manager in Mageia, so it did not require many changes compared to Fedora configuration.


Conan-Kudo:
There's now a `mageia:7` image on Docker Hub.

PR #248, opened 2019-03-19
Fixes #243: Eliminate typecasting warning for modulemd_hash_table_equals()

### LINKED ISSUE #243: Eliminate typecasting warning for modulemd_hash_table_equals()
https://github.com/fedora-modularity/libmodulemd/pull/213 introduced a warning when passing `modulemd_hash_table_sets_are_equal()` as an argument to `modulemd_hash_table_equals()` because the latter expects a `(void *, void *)` argument and `modulemd_hash_table_sets_are_equal()` accepts `(GHashTable *, GHashTable *).

The solution here is to add a new function to wrap `modulemd_hash_table_sets_are_equal()` and cast the `(void *)` pointers explicitly to `(GHashTable *)` pointers.

OrionStar25:
Fixes #243 

Now I'm getting this warning  :- 

```
/modulemd/v2/modulemd-util.c: In function ‘modulemd_hash_table_sets_are_equal_wrapper’:
../modulemd/v2/modulemd-util.c:140:10: warning: returning ‘gboolean’ {aka ‘int’} from a function with return type ‘GEqualFunc’ {aka ‘int (*)(const void *, const void *)’} makes pointer from integer without a cast [-Wint-conversion]
   return modulemd_hash_table_sets_are_equal ((GHashTable *)a, (GHashTable *)b);
```
But `GEqualFunc` has a `gboolean` return type as well. Also, how is it working for `g_compare_string` and not for this?

Btw, I did cast `modulemd_hash_table_sets_are_equal ()` to `GEqualFunc` but that gave a whole new set of warnings.

### DIFF
--- a/modulemd/v2/include/modulemd-2.0/private/modulemd-util.h
+++ b/modulemd/v2/include/modulemd-2.0/private/modulemd-util.h
@@ -77,6 +77,9 @@ GHashTable *
 modulemd_hash_table_deep_str_str_set_copy (GHashTable *orig);
 
 gboolean
+modulemd_hash_table_sets_are_equal_wrapper (const void *a, const void *b);
+
+gboolean
 modulemd_hash_table_sets_are_equal (GHashTable *a, GHashTable *b);
 
 gboolean
--- a/modulemd/v2/modulemd-util.c
+++ b/modulemd/v2/modulemd-util.c
@@ -134,6 +134,11 @@ modulemd_hash_table_deep_str_str_set_copy (GHashTable *orig)
   return new;
 }
 
+gboolean
+modulemd_hash_table_sets_are_equal_wrapper (const void *a, const void *b)
+{
+  return modulemd_hash_table_sets_are_equal ((GHashTable *)a, (GHashTable *)b);
+}
 
 gboolean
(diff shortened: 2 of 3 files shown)


modulemd/v2/include/modulemd-2.0/private/modulemd-util.h
```diff
@@ -76,6 +76,9 @@ modulemd_hash_table_deep_str_set_copy (GHashTable *orig);
 GHashTable *
 modulemd_hash_table_deep_str_str_set_copy (GHashTable *orig);
 
+GEqualFunc
```
sgallagher:
```suggestion
gboolean
```

A GEqualFunc is actually a pointer type, which is why you're getting type errors. It describes a pointer to a function that must return `gboolean`.

modulemd/v2/modulemd-util.c
```diff
@@ -134,6 +134,11 @@ modulemd_hash_table_deep_str_str_set_copy (GHashTable *orig)
   return new;
 }
 
+GEqualFunc
```
sgallagher:
```suggestion
gboolean
```

OrionStar25:
@sgallagher Done.

Issue #95, opened 2018-10-22
implement "buildonly" for components/rpms


ignatenkobrain:
For instance, in rust ecosystem we just want to pull components for the buildroot. Yes, you can use `filter`, but since that doesn't support globs/regex, it would be to complex to list 2-30 rpms per component.

contyk:
This also makes sense to me. MBS could also translate this into `filter` for easier debugging later.

This will need an epic.

ignatenkobrain:
any chance to get this prioritized?

https://src.fedoraproject.org/modules/stratis/blob/1/f/stratis.yaml#_31

my filter list is taking 368 lines... and it's just one module

sgallagher:
@ignatenkobrain I agree that this makes sense, but it's also non-trivial to implement. I can add the field to libmodulemd easily-enough, but MBS will also need to be updated to convert the artifacts produced by the component into the filter list.

ignatenkobrain:
I think we've pretty much agreed today on F2F meeting that libmodulemd will just add field and MBS will have to find all artifacts built by component and translate it to the filter list.

sgallagher:
Yes, during the recent Modularity Hackfest, this was discussed. I think what we will want to do is add a new optional attribute to the `data.components.rpms` section like:
```
components:
  rpms:
    rust-bar:
      rationale: Needed for frobbing
      ref: 26ca0c0
      buildonly: True
```

The `buildonly` attribute would default to `False` if unspecified. The expected behavior on the build-system side would be to identify any subpackages produced by `buildonly: True` components and automatically add them to the `data.filter` section for the output.

ignatenkobrain:
`true` should be lower case as per yaml's boolean type.

sgallagher:
Sorry, was writing some python and it bled through. I'll edit the above message to correct that.

contyk:
+1 to the proposal.

sgallagher:
OK, this looks like it's accepted, so I'll add this to libmodulemd 2.2.0

PR #250, opened 2019-03-20
Add 'buildonly' option to Components

sgallagher:
Fixes: https://github.com/fedora-modularity/libmodulemd/issues/95

Also adds a utility function to parse boolean values from the modulemd YAML.

When emitting, we will omit the `buildonly` field only if it is `TRUE` for backwards-compatibility.

### DIFF
--- a/modulemd/v2/modulemd-yaml-util.c
+++ b/modulemd/v2/modulemd-yaml-util.c
@@ -392,6 +392,35 @@ modulemd_yaml_parse_string (yaml_parser_t *parser, GError **error)
 }
 
 
+gboolean
+modulemd_yaml_parse_bool (yaml_parser_t *parser, GError **error)
+{
+  MMD_INIT_YAML_EVENT (event);
+
+  YAML_PARSER_PARSE_WITH_EXIT_BOOL (parser, &event, error);
+  if (event.type != YAML_SCALAR_EVENT)
+    {
+      MMD_YAML_ERROR_EVENT_EXIT_INT (
+        error, event, "Expected a scalar boolean");
+    }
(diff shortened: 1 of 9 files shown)


modulemd/v2/modulemd-yaml-util.c
```diff
@@ -392,6 +392,35 @@ modulemd_yaml_parse_string (yaml_parser_t *parser, GError **error)
...
+  MMD_INIT_YAML_EVENT (event);
+
+  YAML_PARSER_PARSE_WITH_EXIT_BOOL (parser, &event, error);
+  if (event.type != YAML_SCALAR_EVENT)
+    {
+      MMD_YAML_ERROR_EVENT_EXIT_INT (
+        error, event, "Expected a scalar boolean");
+    }
+
+  if (g_strcmp0 ((const gchar *)event.data.scalar.value, "false"))
```
ignatenkobrain: Oh, isn't there some better way to handle booleans?

sgallagher: Could you be more specific? We really only get them as a scalar value. I dabbled with the idea of doing a case-insensitive compare, but that seemed like overkill.

We can always enhance this function later, that's why it's separate.

ignatenkobrain: FALSE is not a boolean by yaml specification

I was thinking whether there is libyaml function to get bool

sgallagher: Nothing public: https://github.com/yaml/libyaml/blob/master/include/yaml.h

PR #251, opened 2019-03-22
Copied test_nsvca to C

athira-selvam:
Copied test_nsvca from modulestream.py to module_stream_test_nsvca in
test-modulemd-modulestream.c

### DIFF
--- a/modulemd/v2/tests/test-modulemd-modulestream.c
+++ b/modulemd/v2/tests/test-modulemd-modulestream.c
@@ -184,6 +184,86 @@ module_stream_test_copy (ModuleStreamFixture *fixture, gconstpointer user_data)
     }
 }
 
+
+static void
+module_stream_test_nsvca (ModuleStreamFixture *fixture,
+                          gconstpointer user_data)
+{
+  g_autoptr (ModulemdModuleStream) stream = NULL;
+  g_autofree gchar *s_nsvca = NULL;
+  guint64 version;
+
+  for (version = MD_MODULESTREAM_VERSION_ONE;
+       version <= MD_MODULESTREAM_VERSION_LATEST;
+       version++)
+    {
+      // First test that NSVCA is None for a module with no name
+      stream = modulemd_module_stream_new (version, NULL, NULL);
+      s_nsvca = modulemd_module_stream_get_NSVCA_as_string (stream);
+      g_assert_null (s_nsvca);
+      g_clear_pointer (&s_nsvca, g_free);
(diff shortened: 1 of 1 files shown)


athira-selvam:
@sgallagher  Hi,
This is giving me an error while running ninja tests ----


The output from the failed tests:

57/57 valgrind                                FAIL    28.50 s (exit status 1)



--- stderr ---
Memory leak detected in modulestream_v2_debug
Memory leak detected in modulestream_v2_debug
Memory leak detected in modulestream_v2_debug
Memory leak detected in modulestream_v2_debug
Memory leak detected in modulestream_v2_debug
Memory leak detected in modulestream_v2_debug
Memory leak detected in modulestream_v2_debug
Memory leak detected in modulestream_v2_debug
Memory leak detected in modulestream_v2_debug
-------

athira-selvam:
Could the memory leak be occurring in my test code or in the code the test is executing?

OrionStar25:
@athira-selvam I may be wrong but a similar thing happened with me. Its most probably a problem in your test code. You've not initialized `version` in the beginning. I think initializing it to `0` should make it fine.

athira-selvam:
@OrionStar25Thank you.  I tried that. But I am getting the same error. I think that is not the problem.

sgallagher:
Sorry, I missed this conversation somehow. What you want to do is run valgrind on the specific test that is failing. You can do this by running:
`meson test --wrap="valgrind --leak-check=full" modulestream_v2_debug --test-args="-p /modulemd/v2/modulestream/nsvca"`

Which translates to "run the `modulestream_v2_debug` tests that match the `/modulemd/v2/modulestream/nsvca` path name under valgrind". The output will include a link to a log that has all the valgrind output. You want to search that file for the word "definitely", which will show you any memory leaks. It's possible that the leaks are in your tests or that the NSVCA code itself is leaking, in which case I'd appreciate a fix from you. That's the point of writing tests, after all.

athira-selvam:
I found the error. In the gchar * modulemd_module_stream_get_NSVCA_as_string function,
nsvca is declared as gchar *nsvca and it is returned by the function. So when I called the function inside g_assert_cmpstr() it was not getting freed. Now I instead assigned it to another variable, used it and freed it. So now that is working.

Do you recommend any other way to fix this inside the code itself, so that the we don't have to worry about freeing it in function call.

sgallagher:
This PR is no longer required, now that #255 is merged.

Issue #55, opened 2018-04-19
modulemd: include checksum information into artifacts.rpm


ignatenkobrain:
Lookup by NEVRA is extremely slow... See filelists.xml for example how checksum is used.

I was pretty sure I have filled this RFE before, but can't find it.

dmach:
I don't recommend using checksums. Query performance improvements will be part of new libdnf release (part of DNF 3).
If you implement the checksums, you'll have to deal with sigkeys first.
RPMs get built unsigned. Then they're signed with a Beta key and then Release key.
I don't think you want to rebuild modulemd every time an RPM gets signed.

ignatenkobrain:
@dmach note that filelists already have to include this information so metadata has to be regenerated anyway.

Also libdnf/dnf is irrelevant here, because I'm not interested in using it.

dmach:
Igor,
you don't seem to understand the problem.
Filelists contain checksum information, but it's an implementation detail in repomd repodata - how to connect data across 3 files where checksum represents a primary key.

I'm talking about situation when you replace an unsigned RPM in a repo with a signed copy.
Repodata refresh is expected, because repodata is not build and tracked as a proper artifact, it's an index of RPMs.

On the other hand, modulemd represents an artifact - a module. And you don't want it to depend on exact RPM checksums - or do you? In that case, we'd have to invalidate old modulemd and build a new one with new RPM checksums any time a RPM in repo gets signed.

ignatenkobrain:
> And you don't want it to depend on exact RPM checksums - or do you

You do. Imagine that you changed license and rebuilt package with same NEVRA. This means you need to regenerate `licenses/content` yaml tag.

sgallagher:
You can’t rebuild an RPM with the same NEVRA; the build system won’t allow it.

ignatenkobrain:
@sgallagher depends on the buildsystem.

sgallagher:
We're going to reopen this issue as a result of conversations at today's hackfest in Boston. We will add a new section under artifacts linking the NEVRAs to one or more hashes. The format will be discussed a bit later.

ignatenkobrain:
```yaml
artifacts:
  rpm-map:
    - pkgid: $checksum
       type: sha256 (or any other checksum type)
       name: $rpm-name
       evr: $rpm-evr (without leading 0: )
       arch: $rpm-arch
```

[...]

PR #252, opened 2019-03-23
Add support for RPM repomd checksums

sgallagher:
Fixes https://github.com/fedora-modularity/libmodulemd/issues/55

Creates a new object type, `Modulemd.RpmMapEntry` to store NEVRA data and then associates it to the `Modulemd.ModuleStreamV2` object by digest-type and checksum.


spec.v2.yaml
```diff
@@ -357,3 +357,24 @@ data:
             - xxx-0:1-1.module_deadbeef.x86_64
             - xxx-0:1-1.module_deadbeef.i686
             - xyz-0:1-1.module_deadbeef.x86_64
+        # The rpm-map exists to link checksums from repomd to specific
+        # artifacts produced by this module. Any item in this list must match
+        # a
```
ignatenkobrain: seems like something is missing here.

modulemd/v2/include/modulemd-2.0/modulemd-rpm-map-entry.h
```diff
@@ -0,0 +1,230 @@
...
+ *
+ * Create a new rpm-map entry. This will contain the decomposed NEVRA of an
+ * RPM artifact.
+ *
+ * Returns: (transfer full): A newly-allocated #ModulemdRpmMapEntry with the
+ * provided values.
+ */
+ModulemdRpmMapEntry *
+modulemd_rpm_map_entry_new (const gchar *name,
+                            guint32 epoch,
```
Conan-Kudo: Why isn't this a const as well?

spec.v2.yaml
sgallagher: Hmm, I wonder how that happened...

modulemd/v2/include/modulemd-2.0/modulemd-rpm-map-entry.h
sgallagher: The epoch value? The `const` would be meaningless here as the value is copied, not passed by reference.

sgallagher:
Pushed a fix for the missing half a sentence in `spec.v2.yaml`

spec.v2.yaml
```diff
@@ -357,3 +357,24 @@ data:
...
+        # Optional.
+        rpm-map:
+          # The digest-type of this checksum. Required.
+          sha256:
+            # The checksum of the artifact being sought. Required.
+            ee47083ed80146eb2c84e9a94d0836393912185dcda62b9d93ee0c2ea5dc795b:
+              # The RPM name. Required.
+              name: bar
+              # The RPM epoch. Optional
+              epoch: 0
```
Conan-Kudo: I don't know if we should make this property optional. It's not optional in rpm-md, as null epoch is automatically forced to zero epoch.

sgallagher: I made it optional for reading, but libmodulemd outputs it unconditionally. Is that good enough or do you feel strongly that it should be mandatory in all cases?

Conan-Kudo: I think it should remain mandatory in all cases. Remember, this should be consistent for other YAML-based parsers for modulemd.

sgallagher:
I rebased atop the current master branch and made the epoch field mandatory (and validated that it's present during parsing).

sgallagher:
I added the most recent changes as a fixup commit that I'll squash before merging, but I left it separate for ease of review.

PR #254, opened 2019-03-24
Fixes #216: Provide Modulemd.Dependencies.equals() method

### LINKED ISSUE #216: Provide Modulemd.Dependencies.equals() method
In order to implement #186, we need to be able to compare two `Modulemd.Dependencies` objects for equality.

OrionStar25:
Fixes #216 

This is my error log:
```
/home/orionstar25/libmodulemd/api2/modulemd/v2/dependencies_v2:8907): 
libmodulemd-WARNING **: 15:31:18.568: Streams requested for unknown module: buildmod1** 
libmodulemd:ERROR:../modulemd/v2/tests/test-modulemd-dependencies.c:194:dependencies_test_equals: 
'modulemd_dependencies_equals (d_1, d_2)' should be FALSE
```

I tried the individual assertions and turns out that the `modulemd_dependencies_add_buildtime_stream (d_1, "buildmod1", "stream2");` is not even adding any streams but directly going to `modulemd_dependencies_set_empty_buildtime_dependencies_for_module (d_1, "builddef");`

### DIFF
--- a/modulemd/v2/include/modulemd-2.0/modulemd-dependencies.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-dependencies.h
@@ -41,6 +41,22 @@ modulemd_dependencies_new (void);
 
 
 /**
+ * modulemd_dependencies_equals:
+ * @self_1: A #ModulemdDependencies
+ * @self_2: A #ModulemdDependencies
+ *
+ * Check if self_1 and self_2 are equal objects
+ *
+ * Returns: TRUE, if the're equal. FALSE, otherwise.
+ *
+ * Since: 2.2
+ */
+gboolean
+modulemd_dependencies_equals (ModulemdDependencies *self_1,
+                              ModulemdDependencies *self_2);
+
+
+/**
  * modulemd_dependencies_copy:
  * @self: This #ModulemdDependencies
  *
--- a/modulemd/v2/modulemd-dependencies.c
+++ b/modulemd/v2/modulemd-dependencies.c
@@ -24,7 +24,14 @@ struct _ModulemdDependencies
 {
   GObject parent_instance;
 
+  /* @key: dependent modules.
+   * @value: GHashTable set of compatible streams
+   */
   GHashTable *buildtime_deps;
+
+  /* @key: dependent modules.
+   * @value: GHashTable set of compatible streams
+   */
   GHashTable *runtime_deps;
 };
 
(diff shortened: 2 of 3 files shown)


sgallagher:
> I tried the individual assertions and turns out that the `modulemd_dependencies_add_buildtime_stream (d_1, "buildmod1", "stream2");` is not even adding any streams but directly going to `modulemd_dependencies_set_empty_buildtime_dependencies_for_module (d_1, "builddef");`

I'm not sure I understand what you mean here. Is there a problem with the implementation of `modulemd_dependencies_add_buildtime_stream ()`? If so, please use your tests to help locate it. You can use `meson test --gdb dependencies_v2_debug"` to launch a GDB debugger session and then do `run -p /modulemd/v2/dependencies/equals` to debug just that single test.

OrionStar25:
> > I tried the individual assertions and turns out that the `modulemd_dependencies_add_buildtime_stream (d_1, "buildmod1", "stream2");` is not even adding any streams but directly going to `modulemd_dependencies_set_empty_buildtime_dependencies_for_module (d_1, "builddef");`
> 
> I'm not sure I understand what you mean here. Is there a problem with the implementation of `modulemd_dependencies_add_buildtime_stream ()`? If so, please use your tests to help locate it. You can use `meson test --gdb dependencies_v2_debug"` to launch a GDB debugger session and then do `run -p /modulemd/v2/dependencies/equals` to debug just that single test.

I'm getting the same error after that. This is what is happening:
`modulemd_dependencies_add_buildtime_stream (d_1, "buildmod1", "stream2");`  should be adding a buildtime stream for 'buildmod1' but its not. Instead, its executing `modulemd_dependencies_set_empty_buildtime_dependencies_for_module (d_1, "builddef");` directly.

So now, my first object that I created has just the module '`builddef`. Same with my 2nd object. So, technically, those 2 objects are equal. But my tests are written such that they have different `buildtime_streams`. So, its showing an error at that particular assertion.

OrionStar25:
Alright! Found what's wrong. 

Suppose object 1 has this:
```
Modulename: buildmod1
Streams: ['a', 'b']
```
And object 2 has this:
```
Modulename: buildmod1
Streams: ['a', 'b', 'c']
```
Its still considering both the objects as equal. Whereas we should call them unequal (?)

OrionStar25:
And solved. I took the 2 hashtables as sets when they were normal hashtables. Is there any documentation for them because there aren't any comments specifying what type they are.

OrionStar25:
@sgallagher Done.

PR #256, opened 2019-03-27
Install overrides for Python 2

sgallagher:
The Python 2 GObject Introspection bindings should get the same overrides that the Python 3 implementation does.

Also includes a minor typo I spotted in the openSUSE tests. The effect was only cosmetic because the variable was not being used yet.


sgallagher:
Another patch pushed to run all Python tests with both interpreter versions.

PR #260, opened 2019-04-01
Fixes #215: Provide Modulemd.Component.equals() method

### LINKED ISSUE #215: Provide Modulemd.Component.equals() method
Implement an equality check for `Modulemd.Component` objects (and their child-classes, `Modulemd.ComponentModule` and `Modulemd.ComponentRpm`). This is necessary to implement #186.

OrionStar25:
Fixes #215

### DIFF
--- a/modulemd/v2/include/modulemd-2.0/modulemd-component.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-component.h
@@ -36,11 +36,26 @@ struct _ModulemdComponentClass
   void (*set_name) (ModulemdComponent *self, const gchar *name);
   const gchar *(*get_name) (ModulemdComponent *self);
   gboolean (*validate) (ModulemdComponent *self, GError **error);
+  gboolean (*equals) (ModulemdComponent *self_1, ModulemdComponent *self_2);
 
   /* Padding to allow adding up to 8 new virtual functions without
    * breaking ABI. */
   gpointer padding[8];
 };
+
+/**
+ * modulemd_component_equals:
+ * @self_1: A #ModulemdComponent object
+ * @self_2: A #ModulemdComponent object
+ *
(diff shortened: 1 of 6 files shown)


sgallagher:
Also, please include new tests when submitting new functions, as it's the easiest way for both of us to determine if things worked correctly (and if the tests are in C, we can see if the new methods have memory leaks).

OrionStar25:
@sgallagher please review.

sgallagher:
@OrionStar25 There appears to be a linking problem in here somewhere, at least on Fedora 30 and Rawhide. I'm going to look into whether this is due to Fedora changing default linker settings and us having been implicitly pulling things in or if there's a bug in Fedora's linker.

sgallagher:
I found the problem, @OrionStar25 
I hadn't bumped the release version after the latest 2.2.2 release and on Fedora 30+, that version went stable. As a result, the CI tests were installing the version from the distro repository instead of the ones from the pull-request and they were failing to link.

Please rebase atop the current master branch and squash your changes and it should work properly. Once that is done, I will review again (and probably merge it).

OrionStar25:
@sgallagher Please review! :+1:

modulemd/v2/include/modulemd-2.0/modulemd-component.h
```diff
@@ -36,12 +36,27 @@ struct _ModulemdComponentClass
...
 };
 
+/**
+ * modulemd_component_equals:
+ * @self_1: A #ModulemdComponent object
+ * @self_2: A #ModulemdComponent object
+ *
+ * Returns: TRUE, if both the objects are equal. FALSE, otherwise
+ *
+ * Since: 2.2
```
sgallagher: I missed this before, sorry. 2.2 has already been released, so new functions need to be `Since: 2.3` now. I will fix this while merging. No need to resubmit.

Issue #208, opened 2019-03-04
RFE: Support directly reading in compressed yaml streams (gzip, bz2, xz, zck,...)


Conan-Kudo:
In most cases, the module metadata is stored in a compressed form (with some kind of compression algorithm like gzip, bz2, xz, zck, etc.). As far as libmodulemd is concerned, it can't deal with these files directly.

It would be nice to be able to handle these files directly. This is likely better as a new function as to not break API expectations or ABI.

NishkaDG:
May I work on this?

sgallagher:
If you want to prepare a proposal for how to implement this, go ahead. Please don’t start writing any code; it’s a big task and needs significant design work first.

sgallagher:
OK, looks like I'm going to need to implement at least part of this to work with `createrepo_c`/`mergerepo_c`.

What I plan to do is add a new function, `modulemd_module_index_update_from_custom()` that will take a function pointer to a custom `read()` function of the type [`yaml_read_handler_t`](https://github.com/yaml/libyaml/blob/master/include/yaml.h#L970) that will be passed directly to [`yaml_parser_set_input()`](https://github.com/yaml/libyaml/blob/master/include/yaml.h#L1369) and will then proceed as normal.

In the specific case of createrepo_c, this will mean wrapping the `cr_read()` function that handles the various compression formats into a `yaml_read_handler_t` function.

@Conan-Kudo Do you think this will be sufficient from the libmodulemd side? I don't really think libmodulemd needs to implement the compression-handling logic as long as the consumer can provide it with a valid stream of data.

Conan-Kudo:
Honestly, the reason I think it should have it is because it is often the preferred way to store the metadata because it is huge otherwise.

If you don't think this is an issue, then I guess it isn't needed...

sgallagher:
@Conan-Kudo I think I'd prefer to avoid adding all of the codec dependencies to libmodulemd itself, to keep things neat. That said, maybe we could look into adding a helper library (libmodulemd-codecs?) that a consumer could optionally link against.

The thing is, I suspect that the majority of consumers will need to have their own logic for accessing compressed content and it would be best in general if we just allowed them to pass this to us as a custom read handler.

[...]

PR #261, opened 2019-04-01
Add Modulemd.ModuleIndex.update_from_custom()

sgallagher:
In order to support YAML streams from arbitrary sources (such as those
being read from a compressed file through an intermediate library), we
need to support reading from a custom function. This patch enables
consumers of libmodulemd to have indirect access to libyaml's
yaml_parser_set_input() to accomplish this.

Adds a basic test to read through a long string and verify that it
produced an index with the read module.

Related: https://github.com/fedora-modularity/libmodulemd/issues/208

### DIFF
--- a/modulemd/v2/include/modulemd-2.0/modulemd-module-index.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-module-index.h
@@ -72,6 +72,35 @@ G_DECLARE_FINAL_TYPE (
 
 
 /**
+ * ModulemdReadHandler:
+ * @data: (inout): A private pointer to the data being read.
+ * @buffer: (out): The buffer to write the data from the source.
+ * @size: (in): The size of the buffer.
+ * @size_read: (out): The actual number of bytes read from the source.
+ *
+ * The prototype of a read handler.
+ *
+ * The read handler is called when the parser needs to read more bytes from the
+ * source. The handler should write not more than @size bytes to the @buffer.
+ * The number of written bytes should be set to the @size_read variable.
+ *
+ * This handler is identical to a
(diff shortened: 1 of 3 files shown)


modulemd/v2/include/modulemd-2.0/modulemd-module-index.h
```diff
@@ -71,6 +71,35 @@ G_DECLARE_FINAL_TYPE (
...
+ *
+ * The prototype of a read handler.
+ *
+ * The read handler is called when the parser needs to read more bytes from the
+ * source. The handler should write not more than @size bytes to the @buffer.
+ * The number of written bytes should be set to the @size_read variable.
+ *
+ * This handler is identical to a
+ * [yaml_read_handler_t](https://github.com/yaml/libyaml/blob/master/include/yaml.h#L988)
+ * but is included here to avoid depending on yaml.h in modulemd headers.
```
Conan-Kudo: To be clear, this is about preventing the exposure of libyaml interfaces as an effect of including the libmodulemd headers, right?

sgallagher: Yes. It's mostly to avoid needing to have `#include <yaml.h>` in public headers and thereby require `libyaml-devel` to be installed when developing against `libmodulemd-devel`.

Conan-Kudo: 👍

ignatenkobrain:
To be honest, this seems bad design choice..

It would be much better to make integration with GInputStream instead.

sgallagher:
@ignatenkobrain That definitely seems like overkill, given that the specific goal right now is "get something that I can use with createrepo_c". That project does not use GInputStream.

PR #265, opened 2019-04-03
Fixes #186: Provide .equals for two ModuleStream objects

### LINKED ISSUE #186: Provide method which compares two ModuleStream objects

OrionStar25:
Fixes #186

I think we need a `modulemd.translation.equals()` for the object in the parent class. Also, I didn't find any definition for what's contained in the 2 child classes. So I couldn't decide what all equality checks to perform. Can you please help me here?

### DIFF
--- a/modulemd/v2/include/modulemd-2.0/modulemd-component.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-component.h
@@ -44,6 +44,18 @@ struct _ModulemdComponentClass
 };
 
 /**
+ * modulemd_component_equals_wrapper:
+ * @a: A void pointer
+ * @b: A void pointer
+ *
+ * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
+ *
+ * Since: 2.3
+ */
+gboolean
+modulemd_component_equals_wrapper (const void *a, const void *b);
+
+/**
  * modulemd_component_equals:
  * @self_1: A #ModulemdComponent object
  * @self_2: A #ModulemdComponent object
--- a/modulemd/v2/include/modulemd-2.0/modulemd-component.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-component.h
@@ -44,18 +44,6 @@ struct _ModulemdComponentClass
 };
 
 /**
- * modulemd_component_equals_wrapper:
- * @a: A void pointer
- * @b: A void pointer
- *
- * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
- *
- * Since: 2.3
- */
-gboolean
-modulemd_component_equals_wrapper (const void *a, const void *b);
-
-/**
  * modulemd_component_equals:
  * @self_1: A #ModulemdComponent object
  * @self_2: A #ModulemdComponent object
--- a/modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
@@ -80,9 +80,12 @@ struct _ModulemdModuleStreamClass
                                        const gchar *module_name,
                                        const gchar *stream_name);
 
-  /* Padding to allow adding up to 8 new virtual functions without
+  gboolean (*equals) (ModulemdModuleStream *self_1,
+                      ModulemdModuleStream *self_2);
+
+  /* Padding to allow adding up to 7 new virtual functions without
    * breaking ABI. */
-  gpointer padding[8];
+  gpointer padding[7];
 };
 
 
@@ -197,6 +200,22 @@ modulemd_module_stream_read_stream (FILE *stream,
                                     const gchar *module_name,
                                     const gchar *module_stream,
                                     GError **error);
+
+
+/**
+ * modulemd_module_stream_equals:
+ * @self_1: (in): A #ModulemdModuleStream.
+ * @self_2: (in): A #ModulemdModuleStream.
+ *
+ * Checks if self_1 and self_2 are identical objects.
+ *
+ * Returns: TRUE, If both objects are equal. FALSE, otherwise.
+ *
+ * Since: 2.3
+ */
+gboolean
+modulemd_module_stream_equals (ModulemdModuleStream *self_1,
+                               ModulemdModuleStream *self_2);
 
 
 /**
--- a/modulemd/v2/include/modulemd-2.0/modulemd-profile.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-profile.h
@@ -31,19 +31,6 @@ G_DECLARE_FINAL_TYPE (
 
 
 /**
- * modulemd_profile_equals_wrapper:
- * @a: A void pointer
- * @b: A void pointer
- *
- * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
- *
- * Since: 2.3
- */
-gboolean
-modulemd_profile_equals_wrapper (const void *a, const void *b);
-
-
-/**
  * modulemd_profile_equals:
  * @self_1: A #ModulemdProfile
  * @self_2: A #ModulemdProfile
--- a/modulemd/v2/include/modulemd-2.0/modulemd-profile.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-profile.h
@@ -31,6 +31,19 @@ G_DECLARE_FINAL_TYPE (
 
 
 /**
+ * modulemd_profile_equals_wrapper:
+ * @a: A void pointer
+ * @b: A void pointer
+ *
+ * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
+ *
+ * Since: 2.3
+ */
+gboolean
+modulemd_profile_equals_wrapper (const void *a, const void *b);
+
+
+/**
  * modulemd_profile_equals:
  * @self_1: A #ModulemdProfile
  * @self_2: A #ModulemdProfile
--- a/modulemd/v2/include/modulemd-2.0/modulemd-rpm-map-entry.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-rpm-map-entry.h
@@ -68,34 +68,6 @@ modulemd_rpm_map_entry_copy (ModulemdRpmMapEntry *self);
 
 
 /**
- * modulemd_rpm_map_entry_equals_wrapper:
- * @a: const void pointer
- * @b: const void pointer
- *
- * Returns: TRUE if the two entries contain equivalent data. FALSE if they
- * differ.
- *
- * Since: 2.5
- */
-gboolean
-modulemd_rpm_map_entry_equals_wrapper (const void *a, const void *b);
-
-
-/**
- * modulemd_RpmMapEntry_hash_table_equals_wrapper:
- * @a: const void pointer
- * @b: const void pointer
- *
- * Returns: TRUE if the two entries contain equivalent data. FALSE if they
- * differ.
- *
- * Since: 2.5
- */
-gboolean
-modulemd_RpmMapEntry_hash_table_equals_wrapper (const void *a, const void *b);
-
-
-/**
  * modulemd_rpm_map_entry_equals:
  * @self: A #ModuleRpmMapEntry
  * @other: Another #ModulemdRpmMapEntry
--- a/modulemd/v2/include/modulemd-2.0/modulemd-rpm-map-entry.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-rpm-map-entry.h
@@ -68,6 +68,34 @@ modulemd_rpm_map_entry_copy (ModulemdRpmMapEntry *self);
 
 
 /**
+ * modulemd_rpm_map_entry_equals_wrapper:
+ * @a: const void pointer
+ * @b: const void pointer
+ *
+ * Returns: TRUE if the two entries contain equivalent data. FALSE if they
+ * differ.
+ *
+ * Since: 2.5
+ */
+gboolean
+modulemd_rpm_map_entry_equals_wrapper (const void *a, const void *b);
+
+
+/**
+ * modulemd_RpmMapEntry_hash_table_equals_wrapper:
+ * @a: const void pointer
+ * @b: const void pointer
+ *
+ * Returns: TRUE if the two entries contain equivalent data. FALSE if they
+ * differ.
+ *
+ * Since: 2.5
+ */
+gboolean
+modulemd_RpmMapEntry_hash_table_equals_wrapper (const void *a, const void *b);
+
+
+/**
  * modulemd_rpm_map_entry_equals:
  * @self: A #ModuleRpmMapEntry
  * @other: Another #ModulemdRpmMapEntry
--- a/modulemd/v2/include/modulemd-2.0/modulemd-service-level.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-service-level.h
@@ -151,17 +151,5 @@ gboolean
 modulemd_service_level_equals (ModulemdServiceLevel *self_1,
                                ModulemdServiceLevel *self_2);
 
-/**
- * modulemd_service_level_equals_wrapper:
- * @a: A void pointer
- * @b: A void pointer
- *
- * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
- *
- * Since: 2.3
- */
-gboolean
-modulemd_service_level_equals_wrapper (const void *a, const void *b);
-
 
 G_END_DECLS
--- a/modulemd/v2/include/modulemd-2.0/modulemd-service-level.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-service-level.h
@@ -151,5 +151,17 @@ gboolean
 modulemd_service_level_equals (ModulemdServiceLevel *self_1,
                                ModulemdServiceLevel *self_2);
 
+/**
+ * modulemd_service_level_equals_wrapper:
+ * @a: A void pointer
+ * @b: A void pointer
+ *
+ * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
+ *
+ * Since: 2.3
+ */
+gboolean
+modulemd_service_level_equals_wrapper (const void *a, const void *b);
+
 
 G_END_DECLS
--- a/modulemd/v2/include/private/modulemd-component-private.h
+++ b/modulemd/v2/include/private/modulemd-component-private.h
@@ -127,7 +127,7 @@ modulemd_component_emit_yaml_build_common (ModulemdComponent *self,
  *
  * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
  *
- * Since: 2.3
+ * Since: 2.5
  */
 gboolean
 modulemd_component_equals_wrapper (const void *a, const void *b);
--- a/modulemd/v2/include/private/modulemd-component-private.h
+++ b/modulemd/v2/include/private/modulemd-component-private.h
@@ -119,3 +119,15 @@ gboolean
 modulemd_component_emit_yaml_build_common (ModulemdComponent *self,
                                            yaml_emitter_t *emitter,
                                            GError **error);
+
+/**
+ * modulemd_component_equals_wrapper:
+ * @a: A void pointer
+ * @b: A void pointer
+ *
+ * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
+ *
+ * Since: 2.3
+ */
+gboolean
+modulemd_component_equals_wrapper (const void *a, const void *b);
--- a/modulemd/v2/include/private/modulemd-profile-private.h
+++ b/modulemd/v2/include/private/modulemd-profile-private.h
@@ -65,3 +65,15 @@ gboolean
 modulemd_profile_emit_yaml (ModulemdProfile *self,
                             yaml_emitter_t *emitter,
                             GError **error);
+
+/**
+ * modulemd_profile_equals_wrapper:
+ * @a: A void pointer
+ * @b: A void pointer
+ *
+ * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
+ *
+ * Since: 2.3
+ */
+gboolean
+modulemd_profile_equals_wrapper (const void *a, const void *b);
--- a/modulemd/v2/include/private/modulemd-profile-private.h
+++ b/modulemd/v2/include/private/modulemd-profile-private.h
@@ -73,7 +73,7 @@ modulemd_profile_emit_yaml (ModulemdProfile *self,
  *
  * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
  *
- * Since: 2.3
+ * Since: 2.5
  */
 gboolean
 modulemd_profile_equals_wrapper (const void *a, const void *b);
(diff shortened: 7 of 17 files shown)


modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
```diff
@@ -82,7 +85,7 @@ struct _ModulemdModuleStreamClass
 
   /* Padding to allow adding up to 8 new virtual functions without
    * breaking ABI. */
-  gpointer padding[8];
+  gpointer padding[10];
```
sgallagher: This number must decrease when adding a new function. The padding exists to ensure that the data structure size remains the same when adding new functions.

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
```diff
@@ -68,6 +68,9 @@ struct _ModulemdModuleStreamClass
                                  const gchar *module_name,
                                  const gchar *module_stream);
 
+  gboolean (*equals) (ModulemdModuleStream *self_1,
```
sgallagher: This needs to be added at the end of the structure, just before the padding. Otherwise it breaks ABI.

And I just realized this may have happened in the other `.equals()` patches without my noticing and I need to fix that.

sgallagher: The reason is that any software that was compiled against an earlier version of libmodulemd that may have been accessing this class directly needs to have the offset of the functions remain at the same place. Putting this above them means that code compiled against an earlier version trying to access `stream_class>validate()` will now be incorrectly calling `stream_class->equals()`.

modulemd/v2/modulemd-module-stream.c
```diff
@@ -297,6 +297,61 @@ modulemd_module_stream_finalize (GObject *object)
...
+
+  if (g_strcmp0 (modulemd_module_stream_get_context (self_1),
+                 modulemd_module_stream_get_context (self_2)) != 0)
+    return FALSE;
+
+  if (g_strcmp0 (modulemd_module_stream_get_arch (self_1),
+                 modulemd_module_stream_get_arch (self_2)) != 0)
+    return FALSE;
+
+  //need a function for this
```
sgallagher: Yes, this should be done first.

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
sgallagher: Turns out that `ModulemdComponents` and `ModulemdDefaults` both had the `.equals()` in the right place, but the padding wasn't decremented. I just pushed https://github.com/fedora-modularity/libmodulemd/commit/06611a77ac4da954d125b42f2452165c6f544879 to fix that.

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
OrionStar25: Oh! I didn't decrement this in any of the new functions I added :/. I'll fix it if you want me to. I thought that array helps to accommodate new virtual functions.

So how will we add any new virtual functions when the padding will reach 0?

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
OrionStar25: So does this mean that softwares look for linewise implementation in libmodulemd rather than syntactically checking against the function calls?

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
sgallagher: I just fixed that in ModulemdComponent and ModulemdDefaults. If you know of any other places besides that, they should also be fixed.

If we run out of padding, we need to make a new major release (libmodulemd 3.x). We will need to bump the soname to libmodulemd.so.3 and any package that needs to consume libmodulemd will need to be rebuilt against the new library. This is something to avoid as much as possible (ideally only when it's discovered that the current API has a critical flaw that can only be solved by replacing it entirely, such as what happened between 1.x and 2.x). Hence, we keep the padding in place to reduce the risk of this. It's unlikely that we will hit zero on any of them, though.

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
sgallagher: That's an easy way to think about it, but really what happens is that the object is converted by the compiler into a contiguous memory block. Programs compiled against libmodulemd would look at an offset from the start of that block to retrieve a [32|64]-bit pointer to the function to call. If we reorder those pointers in the C file, that translates into changing the offset from the start of that memory location. So if an application was looking for a function pointer at block_start+4 which was "validate()" when this application was compiled, but now block_start+4 is returning the pointer to "equals()", the program won't behave correctly.

OrionStar25: Oh okay, got it! I'll make all the changes soon

modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
OrionStar25: So its not needed anywhere else since the others weren't getting inherited.

modulemd/v2/modulemd-module-stream-v2.c
```diff
@@ -106,6 +106,129 @@ modulemd_module_stream_v2_finalize (GObject *object)
...
+  if (!modulemd_hash_table_equals (v2_self_1->servicelevels,
+                                   v2_self_2->servicelevels,
+                                   modulemd_service_level_equals))
+    {
+      return FALSE;
+    }
+
+  /* 
+  //need an equals function for Modulemd.RpmMapEntry
+  // < string, GHashTable <string, Modulemd.RpmMapEntry> >
```
OrionStar25: Equality checking for this is a little tricky. How can I pass `modulemd_rpm_map_entry_equals` ?

OrionStar25:
Also, by using `modulemd_hash_table_equals` , lots of typecasting warnings are being produced. Will I require a wrapper function for all of them? :O

modulemd/v2/modulemd-module-stream-v2.c
sgallagher: You need to create a wrapper function for the comparison.

sgallagher:
> Also, by using `modulemd_hash_table_equals` , lots of typecasting warnings are being produced. Will I require a wrapper function for all of them? :O

Yes

modulemd/v2/include/modulemd-2.0/modulemd-component.h
```diff
@@ -43,6 +43,18 @@ struct _ModulemdComponentClass
...
+ * modulemd_component_equals_wrapper:
+ * @a: A void pointer
+ * @b: A void pointer
+ *
+ * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
+ *
+ * Since: 2.3
+ */
+gboolean
+modulemd_component_equals_wrapper (const void *a, const void *b);
```
sgallagher: Do not make these wrappers public. They aren't meaningful for the external API. If you need them in multiple places (e.g. both ModuleStream objects), then these belong in the private headers like `modulemd/v2/private/modulemd-component-private.h`)

modulemd/v2/include/modulemd-2.0/modulemd-profile.h
```diff
@@ -30,6 +30,19 @@ G_DECLARE_FINAL_TYPE (
...
+ * modulemd_profile_equals_wrapper:
+ * @a: A void pointer
+ * @b: A void pointer
+ *
+ * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
+ *
+ * Since: 2.3
+ */
+gboolean
+modulemd_profile_equals_wrapper (const void *a, const void *b);
```
sgallagher: Same as for components; do not make this a public function.

modulemd/v2/include/modulemd-2.0/modulemd-service-level.h
```diff
@@ -151,5 +151,17 @@ gboolean
...
+ * modulemd_service_level_equals_wrapper:
+ * @a: A void pointer
+ * @b: A void pointer
+ *
+ * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
+ *
+ * Since: 2.3
+ */
+gboolean
+modulemd_service_level_equals_wrapper (const void *a, const void *b);
```
sgallagher: Don't make this public either.

modulemd/v2/modulemd-component.c
```diff
@@ -48,6 +48,14 @@ enum
 static GParamSpec *properties[N_PROPS];
 
 
+gboolean
+modulemd_component_equals_wrapper (const void *a, const void *b)
+{
+  return modulemd_component_equals ((ModulemdComponent *)a,
```
sgallagher: We want to `assert()` on programming errors here.

Also, GLib has a cast mechanism for objects that we should prefer:

```suggestion
  g_return_val_if_false (MODULEMD_IS_COMPONENT(a), FALSE);
  g_return_val_if_false (MODULEMD_IS_COMPONENT(b), FALSE);
  
  return modulemd_component_equals (MODULEMD_COMPONENT(a),
```

modulemd/v2/modulemd-component.c
```diff
@@ -48,6 +48,14 @@ enum
...
+                                    (ModulemdComponent *)b);
```
sgallagher:
```suggestion
                                    MODULEMD_COMPONENT(b));
```

modulemd/v2/modulemd-module-stream-v1.c
```diff
@@ -873,6 +873,120 @@ modulemd_module_stream_v1_get_xmd (ModulemdModuleStreamV1 *self)
...
+      return FALSE;
+    }
+
+  if (!modulemd_hash_table_equals (
+        v1_self_1->runtime_deps, v1_self_2->runtime_deps, g_str_equal))
+    {
+      return FALSE;
+    }
+
+  if (g_variant_compare (v1_self_1->xmd, v1_self_2->xmd) != 0)
```
sgallagher: `g_variant_compare()` won't work here. These aren't guaranteed to be basic types (in fact, they are pretty much guaranteed *not* to be).

From the [docs](https://developer.gnome.org/glib/stable/glib-GVariant.html#g-variant-compare): "It is a programmer error to attempt to compare container values or two values that have types that are not exactly equal. For example, you cannot compare a 32-bit signed integer with a 32-bit unsigned integer. Also note that this function is not particularly well-behaved when it comes to comparison of doubles; in particular, the handling of incomparable values (ie: NaN) is undefined.

If you only require an equality comparison, [`g_variant_equal()`](https://developer.gnome.org/glib/stable/glib-GVariant.html#g-variant-equal) is more general."

modulemd/v2/modulemd-module-stream-v2.c
sgallagher: Hmm, looking at this more, you're right. This is going to be a bit trickier than I was originally thinking. You will probably need to write a helper function here to handle this case, rather than using `modulemd_hash_table_equals()` at the top-level here.

modulemd/v2/modulemd-module-stream-v2.c
```diff
@@ -106,6 +106,130 @@ modulemd_module_stream_v2_finalize (GObject *object)
...
+
+  for (guint i = 0; i < v2_self_1->dependencies->len; i++)
+    {
+      if (!modulemd_dependencies_equals (
+            g_ptr_array_index (v2_self_1->dependencies, i),
+            g_ptr_array_index (v2_self_2->dependencies, i)))
+        return FALSE;
+    }
+
+  if (g_variant_compare (v2_self_1->xmd, v2_self_2->xmd) != 0)
```
sgallagher: Same as with V1; you need to use `g_variant_equals()` here instead.

modulemd/v2/modulemd-module-stream.c
```diff
@@ -297,6 +297,62 @@ modulemd_module_stream_finalize (GObject *object)
...
+  if (g_strcmp0 (modulemd_module_stream_get_context (self_1),
+                 modulemd_module_stream_get_context (self_2)) != 0)
+    return FALSE;
+
+  if (g_strcmp0 (modulemd_module_stream_get_arch (self_1),
+                 modulemd_module_stream_get_arch (self_2)) != 0)
+    return FALSE;
+
+  /*need a function for this
+  if(!modulemd_translation_equals (modulemd_module_stream_get_translation (self_1),
```
sgallagher: Skip translation equivalence. It's safe to do because setting the translation is an internal-only function that is handled automatically when adding a module to an index. We only need to know that the untranslated module is equal in this comparison.

sgallagher: For additional information: This is really just a back-reference to the translation object stored in the ModulemdModule object this stream belongs to. So as long as the module name is the same, this object is guaranteed to just be a pointer to that translation object, and thus equal.

modulemd/v2/modulemd-profile.c
```diff
@@ -46,6 +46,13 @@ enum
 static GParamSpec *properties[N_PROPS];
 
 
+gboolean
+modulemd_profile_equals_wrapper (const void *a, const void *b)
+{
+  return modulemd_profile_equals ((ModulemdProfile *)a, (ModulemdProfile *)b);
```
sgallagher:
```suggestion
  return modulemd_profile_equals (MODULEMD_PROFILE(a), (MODULEMD_PROFILE(b));
```

modulemd/v2/modulemd-service-level.c
```diff
@@ -55,6 +55,14 @@ modulemd_service_level_new (const gchar *name)
 }
 
 
+gboolean
+modulemd_service_level_equals_wrapper (const void *a, const void *b)
+{
+  return modulemd_service_level_equals ((ModulemdServiceLevel *)a,
```
sgallagher:
```suggestion
  return modulemd_service_level_equals (MODULEMD_SERVICE_LEVEL(a),
```

modulemd/v2/modulemd-service-level.c
```diff
@@ -55,6 +55,14 @@ modulemd_service_level_new (const gchar *name)
...
+                                        (ModulemdServiceLevel *)b);
```
sgallagher:
```suggestion
                                        (MODULEMD_SERVICE_LEVEL(b));
```

sgallagher:
> I think we need a `modulemd.translation.equals()` for the object in the parent class.

We do not. See my comments on the changes themselves for an explanation.

OrionStar25:
I get these warnings for all of those assertions:
`warning: passing argument 1 of ‘MODULEMD_IS_COMPONENT’ discards ‘const’ qualifier from pointer target type`

Should I typecast and send it in?

sgallagher:
> I get these warnings for all of those assertions:
> `warning: passing argument 1 of ‘MODULEMD_IS_COMPONENT’ discards ‘const’ qualifier from pointer target type`
> 
> Should I typecast and send it in?

Yes, in this case it's safe to cast it.

OrionStar25:
@sgallagher I'll add tests once you approve this. Please review.

modulemd/v2/include/modulemd-2.0/modulemd-rpm-map-entry.h
```diff
@@ -67,6 +67,34 @@ ModulemdRpmMapEntry *
...
+ * @a: const void pointer
+ * @b: const void pointer
+ *
+ * Returns: TRUE if the two entries contain equivalent data. FALSE if they
+ * differ.
+ *
+ * Since: 2.5
+ */
+gboolean
+modulemd_rpm_map_entry_equals_wrapper (const void *a, const void *b);
```
sgallagher: Do not make the wrapper functions public API. They should be either static functions within the C file that consumes them or else in the headers in the `include/private/` path so that they are visible only to internal consumers. There's no reason to expose these to the consumers of libmodulemd.

modulemd/v2/include/modulemd-2.0/modulemd-rpm-map-entry.h
```diff
@@ -67,6 +67,34 @@ ModulemdRpmMapEntry *
...
+ * @a: const void pointer
+ * @b: const void pointer
+ *
+ * Returns: TRUE if the two entries contain equivalent data. FALSE if they
+ * differ.
+ *
+ * Since: 2.5
+ */
+gboolean
+modulemd_RpmMapEntry_hash_table_equals_wrapper (const void *a, const void *b);
```
sgallagher: See above: don't make the wrappers into public API.

modulemd/v2/include/private/modulemd-component-private.h
```diff
@@ -119,3 +119,15 @@ gboolean
...
                                            GError **error);
+
+/**
+ * modulemd_component_equals_wrapper:
+ * @a: A void pointer
+ * @b: A void pointer
+ *
+ * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
+ *
+ * Since: 2.3
```
sgallagher: This is correctly private API, but note that `Since: 2.3` will no longer be accurate..
```suggestion
 * Since: 2.5
```

modulemd/v2/include/private/modulemd-profile-private.h
```diff
@@ -65,3 +65,15 @@ gboolean
...
                             GError **error);
+
+/**
+ * modulemd_profile_equals_wrapper:
+ * @a: A void pointer
+ * @b: A void pointer
+ *
+ * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
+ *
+ * Since: 2.3
```
sgallagher: This is correctly private API, but note that `Since: 2.3` will no longer be accurate..
```suggestion
 * Since: 2.5
```

modulemd/v2/include/private/modulemd-service-level-private.h
```diff
@@ -65,3 +65,15 @@ gboolean
...
                                   GError **error);
+
+/**
+ * modulemd_service_level_equals_wrapper:
+ * @a: A void pointer
+ * @b: A void pointer
+ *
+ * Returns: TRUE, if both the pointers are equal. FALSE, otherwise
+ *
+ * Since: 2.3
```
sgallagher: This is correctly private API, but note that `Since: 2.3` will no longer be accurate..
```suggestion
 * Since: 2.5
```

modulemd/v2/modulemd-component.c
```diff
@@ -48,6 +48,18 @@ enum
...
 
 
+gboolean
+modulemd_component_equals_wrapper (const void *a, const void *b)
+{
+  g_return_val_if_fail (MODULEMD_IS_COMPONENT ((ModulemdComponent *)a), FALSE);
+  g_return_val_if_fail (MODULEMD_IS_COMPONENT ((ModulemdComponent *)b), FALSE);
+
+  return modulemd_component_equals (
+    MODULEMD_COMPONENT ((ModulemdComponent *)a),
```
sgallagher: You don't need to wrap this in `MODULEMD_COMPONENT()` because you've already validated it with `MODULEMD_IS_COMPONENT()` above. So just cast it directly (needed to avoid the `const` warnings) and skip the extra cast.

```suggestion
    (ModulemdComponent *)a,
```

modulemd/v2/modulemd-module-stream-v2.c
```diff
@@ -106,6 +106,130 @@ modulemd_module_stream_v2_finalize (GObject *object)
...
+        modulemd_RpmMapEntry_hash_table_equals_wrapper))
+    {
+      return FALSE;
+    }
+
+
+  if (v2_self_1->dependencies->len != v2_self_2->dependencies->len)
+    return FALSE;
+
+  for (guint i = 0; i < v2_self_1->dependencies->len; i++)
```
sgallagher: Please include a comment here explaining that ordering is important for the dependencies, so each array index must be the same.

modulemd/v2/modulemd-profile.c
```diff
@@ -46,6 +46,17 @@ enum
 static GParamSpec *properties[N_PROPS];
 
 
+gboolean
+modulemd_profile_equals_wrapper (const void *a, const void *b)
+{
+  g_return_val_if_fail (MODULEMD_IS_PROFILE ((ModulemdProfile *)a), FALSE);
+  g_return_val_if_fail (MODULEMD_IS_PROFILE ((ModulemdProfile *)b), FALSE);
+
+  return modulemd_profile_equals (MODULEMD_PROFILE ((ModulemdProfile *)a),
```
sgallagher: Same as above; you don't need to double-cast here.
```suggestion
  return modulemd_profile_equals ((ModulemdProfile *)a,
```

modulemd/v2/modulemd-rpm-map-entry.c
```diff
@@ -65,6 +65,27 @@ modulemd_rpm_map_entry_new (const gchar *name,
...
+gboolean
+modulemd_rpm_map_entry_equals_wrapper (const void *a, const void *b)
+{
+  g_return_val_if_fail (MODULEMD_IS_RPM_MAP_ENTRY ((ModulemdRpmMapEntry *)a),
+                        FALSE);
+  g_return_val_if_fail (MODULEMD_IS_RPM_MAP_ENTRY ((ModulemdRpmMapEntry *)b),
+                        FALSE);
+
+  return modulemd_rpm_map_entry_equals (
+    MODULEMD_RPM_MAP_ENTRY ((ModulemdRpmMapEntry *)a),
```
sgallagher:
```suggestion
    (ModulemdRpmMapEntry *)a,
```

modulemd/v2/modulemd-service-level.c
```diff
@@ -55,6 +55,20 @@ modulemd_service_level_new (const gchar *name)
...
+gboolean
+modulemd_service_level_equals_wrapper (const void *a, const void *b)
+{
+  g_return_val_if_fail (MODULEMD_IS_SERVICE_LEVEL ((ModulemdServiceLevel *)a),
+                        FALSE);
+  g_return_val_if_fail (MODULEMD_IS_SERVICE_LEVEL ((ModulemdServiceLevel *)b),
+                        FALSE);
+
+  return modulemd_service_level_equals (
+    MODULEMD_SERVICE_LEVEL ((ModulemdServiceLevel *)a),
```
sgallagher:
```suggestion
    (ModulemdServiceLevel *)a,
```

sgallagher:
Changes look good now, but please add tests next.

OrionStar25:
@sgallagher Please review.

sgallagher:
@OrionStar25 Initial test looks good. Please add tests for every field in the two modulestream versions.

OrionStar25:
@sgallagher  Please review. Added tests for v1 and v2. I made 2 new functions though most of the initial part is repetitive. If all of it can be made into a loop, do tell me please!

sgallagher:
I think this all looks fine. Assuming the CI tests complete successfully, I'll squash and merge these commits. Great work!

Issue #286, opened 2019-05-07
Add the ability to remove entries from the dependencies array


mprahl:
MBS currently uses libmodulemd v1. As part of module stream expansion, MBS creates a copy of the source `Module` object, and runs `set_dependencies((new_dep, ))` with the resulting dependency combination. This is repeated for every dependency combination.

In other parts of the code, MBS also modifies the dependency objects directly with things such as `deps.set_requires(expanded)`.

As part of the migration to libmodulde v2, MBS needs the ability to remove `Modulemd.Dependencies` objects from the dependencies array. This can be clearing all existing dependencies, or removing a single dependency entry at a time. Although the latter is preferred, the former would also work.

sgallagher:
Scratch-build with the referenced PR applied: https://koji.fedoraproject.org/koji/taskinfo?taskID=34693382

PR #287, opened 2019-05-07
ModuleStreamV2: add clear_dependencies() and remove_dependencies()

sgallagher:
Adds two new methods to make management of dependencies on a stream
simpler.

Fixes: https://github.com/fedora-modularity/libmodulemd/issues/286

### DIFF
--- a/modulemd/v2/modulemd-module-stream-v2.c
+++ b/modulemd/v2/modulemd-module-stream-v2.c
@@ -777,6 +777,38 @@ modulemd_module_stream_v2_replace_dependencies (ModulemdModuleStreamV2 *self,
 }
 
 
+void
+modulemd_module_stream_v2_clear_dependencies (ModulemdModuleStreamV2 *self)
+{
+  g_return_if_fail (MODULEMD_IS_MODULE_STREAM_V2 (self));
+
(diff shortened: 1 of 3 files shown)


modulemd/v2/modulemd-module-stream-v2.c
```diff
@@ -777,6 +777,38 @@ modulemd_module_stream_v2_replace_dependencies (ModulemdModuleStreamV2 *self,
...
+
+
+void
+modulemd_module_stream_v2_remove_dependencies (ModulemdModuleStreamV2 *self,
+                                               ModulemdDependencies *deps)
+{
+  guint index;
+  g_return_if_fail (MODULEMD_IS_MODULE_STREAM_V2 (self));
+
+  while (g_ptr_array_find_with_equal_func (
```
mprahl: @sgallagher, I'm not sure if this is an issue, but if there are multiple dependencies that are equal, then they will all be removed. I was expecting just the first one that was found to be removed.

For example:
```python
mmd.add_dependencies(mmd.get_dependencies()[0].copy())
assert len(mmd.get_dependencies()) == 2
mmd.remove_dependencies(mmd.get_dependencies()[0])
assert len(mmd.get_dependencies()) == 0
```

Issue #290, opened 2019-05-08
Can't dump a module to string when xmd contains a populated list from Python


mprahl:
From Python, if a modulemd has an `xmd` that contains a populated list (empty lists work), it can't be dumped to a string.

Here is the reproducer:
```python
import requests
import gi
gi.require_version("Modulemd", "2.0")
from gi.repository import Modulemd

rv = requests.get(
    'https://pagure.io/fm-orchestrator/raw/master/f/tests/staged_data/formatted_testmodule.yaml')
rv.encoding = 'utf-8'
mmd_str = rv.text
mmd = Modulemd.ModuleStream.read_string(mmd_str, True)
mmd = mmd.upgrade(2)

xmd = mmd.get_xmd()
xmd["something"] = ["foo", "bar"]
mmd.set_xmd(xmd)

index = Modulemd.ModuleIndex()
index.add_module_stream(mmd)
index.dump_to_string()
```

The output is:
```
gi.repository.GLib.Error: modulemd-yaml-error-quark: Unhandled variant type: "v" (4)
```

sgallagher:
Scratch-build containing all patches since 2.3.1 for Fedora 29
 https://koji.fedoraproject.org/koji/taskinfo?taskID=34769645

sgallagher:
Previous scratch-build failed because I forgot to push some test files into the patch.

https://koji.fedoraproject.org/koji/taskinfo?taskID=34770448 should work now.

PR #291, opened 2019-05-10
Fix issues with XMD handling

sgallagher:
Fix memory issues with XMD handling

This cleans up the usage of g_variant_ref_sink() and re-syncs the
StreamV1 and StreamV2 to both make a proper deep copy of the XMD
when setting it.

The overrides weren't creating the GVariant arrays and dictionaries
quite correctly. Also, the set_xmd() function had the input marked
as (transfer full), which was incorrect and causing refcounting
issues.

Fixes https://github.com/fedora-modularity/libmodulemd/issues/290

### DIFF
--- a/bindings/python/gi/overrides/Modulemd.py
+++ b/bindings/python/gi/overrides/Modulemd.py
@@ -25,7 +25,6 @@
 # For more information on free software, see
 # <https://www.gnu.org/philosophy/free-sw.en.html>.
 
-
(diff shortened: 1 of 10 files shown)


sgallagher:
Hmm, some of the tests are failing due to https://github.com/fedora-modularity/libmodulemd/issues/292

I'll need to address that, but in the meantime I'm going to disable those tests.

PR #295, opened 2019-05-13
Undeprecate ModuleStream.get_nsvc_as_string()

sgallagher:

### DIFF
--- a/modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-module-stream.h
@@ -376,13 +376,17 @@ modulemd_module_stream_get_arch (ModulemdModuleStream *self);
  * modulemd_module_stream_get_nsvc_as_string: (rename-to modulemd_module_stream_get_nsvc)
  * @self: (in): This #ModulemdModuleStream.
  *
+ * Retrieves a representation of the module name, stream name, version and
+ * context of this #ModuleStream in the form
+ * `module_name:stream_name:version:context`. Note that this excludes the
+ * architecture of the module stream and as such is not guaranteed to be unique
+ * within a repository.
+ *
  * Returns: (transfer full): The NSVC (name:stream:version[:context]) of this
  * module stream. NULL if module name or stream stream is unknown.
  *
  * Since: 2.0
- * Deprecated: 2.2
  */
-MMD_DEPRECATED_FOR (modulemd_module_stream_get_NSVCA_as_string)
 gchar *
 modulemd_module_stream_get_nsvc_as_string (ModulemdModuleStream *self);
 


sgallagher:
MBS needs this in a variety of places, none of which are lookups. @mprahl justified it to me and I’m dropping the deprecation to avoid lots of junk in the MBS logs.

Conan-Kudo:
@sgallagher What was the justification, because "junk in the MBS logs" doesn't seem like enough of a justification to endorse potentially stupid API usage.

mprahl:
@Conan-Kudo the NSVC uniquely identifies a module build in MBS. `.get_nsvc()` is used to avoid the need to do the following throughout the code (mostly handy for logging):
```python
nsvc = ":".join((mmd.get_module_name(), mmd.get_stream_name(), str(mmd.get_version), mmd.get_context()))
```

I don't see how using `.get_nsvc()` for our use-cases is "stupid API usage". If for whatever reason it's a bad idea to undeprecate `.get_nsvc()`, we can resort to the example shown above in MBS.

sgallagher:
@mprahl
What I think @Conan-Kudo is asking is whether there is a good reason not to be using `.get_NSVCA()` everywhere that you are currently using `.get_nsvc()`. Or, phrased differently: Is there a reason that also having `:arch` at the end of the string would cause problems?

sgallagher:
@mprahl Would you mind chiming in with an answer here?

mprahl:
> @mprahl
> What I think @Conan-Kudo is asking is whether there is a good reason not to be using `.get_NSVCA()` everywhere that you are currently using `.get_nsvc()`. Or, phrased differently: Is there a reason that also having `:arch` at the end of the string would cause problems?

MBS is not expecting an arch in every use-case of `.get_nsvc()`. The arch at that point shouldn't be set anyways, but there's no code in MBS checking that the arch is set by the user by accident.

sgallagher:
> > @mprahl
> > What I think @Conan-Kudo is asking is whether there is a good reason not to be using `.get_NSVCA()` everywhere that you are currently using `.get_nsvc()`. Or, phrased differently: Is there a reason that also having `:arch` at the end of the string would cause problems?
> 
> MBS is not expecting an arch in every use-case of `.get_nsvc()`. The arch at that point shouldn't be set anyways, but there's no code in MBS checking that the arch is set by the user by accident.

What would it take to make MBS aware of that? I think we really should be standardizing on the guaranteed-unique value.

mprahl:
> > > @mprahl
> > > What I think @Conan-Kudo is asking is whether there is a good reason not to be using `.get_NSVCA()` everywhere that you are currently using `.get_nsvc()`. Or, phrased differently: Is there a reason that also having `:arch` at the end of the string would cause problems?
> > 
> > 
> > MBS is not expecting an arch in every use-case of `.get_nsvc()`. The arch at that point shouldn't be set anyways, but there's no code in MBS checking that the arch is set by the user by accident.
> 
> What would it take to make MBS aware of that? I think we really should be standardizing on the guaranteed-unique value.

The arch is only applicable in MBS when generating the arch specific modulemd files for the Koji Content Generator builds. This is mentioned in the modulemd spec at:
https://github.com/fedora-modularity/libmodulemd/blob/master/spec.v2.yaml#L32

Like I said, we don't use `.get_nsvc()` for lookups in libmodulemd, we just it for convenience since the NSVC is what identifies a unique build entry in the MBS database.

sgallagher:
@mprahl OK, I think that was the part I wasn't getting communicated. That the NSVC is a key in a *different* system (that is explicitly and intentionally not using the arch for valid reasons).

Issue #188, opened 2019-02-14
Enhance IndexMerger to use equality checks


sgallagher:
Right now, we use a highly-simplified comparison when merging ModuleStream objects from two different repositories. We check whether they have the same NSVC and then assume them to be the same content. It would be better to perform a full equality check.

This depends on https://github.com/fedora-modularity/libmodulemd/issues/186

irenge:
Hi 
May I work on this issue ?
Thanks

sgallagher:
@irenge Please go ahead.

athira-selvam:
Hi @irenge are you working on this issue?

irenge:
haha ! I thought the issue was closed . I am going back to it . R u interested in it ?

sgallagher:
This isn't eligible to be worked on until #186 is merged. That issue needs to be resolved first.

athira-selvam:
@sgallagher Hi, All beginner issues have been claimed. Is it okay to work on those issues? I have already copied tests for the depends_on_stream function to C. And I added a test for an empty stream. But that was in python. Can I work on copying that test to C and also copying some more functions from modulestream.py to C?
Can I work on -  [199](https://github.com/fedora-modularity/libmodulemd/issues/199 )or [198](https://github.com/fedora-modularity/libmodulemd/issues/198)
Or can you assign me something else to  work on

sgallagher:
@athira-selvam If there isn't a specific issue available (and I don't have a spare moment today to try to come up with some new ones), my recommendation would be to look through the existing tests in the `v2` subdirectory, look at some edge-cases that aren't exercised there and submit some new tests.

PR #303, opened 2019-05-22
Fix XMD comparison

sgallagher:
Comparison of the XMDs were accidentally returning the opposite
truth value.

Signed-off-by: Stephen Gallagher <sgallagh@redhat.com>

@OrionStar25 please review. I discovered this while working on https://github.com/fedora-modularity/libmodulemd/issues/188.

PR #304, opened 2019-05-22
Check full stream equality when deduplicating

sgallagher:
Previously, we assumed that any two streams having the same NSVCA
were identical. Now we can actually test for full equality before
dropping one while deduplicating. Note that this will result in a
hard failure if they aren't identical, because this means that the
enabled repositories are providing conflicting data.

Fixes: https://github.com/fedora-modularity/libmodulemd/issues/188

@Conan-Kudo Can I get your opinion on whether it's acceptable to treat this as a hard failure? In the unlikely event that two repositories (or the same repository) are providing two `modulemd` documents with the same NSVCA but different content elsewhere, I don't think there's any fail-safe way to pick one over another, so I think our only option is to return an error.

Also includes a dependent fix being tracked and reviewed in https://github.com/fedora-modularity/libmodulemd/pull/303

### DIFF
--- a/modulemd/v2/modulemd-module-stream-v1.c
+++ b/modulemd/v2/modulemd-module-stream-v1.c
@@ -1075,7 +1075,7 @@ modulemd_module_stream_v1_equals (ModulemdModuleStream *self_1,
   if (v1_self_1->xmd == NULL || v1_self_2->xmd == NULL)
     return FALSE;
 
-  if (g_variant_equal (v1_self_1->xmd, v1_self_2->xmd) != 0)
+  if (!g_variant_equal (v1_self_1->xmd, v1_self_2->xmd))
     return FALSE;
 
   return TRUE;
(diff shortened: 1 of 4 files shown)


Conan-Kudo:
@sgallagher My understanding is that the behavior you want to break is why we're even using YAML in the first place? The free-form merging to allow a module to grow with multiple sources was one of the points brought up against changing the output format to XML.

What changed your mind about disabling that behavior?

sgallagher:
@Conan-Kudo I think you misunderstand. The NSVCA (`module_name:stream_name:version:context:architecture`) must be unique for the YAML. If you want to extend the module you can change any one of those fields to modify. A different stream name, version or context won't conflict.

The problem is if you have two different repos both claiming to define `nodejs:12:20190522120000:deadbeef:x86_64` but not providing the same content. In that case, there's really no way I can think of to decide which one gets to own that NSVCA cleanly.

Conan-Kudo:
@sgallagher It was previously explained to me that it was intentionally possible to do that, and that the intent was that the documents would be merged on the fly to create a module superset (which is incredibly difficult to reason out!) that could be installed with content from both repositories.

sgallagher:
@Conan-Kudo Who said that to you? This has never been true so far as I'm aware.

Conan-Kudo:
@sgallagher You and @contyk two months ago.

sgallagher:
@Conan-Kudo It's possible that you misunderstood when I was talking about merging content into the ModuleIndex. It is supported that different streams might come from different repos, but never that the YAML for a specific NSVCA would come from two different places. That would be madness!

PR #307, opened 2019-05-23
YAML --> .pot file

OrionStar25:
Created a dictionary as:

`dictionary: {module,nsvca : translatable strings}` so that exact streams can be retrieved back after translation.

### DIFF
--- a/.travis/archlinux/Dockerfile.deps.tmpl
+++ b/.travis/archlinux/Dockerfile.deps.tmpl
@@ -12,4 +12,5 @@ RUN pacman -Syu --needed --noconfirm \
 	libyaml \
 	meson \
 	python-gobject \
+	python-babel \
 && pacman -Scc --noconfirm
--- a/ModulemdTranslationHelpers/Utils.py
+++ b/ModulemdTranslationHelpers/Utils.py
@@ -1,0 +1,60 @@
+#!/usr/bin/env python3
+
+from babel.messages import Catalog, pofile
+from babel._compat import StringIO, BytesIO
+from datetime import datetime
+from collections import defaultdict
+
+
+def get_translation_catalog_from_index(index, project_name):
+    # Get all Modulemd.Module object names
+    module_names = index.get_module_names()
+
+    # Create a list containing highest version streams of a module
+    final_streams = list()
+
+    for module_name in module_names:
+        module = index.get_module(module_name)
+        stream_names = module.get_stream_names()
+
+        for stream_name in stream_names:
+            # The first item returned is guaranteed to be the highest version
+            # of that stream in that module.
+            stream = module.search_streams(stream_name, 0)[0]
+            final_streams.append(stream)
+
+    # A dictionary to store:
+    # key: all translatable strings
+    # value: their respective locations
+    translation_dict = defaultdict(list)
+
+    for stream in final_streams:
+        # Process description
+        description = stream.get_description("C")
+        location = ("{}:{};description").format(
+            stream.props.module_name, stream.props.stream_name)
+        translation_dict[description].append(location)
+
+        # Process summary
+        summary = stream.get_summary("C")
+        location = ("{}:{};summary").format(
+            stream.props.module_name, stream.props.stream_name)
+        translation_dict[summary].append(location)
+
+        # Process profile descriptions(sometimes NULL)
+        profile_names = stream.get_profile_names()
+        if(profile_names):
+            for pro_name in profile_names:
+                profile = stream.get_profile(pro_name)
+
+                profile_desc = profile.get_description("C")
+                location = ("{}:{};profile;{}").format(
+                    stream.props.module_name, stream.props.stream_name, pro_name)
+                translation_dict[profile_desc].append(location)
+
+    catalog = Catalog(project=project_name)
+
+    for translatable_string, locations in translation_dict.items():
+        catalog.add(translatable_string, locations=locations)
+
+    return catalog
--- a/modulemd/v2/meson.build
+++ b/modulemd/v2/meson.build
@@ -626,6 +626,45 @@ test ('rpmmap_python2_release', python2,
       env : py_test_release_env,
       args : rpmmap_python_script)
 
+# -- Test Modulemd.TranslationHelpers (Python) -- #
+
+translation_helpers_python_script = files('../../ModulemdTranslationHelpers/Utils.py')
+get_babel = '''
+import babel
+'''
+
+# Python 2
+if (get_option('with_py2_overrides'))
+  ret2 = run_command([python2, '-c', get_babel])
+
+  if ret2.returncode() != 0
+    warning('No Babel on Python 2: skipping TranslationHelper tests')
+  else
+    test ('translation_helpers_python2_debug', python2,
+          env : py_test_env,
+          args : translation_helpers_python_script)
+    test ('translation_helpers_python2_release', python2,
+          env : py_test_release_env,
+          args : translation_helpers_python_script)
+  endif
+endif
+
+# Python 3
+if (get_option('with_py3_overrides'))
+  ret2 = run_command([python3, '-c', get_babel])
+
+  if ret2.returncode() != 0
+    warning('No Babel on Python 3: skipping TranslationHelper tests')
+  else
+    test ('translation_helpers_python3_debug', python3,
+          env : py_test_env,
+          args : translation_helpers_python_script)
+    test ('translation_helpers_python3_release', python3,
+          env : py_test_release_env,
+          args : translation_helpers_python_script)
+  endif
+endif
+
 
 # --- GOBject Introspection -- #
 
--- a/modulemd/v2/tests/ModulemdTests/translation_helpers.py
+++ b/modulemd/v2/tests/ModulemdTests/translation_helpers.py
@@ -1,0 +1,73 @@
+#!/usr/bin/python3
+
+# This file is part of libmodulemd
+# Copyright (C) 2018 Red Hat, Inc.
+#
+# Fedora-License-Identifier: MIT
+# SPDX-2.0-License-Identifier: MIT
+# SPDX-3.0-License-Identifier: MIT
+#
+# This program is free software.
+# For more information on the license, see COPYING.
+# For more information on free software, see
+# <https://www.gnu.org/philosophy/free-sw.en.html>.
+
+import os
+import sys
+
+try:
+    import unittest
+    import gi
+    gi.require_version('Modulemd', '2.0')
+    from gi.repository import Modulemd
+except ImportError:
+    # Return error 77 to skip this test on platforms without the necessary
+    # python modules
+    sys.exit(77)
+
+from base import TestBase
+
+
+class TestTranslationHelpers(TestBase):
+
+    def test_translation_catalog_from_index(self):
+        # Create a Modulemd.ModuleIndex object containing the YAML stream
+        index = Modulemd.ModuleIndex.new()
+        yaml_path = "../test_data/f29.yaml"
+        ret, failures = index.update_from_file(yaml_path, True)
+        self.assertIsNotNone(index)
+
+        # Create a babel catalog for the Modulemd.ModuleIndex object
+        catalog = get_translation_catalog_from_index(index)
+        self.assertTrue(catalog)
+
+        # Check if the catalog is internally consistent
+        _, err = catalog.check()
+        self.assertNone(err)
+
+        # Check if each translatable string is mapped to its correct location
+        for msg in catalog:
+            for location in msg.locations:
+                props = location.split(";")
+
+                module_name = props[0].split(":")[0]
+                module = index.get_module(module_name)
+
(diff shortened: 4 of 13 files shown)


OrionStar25:
@sgallagher , I'd like to have your feedback on this! Also, `ninja test` doesn't format code for these files. Any way for doing that?

sgallagher:
> @sgallagher , I'd like to have your feedback on this! Also, `ninja test` doesn't format code for these files. Any way of doing that?

I'll do a full review in a couple hours.

Regarding the formatting, you need to make sure the files are added to the `autopep8_scripts'` variable in `modulemd/meson.build`

ModulemdTranslationHelpers/Utils.py
```diff
@@ -0,0 +1,120 @@
...
+            final_streams[key] = stream.get_NSVCA()
+        else:
+            nsvca = stream.get_NSVCA()
+            version = int(nsvca.split(':')[2])
+            if version > int(final_streams[key].split(
+                    ":")[2]):  # found a higher version for same stream
+                final_streams[key] = nsvca
+
+summary_dict = {}
+description_dict = {}
```
```suggestion
description_dict = defaultdict(list)
```
and add
```
```
at the top of the file.

A `defaultdict` type allows you to skip the `if` below and always call `default_dict[key].append()`, because if it's not there, it will set the default (in this case, an empty `list`) and then append to that.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -0,0 +1,120 @@
...
+    stream = module.get_stream_by_NSVCA(stream_name, version, context, arch)
+
+    # Process description
+    description = stream.get_description("C")
+    if description not in description_dict.keys():
+        description_dict[description] = [nsvca.rsplit(
+            ":", 3)[0] + ";description"]  # name:stream
+    else:
+        description_dict[description].append(
+            nsvca.rsplit(":", 3)[0] + ";description")
```
sgallagher: You're getting too fancy with the `rsplit()` and concatenation here. Just do

```suggestion
        description_dict[description].append("{}:{};description" % (stream.props.module_name, stream.props.stream_name))
```

ModulemdTranslationHelpers/Utils.py
```diff
@@ -0,0 +1,120 @@
...
+
+    # Process summary
+    summary = stream.get_summary("C")
+    if summary not in summary_dict.keys():
+        summary_dict[summary] = [
+            nsvca.rsplit(
+                ":",
+                3)[0] +
+            ";summary"]  # name:stream
```
sgallagher: Make the same change as the description above.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -0,0 +1,120 @@
...
+        else:
+            nsvca = stream.get_NSVCA()
+            version = int(nsvca.split(':')[2])
+            if version > int(final_streams[key].split(
+                    ":")[2]):  # found a higher version for same stream
+                final_streams[key] = nsvca
+
+summary_dict = {}
+description_dict = {}
+profile_desc_default_dict = {}
```
sgallagher: These two dicts are unnecessary. Drop them.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -0,0 +1,120 @@
...
+        profile = stream.get_profile(names)
+        profile_desc = profile.get_description("C")
+        if profile_desc:
+            if profile_desc[0] not in profile_desc_default_dict.keys():
+                profile_desc_default_dict[
+                    profile_desc[0]] = [
+                    nsvca.rsplit(
+                        ":",
+                        3)[0] +
+                    ";profile;default"]
```
sgallagher: `default` and `minimal` are not reserved words, they are just examples. The name of a profile can be any string. You need to record that name here. Also, simplify this like I suggested for summary and description.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -0,0 +1,120 @@
...
+        if key not in final_streams.keys():
+            final_streams[key] = stream.get_NSVCA()
+        else:
+            nsvca = stream.get_NSVCA()
+            version = int(nsvca.split(':')[2])
+            if version > int(final_streams[key].split(
+                    ":")[2]):  # found a higher version for same stream
+                final_streams[key] = nsvca
+
+summary_dict = {}
```
sgallagher: These should not be separate dictionaries. By the end of this, there must be exactly one dictionary whose keys are the untranslated strings and whose values are a list of places where that string was found.

One example problem: you may have a module and stream that has the same string for both the summary and description fields. What that should look like is:
```
strings_dict["extracted string"] = [ "moduleA:streamA:summary", "moduleA:streamA:description" ]
```

ModulemdTranslationHelpers/Utils.py
```diff
@@ -0,0 +1,120 @@
...
+description_dict = {}
+profile_desc_default_dict = {}
+profile_desc_minimal_dict = {}
```
sgallagher: Also, same as with summary and description: the profile locations should just be added to the common list.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -0,0 +1,120 @@
...
+                        3)[0] +
+                    ";profile;minimal"]
+            else:
+                profile_desc_minimal_dict[profile_desc[1]].append(
+                    nsvca.rsplit(":", 3)[0] + ";profile;minimal")
+
+
+catalog = Catalog(project="fedora-modularity-translations")
+
+for key, value in summary_dict.items():
```
sgallagher: The net result here is that you may be adding duplicates to the catalog, which will break gettext. There should be a single dictionary that gets converted into the catalog.

ModulemdTranslationHelpers/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+#
+# Fedora-License-Identifier: MIT
+# SPDX-2.0-License-Identifier: MIT
+# SPDX-3.0-License-Identifier: MIT
+#
+# This program is free software.
+# For more information on the license, see COPYING.
+# For more information on free software, see
+# <https://www.gnu.org/philosophy/free-sw.en.html>.
+
```
sgallagher: This file is irrelevant to this task. Drop it from this commit.

ModulemdTranslationHelpers/cli.py
```diff
@@ -0,0 +1,139 @@
...
+# # SPDX-2.0-License-Identifier: MIT
+# # SPDX-3.0-License-Identifier: MIT
+# #
+# # This program is free software.
+# # For more information on the license, see COPYING.
+# # For more information on free software, see
+# # <https://www.gnu.org/philosophy/free-sw.en.html>.
+
+# from __future__ import print_function
+
```
sgallagher: This file is irrelevant to this task. Drop it from the commit.

ModulemdTranslationHelpers/setup.py
```diff
@@ -0,0 +1,49 @@
...
+# #
+# # Fedora-License-Identifier: MIT
+# # SPDX-2.0-License-Identifier: MIT
+# # SPDX-3.0-License-Identifier: MIT
+# #
+# # This program is free software.
+# # For more information on the license, see COPYING.
+# # For more information on free software, see
+# # <https://www.gnu.org/philosophy/free-sw.en.html>.
+
```
sgallagher: This file is irrelevant to this task. Drop it from the commit.

OrionStar25:
@sgallagher I tried to check if the function is working correctly. But it says that `module` has no method `get_stream_names()` . Even with `get_stream_names_as_strv()` . Can you point me to the place where all the python functions get rendered?

sgallagher:
> @sgallagher I tried to check if the function is working correctly. But it says that `module` has no method `get_stream_names()` . Even with `get_stream_names_as_strv()` . Can you point me to the place where all the python functions get rendered?

As discussed on IRC, you are using functions that aren't yet in the version that is installed on your operating system, so you need to tell Python where to find the copy you built locally.

You need to have `GI_TYPELIB_PATH=/path/to/builddir/modulemd/v2` and `LD_LIBRARY_PATH=/path/to/builddir/modulemd/v2` set in the environment where you are running Python (we have this done automatically for the tests run by `meson test` at https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/meson.build#L106)

OrionStar25:
@sgallagher It wasn't allowing adding `search_streams` into `rename-bindings` without renaming the function. I renamed it just to check if everything was working.

sgallagher:
> @sgallagher It wasn't allowing adding `search_streams` into `rename-bindings` without renaming the function. I renamed it just to check if everything was working.

Why did you add a rename? That's completely unnecessary. It looks like your earlier problem was just that you were trying to call `.search_stream()` when the function is called `.search_streams()`.

OrionStar25:
> It's starting to come together.
> 
> Please create a test for this function (you will need to create a new test, so see the existing examples in [`meson.build`](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/meson.build#L173) to create one (and add it to the `autopep8_scripts` list).

Writing tests for it now.

OrionStar25:
@sgallagher the closest thing I could find to validate a catalog was `.check()` . Hope this works!

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -0,0 +1,48 @@
...
+    def test_translation_catalog_from_index(self):
+        # Create a Modulemd.ModuleIndex object containing the YAML stream
+        index = Modulemd.ModuleIndex.new()
+        yaml_path = "../test_data/f29.yaml"
+        ret, failures = index.update_from_file(yaml_path, True)
+        self.assertIsNotNone(index)
+
+        catalog = get_translation_catalog_from_index(index)
+        self.assertTrue(catalog)
+        self.assertEquals(len(catalog), 34)
```
sgallagher: Explain *why* the number 34 is correct in a comment, please. "Magic" numbers like this always need to be explained, or else if this test starts failing, the next person won't know what to do about it.

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -0,0 +1,48 @@
...
+        _, err = catalog.check()
```
sgallagher: This validates that the catalog is internally consistent, but it doesn't verify that it contains the set of strings we expect to have here.

What I'd like you to do here is iterate through the catalog with 
```python3
for msg in catalog:
```
I'd like you to get the `msg.string` and iterate through the list of `msg.locations` and confirm that the string matches at that location. This, plus the catalog count matching the expected number of them is probably enough to validate that it works.

OrionStar25:
@sgallagher What's happening rn is that `arch linux` and `centos` have babel for `python2` only. So, for the rest of the OS, all tests are running and passing (locally). For these 2, it shows an error message of `Cannot import babel 3`, which is fine, but then it stops there. It doesn't run any tests. 

If I remove the part of the code which is trying to import `python3-babel`, all tests run for `centos` and `archlinux` and they pass.

modulemd/v2/meson.build
```diff
@@ -626,6 +626,45 @@ test ('rpmmap_python2_release', python2,
...
+get_babel = '''
+import babel
+'''
+
+# Python 2
+if (get_option('with_py2_overrides'))
+  ret2 = run_command([python2, '-c', get_babel])
+
+  if ret2.returncode() != 0
+    error('Failed to import Python 2 babel')
```
sgallagher: Here's the problem. The `error()` directive in meson means "Print this error and then quit with a non-zero exit code". You need to store a variable and just use `message()` to print appropriate text instead. Something like:
```suggestion
    warning('No Babel on Python 2: skipping TranslationHelper tests')
```

modulemd/v2/meson.build
```diff
@@ -626,6 +626,45 @@ test ('rpmmap_python2_release', python2,
...
+          args : translation_helpers_python_script)
+  endif
+endif
+
+# Python 3
+if (get_option('with_py3_overrides'))
+  ret2 = run_command([python3, '-c', get_babel])
+
+  if ret2.returncode() != 0
+    error('Failed to import Python 3 babel')
```
sgallagher:
```suggestion
    warning('No Babel on Python 3: skipping TranslationHelper tests')
```

.travis/archlinux/Dockerfile.deps.tmpl
```diff
@@ -12,4 +12,5 @@ RUN pacman -Syu --needed --noconfirm \
 	libyaml \
 	meson \
 	python-gobject \
+	python2-babel \
```
sgallagher: You want the python3 version here:

```suggestion
	python-babel \
```

OrionStar25:
Just checking if tests are passing, then shall merge them into 2 commits. One for the renaming of `get_stream_names` method and one for catalog conversion.

Issue #308, opened 2019-05-28
module.search_streams() should be (transfer container)


sgallagher:
There's a bug in the API for `Modulemd.Module.search_stream()` and it has `(transfer full)` instead of `(transfer container)`, which will lead to memory-corruption issues (since garbage-collection will attempt to free the contents of the array rather than just the array container).

CC @smooge @puiterwijk as an FYI. I will get a patch out immediately.

PR #310, opened 2019-05-28
Fix transfer type for Module.search_streams()

sgallagher:
Technically this is an API-breaking change, but no one is using it
yet and it was always expected to be managed this way.

Fixes: https://github.com/fedora-modularity/libmodulemd/issues/308

### DIFF
--- a/modulemd/common/tests/test-dirty.py
+++ b/modulemd/common/tests/test-dirty.py
@@ -15,6 +15,7 @@
 import os
 import sys
 import git
+import subprocess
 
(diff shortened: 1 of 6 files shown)


sgallagher:
Note that the test does exercise this; if you attempt to run that test against an unmodified 2.5.0, you will see an error `modulemd_module_stream_get_stream_name: assertion 'MODULEMD_IS_MODULE_STREAM (self)' failed`

This is because the memory management frees the elements of the array unexpectedly.

PR #313, opened 2019-05-30
Fixes #309: Add get_stream_names for modules

### LINKED ISSUE #309: Add Modulemd.Module.get_stream_names()
There should be a function to get a list of all stream names for this Module. Among other things, this can be used in conjunction with `Module.search_streams()`.

OrionStar25:
Fixes #309

### DIFF
--- a/modulemd/v2/include/modulemd-2.0/modulemd-module.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-module.h
@@ -34,6 +34,18 @@ G_DECLARE_FINAL_TYPE (
 
 
 /**
+ * modulemd_module_get_stream_names_as_strv
+ * @self: This #ModulemdModule
+ *
+ * Returns: (transfer full): An ordered list of stream names in this module.
+ *
+ * Since: 2.6
+ */
+GStrv
+modulemd_module_get_stream_names_as_strv (ModulemdModule *self);
+
+
+/**
  * modulemd_module_copy:
  * @self: (in): This #ModulemdModule object
  *
--- a/modulemd/v2/modulemd-module.c
+++ b/modulemd/v2/modulemd-module.c
@@ -389,6 +389,28 @@ modulemd_module_add_stream (ModulemdModule *self,
 }
 
 
+GStrv
+modulemd_module_get_stream_names_as_strv (ModulemdModule *self)
+{
+  g_return_val_if_fail (MODULEMD_IS_MODULE (self), NULL);
+
+  if (!self->streams)
+    return NULL;
+
+  g_autoptr (GHashTable) stream_names =
+    g_hash_table_new (g_str_hash, g_str_equal);
+
+  for (guint i = 0; i < self->streams->len; i++)
+    {
+      g_hash_table_add (stream_names,
+                        (void *)modulemd_module_stream_get_stream_name (
+                          g_ptr_array_index (self->streams, i)));
+    }
+
+  return modulemd_ordered_str_keys_as_strv (stream_names);
+}
+
+
 GPtrArray *
 modulemd_module_get_all_streams (ModulemdModule *self)
 {
--- a/modulemd/v2/tests/test-modulemd-module.c
+++ b/modulemd/v2/tests/test-modulemd-module.c
@@ -331,6 +331,84 @@ module_test_streams (ModuleFixture *fixture, gconstpointer user_data)
 
 
 static void
+module_test_get_stream_names (ModuleFixture *fixture, gconstpointer user_data)
+{
+  g_autoptr (ModulemdModule) m = NULL;
+  g_autoptr (ModulemdModuleStream) stream = NULL;
+  g_auto (GStrv) list = NULL;
+
+  /*Test module with no streams*/
+  m = modulemd_module_new ("testmodule");
+  list = modulemd_module_get_stream_names_as_strv (m);
+  g_assert_nonnull (list);
+  g_assert_cmpint (g_strv_length (list), ==, 0);
+  g_clear_pointer (&list, g_strfreev);
+  g_clear_object (&m);
+
+  /*Test module with all same stream names*/
+  m = modulemd_module_new ("testmodule");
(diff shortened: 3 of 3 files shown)


modulemd/v2/include/modulemd-2.0/modulemd-module.h
```diff
@@ -33,6 +33,18 @@ G_DECLARE_FINAL_TYPE (
   ModulemdModule, modulemd_module, MODULEMD, MODULE, GObject)
 
 
+/**
+ * modulemd_module_index_get_stream_names_as_strv: (rename-to modulemd_module_index_get_stream_names)
+ * @self: This #ModulemdModule
+ *
+ * Returns: (transfer full): An ordered list of string keys in this module.
```
sgallagher:
```suggestion
 * Returns: (transfer full): An ordered list of stream names in this module.
```

sgallagher: I'm guessing you copied this from `modulemd_module_index_get_module_names_as_strv()`, but that was a bad description there too. I'll fix that later.

modulemd/v2/include/modulemd-2.0/modulemd-module.h
```diff
@@ -33,6 +33,18 @@ G_DECLARE_FINAL_TYPE (
   ModulemdModule, modulemd_module, MODULEMD, MODULE, GObject)
 
 
+/**
+ * modulemd_module_index_get_stream_names_as_strv: (rename-to modulemd_module_index_get_stream_names)
```
sgallagher: You use the `(rename-to)` directive here, but in order to do that you also need to update
https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/include/private/gi-binding-renames.h

modulemd/v2/modulemd-module.c
```diff
@@ -389,6 +389,15 @@ modulemd_module_add_stream (ModulemdModule *self,
 }
 
 
+GStrv
+modulemd_module_index_get_stream_names_as_strv (ModulemdModule *self)
+{
+  g_return_val_if_fail (MODULEMD_IS_MODULE (self), NULL);
+
+  return modulemd_ordered_str_keys_as_strv ((GHashTable *)self->streams);
```
sgallagher: `self->streams` is a `GPtrArray`, not a `GHashTable`. You can't get at them this way.

The simplest approach would be to create a new `GHashTable` (call it `stream_names`)in this function and iterate through `self->streams`, calling `g_hash_table_add(stream_names, stream_name_from_index)`

Then return `modulemd_ordered_str_keys_as_strv(stream_names)`.

OrionStar25:
@sgallagher 
Screenshot from 2019-06-01 03-44-33

I'm getting this error and I tried to trace it with `gdb`. I don't understand this because I'm not even using the function and the function definition also doesn't have anything related.

sgallagher:
> @sgallagher
> Screenshot from 2019-06-01 03-44-33
> 
> I'm getting this error and I tried to trace it with `gdb`. I don't understand this because I'm not even using the function and the function definition also doesn't have anything related.

That function is called implicitly by `g_hash_table_add()`. Take a look at your changes to modulemd-module.c carefully. Did you properly initialize the hash table before adding to it?

modulemd/v2/include/modulemd-2.0/modulemd-module.h
```diff
@@ -33,6 +33,18 @@ G_DECLARE_FINAL_TYPE (
   ModulemdModule, modulemd_module, MODULEMD, MODULE, GObject)
 
 
+/**
+ * modulemd_module_get_stream_names
+ * @self: This #ModulemdModule
+ *
+ * Returns: (transfer full): An ordered list of stream names in this module.
```
sgallagher: Please don't use a GPtrArray here. When returning an array of strings, the libmodulemd public API prefers `GStrv` format (which is a NULL-terminated array of `gchar *` pointers).

You can use [`modulemd-ordered-str-keys-as-strv()`](https://fedora-modularity.github.io/libmodulemd/latest/modulemd-2.0-Modulemd-Utility-Functions.html#modulemd-ordered-str-keys-as-strv) to get that from a GHashTable.

Also, for the record, you wouldn't want to do `(transfer full)` if you were returning a `GPtrArray` here, because the `GPtrArray` type owns its own contents, so you'd be causing it to do a double-free. Also also, if you return a `GPtrArray`, you need to use the `(element-type ...)` directive in the gtkdoc to indicate what type the elements of the array is carrying. Without this, the python binding can't figure it out.

OrionStar25:
@sgallagher  changed the return type to `GStrv` but still can't figure out how the errors are being generated. From the docs

> Starting from GLib 2.40, `g_hash_table_add()` returns a boolean value to indicate whether the newly added value was already in the hash table or not.

modulemd/v2/modulemd-module.c
```diff
@@ -389,6 +389,28 @@ modulemd_module_add_stream (ModulemdModule *self,
...
+  if (!self->streams)
+    return NULL;
+
+  g_autoptr (GHashTable) stream_names =
+    g_hash_table_new (g_str_hash, g_str_equal);
+
+  for (guint i = 0; i < self->streams->len; i++)
+    {
+      g_hash_table_add (stream_names,
+                        (void *)modulemd_module_stream_get_stream_name (
```
sgallagher: You don't actually need to cast to `(void *)` here, so don't.

OrionStar25: It was giving an error saying expected void * but given gchar **

sgallagher: If it’s getting `char **` there, something is wrong. `get_stream_name()` returns `const gchar *`

OrionStar25: My bad, I edited the warning message. I'm sleepy.

sgallagher: Ah right, it doesn’t like the implicit cast discarding the `const`. That makes sense. Okay, approved.

PR #319, opened 2019-06-16
From catalog to modified Modulemd.ModuleIndex

OrionStar25:
- [x] The tests are passing even when they shouldn't. I tried to assert obvious false values too. Its almost as if the function isn't being tested. 
- [x] The format of locations in the example `.po` file is different than how we set it in the previous function.
`module:stream;string_type` vs `module:stream:string_type` . I've written tests according to the example file.


modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
         # There are 34 unique summaries and descriptions in f29.yaml
         self.assertEquals(len(catalog), 34)
 
+    def test_translations_from_catalog(self):
+        translation_files = ["../test_data/nl.po"]
```
sgallagher: This will only work if the current working directory is the same as the file location. Look at https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/tests/ModulemdTests/modulestream.py#L1111 for an example of how to specify the test data location relative to the `$MESON_SOURCE_ROOT`.

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+
+        catalogs = dict()
```
sgallagher: As discussed on IRC, don't make this a dictionary. Just a `list()` will be sufficient.

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+        for f in translation_files:
+            with open(f, 'r') as infile:
+                catalog = pofile.read_po(infile)
+                self.assertTrue(catalog)
```
sgallagher:
```suggestion
                self.assertIsNotNone(catalog)
```

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+
+                # Check if the catalog is internally consistent
+                _, err = catalog.check()
+                self.assertNone(err)
+                catalogs[catalog.locale_identifier] = catalog
```
sgallagher: Just append to a list here.

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+
+        # Create a Modulemd.ModuleIndex object containing the YAML stream
+        index = Modulemd.ModuleIndex.new()
+        yaml_path = "../test_data/f29.yaml"
```
sgallagher: Same as above regarding relative paths.

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+        ret, failures = index.update_from_file(yaml_path, True)
+        self.assertIsNotNone(index)
```
sgallagher: This should be after the `.new()`. Here, it won't be `None`, but you should check that `ret` is True and `len(failures) == 0`.

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+
+        err = get_modulemd_translations_from_catalog(catalogs, index)
+        self.assertNone(err)
```
sgallagher: This should not return an error, it should `raise` an error if one occurs. Do not return a value here since it will be updating `index` in place.

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+
+        for locale, catalog in catalogs.items():
+            for msg in catalog:
+                locations = [x[0] for x in msg.locations]
```
sgallagher: This is a waste; we don't need to copy the list.

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+                for location in locations:
```
sgallagher:
```suggestion
                for location, lineno in msg.locations:
```

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+                    props = location.split(";")
```
sgallagher: Please make this a utility function that returns named attributes instead of referencing it with array index values.

So something like `(module_name, stream_name, location_type, profile_name) = split_location(location)`

sgallagher: `profile_name` should return `None` for `location_type` of "summary" or "description".

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+
+                    module_name = props[0]
+                    module = index.get_module(module_name)
+                    self.assertIsNotNone(module)
+
+                    stream_name = props[1]
+                    stream = module.search_streams(stream_name, 0)[0]
```
sgallagher: Instead of taking only the first entry in the list, iterate through all matching streams. That way, if there's more than one, we can verify that the translation has taken effect for all of them.

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+                    self.assertIsNotNone(stream)
+
+                    string_type = props[2]
+                    if string_type == "summary":
+                        assert stream.get_summary(locale) == msg.string
```
sgallagher:
```suggestion
                        self.assertEqual(stream.get_summary(), msg.string)
```
The `unittest.assertEqual()` function will show the diff if they don't match. It makes it easier to find the mistake.

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+                    elif string_type == "description":
+                        assert stream.get_description(locale) == msg.string
```
sgallagher: same here

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -68,6 +68,52 @@ def test_translation_catalog_from_index(self):
...
+                    else:
+                        profile = stream.get_profile(props[3])
+                        assert profile.get_description(locale) == msg.string
```
sgallagher: same here

OrionStar25:
I've added the code and updated the design in the doc. Its failing for locations containing profile names. Shall work on that tomorrow. Does the data structure seem fine now though? 

There's also the conflicting issue of location-naming. But I think I can solve that using tuples.  (TODO)

ModulemdTranslationHelpers/Utils.py
```diff
@@ -58,3 +63,94 @@ def get_translation_catalog_from_index(index, project_name):
...
+        translation = props[2]
+    else:
+        profile_name = None
+        translation = props[1]
+
+    return string_type, profile_name, translation
+
+
+def split_location(location):
+    props = location.split(";")
```
```suggestion
    props = location.split(";", maxsplit=4)
```

ModulemdTranslationHelpers/Utils.py
```diff
@@ -58,3 +63,94 @@ def get_translation_catalog_from_index(index, project_name):
...
+
+    data = defaultdict(defaultdict)
+
+    for catalog in catalogs:
+        for msg in catalog:
+            for location, _ in msg.locations:
+                (module_name, stream_name, string_type,
+                 profile_name) = split_location(location)
+
+                key_1 = ("{}:{}").format(module_name, stream_name)
```
sgallagher: Don't make the key into a string. Python dictionaries can handle any basic type here, so it'll be easiest to use a tuple of `(module_name, stream_name)` as the key instead of `"module_name:stream_name"` which you'll have to split again later.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -58,3 +63,94 @@ def get_translation_catalog_from_index(index, project_name):
...
+                if profile_name is None:
+                    value_2 = ("{}:{}").format(string_type, msg.string)
+                else:
+                    value_2 = ("{}:{}:{}").format(
+                        string_type, profile_name, msg.string)
+
+                # Adding into data
+                data[key_1].setdefault(
```
sgallagher: So, since we can't use a defaultdict, you'll need to just do things the slightly-ugly-but-pythonic way of:

```python3
try:
    entry = data[(module_name, stream_name)]
except KeyError:
    entry = Modulemd.TranslationEntry.new(catalog.locale_identifier)

# set the appropriate properties for this locale

# save it back
data[(module_name, stream_name)] = entry
```

ModulemdTranslationHelpers/Utils.py
```diff
@@ -58,3 +63,94 @@ def get_translation_catalog_from_index(index, project_name):
         catalog.add(translatable_string, locations=locations)
 
     return catalog
+
+
+def get_modulemd_translations_from_catalog(catalogs, index):
+    # Nested dictionary `data` contains information from catalog like:
+    # Key_1: <module>:<stream>
+    # Value_1: Key_2: <locale>
+    #          Value_2: list of (<string_type>:<msg.string>)
```
sgallagher: Instead of making a list here, we already know that the contents of this are going to match exactly to a `TranslationEntry`, so let's just create one of those instead. We won't be able to use a `defaultdict` though, because we need to construct the `TranslationEntry` with the correct locale (since this is not changeable after creation).

ModulemdTranslationHelpers/Utils.py
```diff
@@ -58,3 +63,94 @@ def get_translation_catalog_from_index(index, project_name):
...
+                    value_2 = ("{}:{}").format(string_type, msg.string)
+                else:
+                    value_2 = ("{}:{}:{}").format(
+                        string_type, profile_name, msg.string)
+
+                # Adding into data
+                data[key_1].setdefault(
+                    catalog.locale_identifier, []).append(value_2)
+
+    for mod_stream, translations in data.items():
```
sgallagher: This will become:
```suggestion
    for (module_name, stream_name), translations in data.items():
```

ModulemdTranslationHelpers/Utils.py
```diff
@@ -58,3 +63,94 @@ def get_translation_catalog_from_index(index, project_name):
...
+        (module_name, stream_name) = mod_stream.split(":")
+        now = datetime.utcnow()
```
sgallagher: Don't do this multiple times through the loop. Asking for the current time is expensive. Get the time exactly once at the start of processing and use the same value everywhere.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -58,3 +63,94 @@ def get_translation_catalog_from_index(index, project_name):
...
+            now.date().month,
+            now.date().day,
+            now.time().hour,
+            now.time().minute,
+            now.time().second
+        ))
+        mmd_translation = Modulemd.Translation.new(
+            1, module_name, stream_name, modified)
+
+        for locale, strings in translations.items():
```
sgallagher: If we just create and save the `TranslationEntry` earlier, this loop just becomes:
```suggestion
        for locale, entry in translations.items():
            mmd_translation.set_translation_entry(entry)
```

ModulemdTranslationHelpers/Utils.py
```diff
@@ -58,3 +63,94 @@ def get_translation_catalog_from_index(index, project_name):
...
+                if string_type == "summary":
+                    entry.set_summary(translation)
+                elif string_type == "description":
+                    entry.set_description(translation)
+                else:
+                    entry.set_profile_description(profile_name, translation)
+
+            mmd_translation.set_translation_entry(entry)
+
+        ret = index.add_translation(mmd_translation)
```
sgallagher: In python, `ModuleIndex.add_translation()` will throw an exception if the translation cannot be added (the boolean is intended for use with C). So if you want to trap a failure here, wrap it in a try/except block.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -58,3 +63,94 @@ def get_translation_catalog_from_index(index, project_name):
...
+        if not ret:
+            print("Error!")
```
sgallagher: This is not sufficient. If you're going to report an error here, logging "error" and then continuing is probably incorrect. Best is to catch the exception and handle it if appropriate or else fail.

In this particular case, probably we should just log that the translation for this locale for this module:stream could not be added. It need not be fatal.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -58,3 +63,94 @@ def get_translation_catalog_from_index(index, project_name):
...
+
+    return None
+
+
+def split_translation_string(string):
```
sgallagher: This entire function should go away, since we should just be operating on the TranslationEntry itself instead of duplicating effort by creating a list of joined strings only to have to split them apart again moments later.

OrionStar25:
@sgallagher The `profile descriptions` aren't getting added for some reason. If I remove assertions for `profile descriptions`, everything else is passing. I even tried to retrieve the string after adding into `index` and got a translated version. But somehow the assertion is failing.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -58,3 +63,83 @@ def get_translation_catalog_from_index(index, project_name):
...
+                mmd_translation = Modulemd.Translation.new(
+                    1, module_name, stream_name, modified)
+
+            mmd_translation.set_translation_entry(entry)
+            translations[(module_name, stream_name)] = mmd_translation
+
+    for (module_name, stream_name), mmd_translation in translations.items():
+        try:
+            ret = index.add_translation(mmd_translation)
+        except ValueError:
```
sgallagher: This won't be a `ValueError`, it will be a `GLib.Error`.

sgallagher:
So, something I just realized while fixing #321 

There's a case we aren't handling properly that causes some breakage. The catalog retrieved from the translators may include multiple entries for the same location. In particular, the `nl.po` that we're using here contains two different entries for the `user` profile of the `dwm` module (and for several different streams). This is because there was a typo in the original English version and the translation server contains both the original incorrect string and the corrected one, both referencing the same location.

This is a perfectly reasonable use-case to handle, so we're going to need to figure out how to do it. Let's schedule a meeting tomorrow or Friday and hash it out.

OrionStar25:
> Good news: there's nothing wrong with your test that I can see (other than the incorrect exception being caught in one place).
> 
> Better news: you've caught a bug in `Profile.get_description()`. If you look at https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/modulemd-profile.c#L158 I think you'll spot the problem...

`  /* TODO: retrieve translated strings */` DEAD :rofl:

OrionStar25:
Resolving the travis issue now.

OrionStar25:
@sgallagher All Fedora docker tests are failing because it can't find `ModulemdTranslationHelpers`. But since we have provided the entire build directory into the docker, shouldn't the `PYTHONPATH` be set inside it? Also, then how is this working for the other dockers?

sgallagher:
@OrionStar25 Look carefully at the output. It's not failing when you're running the tests as part of the build process. It's failing when you run the tests that are designed to test *installed* code (after `dnf install libmodulemd`). We aren't installing the ModulemdTranslationHelpers onto the target system. We don't have any mechanism for doing this in the build-system at all yet.

So you should modify the meson.build script to skip the ModulemdTranslationHelpers tests when running with `meson -Dtest_installed_lib=True`.

sgallagher:
The reason it's passing on the non-Fedora CI is because those platforms don't do an "installed packages" test. Just the build-time tests.

OrionStar25:
@sgallagher Green after so long :sob: :green_heart:

ModulemdTranslationHelpers/Utils.py
```diff
@@ -58,3 +63,84 @@ def get_translation_catalog_from_index(index, project_name):
...
+
+    now = datetime.utcnow()
+    modified = int("%04d%02d%02d%02d%02d%02d" % (
+        now.date().year,
+        now.date().month,
+        now.date().day,
+        now.time().hour,
+        now.time().minute,
+        now.time().second
+    ))
```
mmathesius: This assignment can be simplified to `modified = int(now.strftime("%Y%m%d%H%M%S"))`

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -33,40 +36,102 @@ class TestTranslationHelpers(TestBase):
...
+                self.assertIsNotNone(stream)
 
-                string_type = props[1]
                 if string_type == "summary":
-                    assert stream.get_summary("C") == msg.id
+                    self.assertEqual(stream.get_summary("C"), msg.id)
                 elif string_type == "description":
-                    assert stream.get_description("C") == msg.id
+                    self.assertEqual(stream.get_description("C"), msg.id)
                 else:
```
mmathesius: At this point, I'd suggest asserting that `string_type` is `"profile"`, and that `profile_name` has a value.

mmathesius: Skip the part about checking that `profile_name` has a value. What I intended to suggest was checking that `profile` has a value below, right after calling `get_profile()`.

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -33,40 +36,102 @@ class TestTranslationHelpers(TestBase):
...
+                        if string_type == "summary":
+                            self.assertEqual(
+                                stream.get_summary(
+                                    str(catalog.locale)), msg_string)
+                        elif string_type == "description":
+                            self.assertEqual(
+                                stream.get_description(
+                                    str(catalog.locale)), msg_string)
+                        else:
+                            profile = stream.get_profile(profile_name)
```
mmathesius: Same as suggested above about verifying `string_type` and `profile`.

mmathesius:
A few suggestions from me, but looks good!

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -33,40 +36,104 @@ class TestTranslationHelpers(TestBase):
...
+                self.assertIsNotNone(stream)
 
-                string_type = props[1]
                 if string_type == "summary":
-                    assert stream.get_summary("C") == msg.id
+                    self.assertEqual(stream.get_summary("C"), msg.id)
                 elif string_type == "description":
-                    assert stream.get_description("C") == msg.id
+                    self.assertEqual(stream.get_description("C"), msg.id)
                 else:
```
mmathesius: I still suggest adding `self.assertEqual(string_type, "profile")` here...

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -33,40 +36,104 @@ class TestTranslationHelpers(TestBase):
...
-                    profile = stream.get_profile(stream_type.split(";")[1])
-                    assert profile.get_description("C") == msg.id
+                    profile = stream.get_profile(profile_name)
+                    self.assertTrue(profile)
```
mmathesius:
```suggestion
                    self.assertIsNotNone(profile)
```

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -33,40 +36,104 @@ class TestTranslationHelpers(TestBase):
...
+                    for stream in streams:
+                        if string_type == "summary":
+                            self.assertEqual(
+                                stream.get_summary(
+                                    str(catalog.locale)), msg_string)
+                        elif string_type == "description":
+                            self.assertEqual(
+                                stream.get_description(
+                                    str(catalog.locale)), msg_string)
+                        else:
```
mmathesius: Same as above: `self.assertEqual(string_type, "profile")`

modulemd/v2/tests/ModulemdTests/translation_helpers.py
```diff
@@ -33,40 +36,104 @@ class TestTranslationHelpers(TestBase):
...
+                            profile = stream.get_profile(profile_name)
+                            self.assertTrue(profile)
```
mmathesius:
```suggestion
                            self.assertIsNotNone(profile)
```

OrionStar25:
@sgallagher  @mmathesius Please review.

Issue #321, opened 2019-06-26
Profile.get_description() is not handling translations


sgallagher:
https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/modulemd-profile.c#L158

Discovered during the implementation of #319 

We need to make sure that the profile is properly associated with the correct Translation object in the index.

PR #322, opened 2019-06-26
Return translated profile descriptions

sgallagher:
This causes profiles to be associated with their owning ModuleStream
when they are added to it so they can look up the appropriate
translations for their description.

Fixes: https://github.com/fedora-modularity/libmodulemd/issues/321

### DIFF
--- a/modulemd/v2/modulemd-profile.c
+++ b/modulemd/v2/modulemd-profile.c
@@ -14,8 +14,10 @@
 #include <glib.h>
 #include <yaml.h>
 
+#include "modulemd-module-stream.h"
 #include "modulemd-profile.h"
 #include "private/glib-extensions.h"
+#include "private/modulemd-module-stream-private.h"
 #include "private/modulemd-profile-private.h"
 #include "private/modulemd-util.h"
 #include "private/modulemd-yaml.h"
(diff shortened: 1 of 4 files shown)


sgallagher:
This change doesn't add tests because I tested it with the tests being added in #319

modulemd/v2/modulemd-profile.c
```diff
@@ -155,7 +160,17 @@ modulemd_profile_get_description (ModulemdProfile *self, const gchar *locale)
...
+  /* Retrieve translated strings, if available */
+  if (self->owner)
+    {
+      ModulemdTranslationEntry *entry =
+        modulemd_module_stream_get_translation_entry (self->owner, locale);
+
+      if (entry != NULL && modulemd_translation_entry_get_profile_description (
+                             entry, self->name) != NULL)
+        return modulemd_translation_entry_get_profile_description (entry,
+                                                                   self->name);
```
mmathesius: (Optional) The above would be a little easier to grok if the flow was along the lines of
```
if (entry != NULL)
  {
    gchar *translation = modulemd_translation_entry_get_profile_description (
                             entry, self->name);
    if (translation != NULL)
      return translation;
  }
```

sgallagher:
@mmathesius I made the two changes you suggested, but the CI also indicated that I had introduced a memory leak. It took me a while to figure out, but it was because I was having `Profile.set_owner()` take a reference on the `ModuleStream` that encapsulated it. The result was that I was introducing a refcounting loop and the `ModuleStream` would never be freed. I switched it to just assigning a pointer rather than taking a reference and the memory leak went away.

Please see the fixup commit and confirm the results and I'll squash and merge then.

PR #326, opened 2019-07-03
Add ModuleIndexMerger.resolve_ext()

sgallagher:
This will allow us to handle default streams strictly when merging. In most end-user
cases, we want default stream conflicts to simply drop the default stream (this is
fail-safe). But when we are creating a repo with defaults, we want to be stricter and
ensure that we aren't producing a repo with inconsistent data.

cc @lubomir

### DIFF
--- a/modulemd/v2/include/modulemd-2.0/modulemd-module-index-merger.h
+++ b/modulemd/v2/include/modulemd-2.0/modulemd-module-index-merger.h
@@ -161,6 +161,10 @@ modulemd_module_index_merger_associate_index (ModulemdModuleIndexMerger *self,
  * #ModulemdModuleIndexMerger is undefined. The only valid action on it after
  * that point is g_object_unref().
  *
+ * This function is equivalent to calling
+ * modulemd_module_index_merger_resolve_ext() with
+ * `strict_default_streams=FALSE`.
+ *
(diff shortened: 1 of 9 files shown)


lubomir:
Looks good to me. The only downside over what currently happens in Pungi is that only the first such problem is reported. However I think it's highly unlikely such error would happen multiple times, so this is not a big deal. I would use it.

modulemd/v2/include/modulemd-2.0/modulemd-module-index-merger.h
```diff
@@ -161,6 +161,10 @@ modulemd_module_index_merger_associate_index (ModulemdModuleIndexMerger *self,
  * #ModulemdModuleIndexMerger is undefined. The only valid action on it after
  * that point is g_object_unref().
  *
+ * This function is equivalent to calling
+ * modulemd_module_index_merger_resolve_ext() with
+ * `strict_default_streams=False`.
```
mmathesius: Here and throughout the newly added documentation, `False` &rarr; `FALSE` and `True` &rarr; `TRUE`.

sgallagher: OK, I'll make that change and push this. Thanks!

sgallagher:
This was merged into 2.6.0

PR #327, opened 2019-07-03
Refactor tests

sgallagher:
The first patch reworks the way we add new tests to meson.build in order to vastly reduce duplication (and by extension, opportunity for mistakes). This also paves the way for the second patch which reworks how valgrind selects the tests to run against.

### DIFF
--- a/modulemd/v2/meson.build
+++ b/modulemd/v2/meson.build
@@ -136,495 +136,76 @@ test_utils_lib = library(
     install : false,
 )
 
-# -- Test Modulemd.Buildopts (C) -- #
-test_v2_buildopts = executable(
-    'buildopts_v2',
-    'tests/test-modulemd-buildopts.c',
-    dependencies : [
-        modulemd_v2_dep,
-    ],
-    link_with : [
-        test_utils_lib,
-    ],
-    install : false,
-)
-test('buildopts_v2_debug', test_v2_buildopts,
-     env : test_env)
-test('buildopts_v2_release', test_v2_buildopts,
-     env : test_release_env)
+v2_c_tests = {
+'buildopts_v2'           : [ 'tests/test-modulemd-buildopts.c' ],
+'component_module_v2'    : [ 'tests/test-modulemd-component-module.c' ],
+'component_rpm_v2'       : [ 'tests/test-modulemd-component-rpm.c' ],
+'defaults_v2'            : [ 'tests/test-modulemd-defaults.c' ],
+'defaultsv1_v2'          : [ 'tests/test-modulemd-defaults-v1.c' ],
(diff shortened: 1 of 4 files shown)


sgallagher:
Note: I didn't bother reworking the v1 API test creation since that code is on life-support. I made the minimal changes in the second patch to keep valgrind running against those tests instead.

modulemd/v2/meson.build
```diff
@@ -136,495 +136,78 @@ test_utils_lib = library(
...
-        test_utils_lib,
-    ],
-    install : false,
-)
-test('module_index_merger_v2_debug', test_v2_module_index_merger,
-     env : test_env)
-test('module_index_merger_v2_release', test_v2_module_index_merger,
-     env : test_release_env)
+v2_c_tests = {
+'buildopts_v2' : [ 'tests/test-modulemd-buildopts.c' ],
```
mmathesius:
```suggestion
'buildopts_v2'           : [ 'tests/test-modulemd-buildopts.c' ],
```

modulemd/v2/meson.build
```diff
@@ -136,495 +136,78 @@ test_utils_lib = library(
...
+'module_v2'              : [ 'tests/test-modulemd-module.c' ],
+'module_index_v2'        : [ 'tests/test-modulemd-moduleindex.c' ],
+'module_index_merger_v2' : [ 'tests/test-modulemd-merger.c' ],
+'modulestream_v2'        : [ 'tests/test-modulemd-modulestream.c' ],
+'profile_v2'             : [ 'tests/test-modulemd-profile.c' ],
+'rpm_map'                : [ 'tests/test-modulemd-rpmmap.c' ],
+'service_level_v2'       : [ 'tests/test-modulemd-service-level.c' ],
+'translation_v2'         : [ 'tests/test-modulemd-translation.c' ],
+'translation_entry_v2'   : [ 'tests/test-modulemd-translation-entry.c' ],
+}
```
mmathesius: Just curious... Why did you choose to now represent the source file in an array and pass that to `executable()`? It was previously just a simple string.

sgallagher: The sources argument accepts either a single file or a list of files. Currently, all of the tests use only a single file, but I wanted to make it consistent with a future where some tests might require multiple files.

modulemd/v2/meson.build
sgallagher: Missed that one. Fixed now.

modulemd/v2/meson.build
mmathesius: I figured that was intentional, but wanted to know why. Thanks!

PR #328, opened 2019-07-11
TESTS: Check the return value of get_buildtime_streams()

sgallagher:
Silences Coverity defect #341772

### DIFF
--- a/modulemd/v2/tests/test-modulemd-modulestream.c
+++ b/modulemd/v2/tests/test-modulemd-modulestream.c
@@ -866,12 +866,16 @@ module_stream_v2_test_dependencies (ModuleStreamFixture *fixture,
 
   list = modulemd_dependencies_get_buildtime_streams_as_strv (
     g_ptr_array_index (deps_list, 0), "foo");
+  g_assert_nonnull (list);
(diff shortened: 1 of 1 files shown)


modulemd/v2/tests/test-modulemd-modulestream.c
```diff
@@ -866,7 +866,9 @@ module_stream_v2_test_dependencies (ModuleStreamFixture *fixture,
 
   list = modulemd_dependencies_get_buildtime_streams_as_strv (
     g_ptr_array_index (deps_list, 0), "foo");
+  g_assert_nonnull (list);
   g_assert_cmpstr (list[0], ==, "stable");
+  g_assert_null (list[1]);
```
mmathesius: LGTM.

Do you know why the nearly identical call to `g_assert_cmpstr()` a few lines below isn't flagged as a defect--and in need of the same sort of remediation?

sgallagher: I hadn't noticed that, and no I don't know why it's not flagged, but I should probably do this there anyway. Likely Coverity would suddenly discover it once this is merged :-/

PR #334, opened 2019-07-15
Drop v1 code and refactor build system

sgallagher:
Drop all v1 code from the master branch and rework the meson build
scripts to adjust for it. Also makes changes to the way tests
locate their data files to make such refactorings simpler in the
future.


sgallagher:
I'll be sending a similar PR later for the v1 code (on a new `1.x-maint` branch) for building v1 changes if we have urgent fixes that are needed for consumers still using the deprecated API.

mmathesius:
Good riddance!

Everything looks good while skimming it over. I'd like to spend more time tomorrow carefully going through the changes (especially to `modulemd/meson.build`) to make sure I fully understand them before approving.

sgallagher:
Trivially rebased atop the latest doc updates.

mmathesius:
Additional information from @sgallagher:
> A little background on the v1/v2 setup. Originally, my plan was to keep them both in the same source tree because I was intending to have the v1 API as a wrapper around the v2 implementation (so all of the logic would be in v2 but we'd keep a compat layer)
> It turns out that 1) we don't really need that and 2) the APIs are sufficiently different that doing so is far more work than it's worth.
> So I'm just going to split them into the master branch and the 1.x-maint branch
> (Another bonus to this is that we can stop with the constant minor-release bumps of 1.x at each of the 2.x releases).

modulemd/meson.build
```diff
@@ -35,243 +39,397 @@ if not valgrind.found()
...
-    'v2/include/modulemd-2.0/modulemd-module-stream.h',
-    'v2/include/modulemd-2.0/modulemd-module-stream-v1.h',
-    'v2/include/modulemd-2.0/modulemd-module-stream-v2.h',
-    'v2/include/modulemd-2.0/modulemd-profile.h',
-    'v2/include/modulemd-2.0/modulemd-rpm-map-entry.h',
-    'v2/include/modulemd-2.0/modulemd-service-level.h',
-    'v2/include/modulemd-2.0/modulemd-subdocument-info.h',
-    'v2/include/modulemd-2.0/modulemd-translation.h',
-    'v2/include/modulemd-2.0/modulemd-translation-entry.h',
+v2_include_dirs = include_directories (
```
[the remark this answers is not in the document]
sgallagher: That was an oversight. My find-replace missed them :)

sgallagher:
I just pushed a fixup patch with the requested changes. Please re-review and if it passes, I'll squash them together before pushing.

README.md
```diff
@@ -225,11 +225,11 @@ First, decide if you are planning to build the sources for the 1.0 API or the 2.
 
 Projects built with the meson build-system require a separate build directory from the source path. The `meson` command will generate this directory for you.
 ```
-meson --buildtype=debug -Db_coverage=true -Dbuild_api_v1=false -Dbuild_api_v2=true  api2
+meson --buildtype=debug -Db_coverage=true  debugbuild
 ```
-The above command (run from the root of the source checkout) will create a new subdirectory - `api2` - configured to compile with debug symbols and `gcov` symbols to measure test coverage. It enables building the 2.0 API and skips the deprecated 1.0 API.
+The above command (run from the root of the source checkout) will create a new subdirectory - `debugbuild` - configured to compile with debug symbols and `gcov` symbols to measure test coverage. It enables building the 2.0 API and skips the deprecated 1.0 API.
```
mmathesius: The "It enables building the 2.0 API and skips the deprecated 1.0 API." part is now extraneous and should probably be dropped as well.

README.md
```diff
@@ -221,15 +221,13 @@ and to make sure that your user has privilege to run `sudo docker` (see the docu
 The libmodulemd project follows the [Github Fork-and-Pull](https://reflectoring.io/github-fork-and-pull/) model of development. To get started, create a fork of the upstream libmodulemd sources, clone those locally and create branches on your fork to make changes. When they are ready for review or feedback, create a pull-request.
 
 ## Building the sources
-First, decide if you are planning to build the sources for the 1.0 API or the 2.0 API. (Note that the 1.0 API is deprecated and will be removed soon). This guideline will describe the procedure for working with the 2.0 API.
-
```
mmathesius: Good additional catch!

PR #335, opened 2019-07-17
Retrieve Modulemd documents from Koji

OrionStar25:
@sgallagher, I reused most of the code you had written because the `listTagged()` method is still working the same way. There is a little change in the `getBuild()` method though. 

As I have understood it this far, the flow is:

1. Get Koji `tags`.
2. Query Koji using these tags and get tagged builds from Koji.
3. Retrieve `YAML` documents from each build.
4. Create a `ModulemdIndex` object from each `YAML` document.
5. Create a `Babel Catalog` from each `ModulemdIndex` Object.
6. Convert catalogs to `.po` files and send to for translation.
7. Retrieve translations from catalog and create `ModulemdTranslation` documents.
8. Modify `ModulemdIndex` object.
9. Modify `YAML`.
10. Push back to Koji.

### DIFF
--- a/ModulemdTranslationHelpers/Utils.py
+++ b/ModulemdTranslationHelpers/Utils.py
@@ -1,17 +1,18 @@
 import sys
+import logging
+import gi
 from babel.messages import Catalog, pofile
 from datetime import datetime
 from collections import defaultdict
 
 try:
-    import gi
     gi.require_version('Modulemd', '2.0')
     from gi.repository import Modulemd
 except ImportError:
     sys.exit(77)
 
 
-def get_latest_modules_in_tag(session, tag, debug=False):
+def get_latest_modules_in_tag(session, tag):
     """
     Get the most-recently built versions of each (module,stream) pair from
     a Koji tag
@@ -26,10 +27,9 @@ def get_latest_modules_in_tag(session, tag, debug=False):
         try:
             tagged = session.listTagged(tag)
         except requests.exceptions.ConnectionError:
-            if debug:
-                print("Connection lost while retrieving builds for tag %s, "
-                      "retrying..." % tag,
-                      file=sys.stderr)
+            logger.debug(
+                "Connection lost while retrieving builds for tag %s, "
+                "retrying..." % tag)
         else:
             # Succeeded this time, so break out of the loop
             break
@@ -55,25 +55,25 @@ def get_latest_modules_in_tag(session, tag, debug=False):
     return latest
 
 
-def get_index_from_tags(session, tags, debug=False):
+def get_index_from_tags(session, tags):
     """
     Construct a ModuleIndex object from the contents of the provided tags.
     :param session: A Koji session
     :param tags: A set of Koji tags from which module metadata should be pulled
-    :param debug: Whether to print debugging information to the console
     :return: A ModuleIndex object. Raises an exception if any of the
     retrieved modulemd is invalid.
     """
 
     tagged_builds = []
     for tag in tags:
-        tagged_builds.extend(get_latest_modules_in_tag(session, tag, debug))
+        tagged_builds.extend(get_latest_modules_in_tag(session, tag))
 
     # Make the list unique since some modules may have multiple tags
     unique_builds = {}
     for build in tagged_builds:
         unique_builds[build['id']] = build
 
+    index = Modulemd.ModuleIndex.new()
     for build_id in unique_builds.keys():
         # Koji sometimes disconnects for no apparent reason. Retry up to 5
         # times before failing.
@@ -81,27 +81,17 @@ def get_index_from_tags(session, tags, debug=False):
             try:
                 build = session.getBuild(build_id)
             except requests.exceptions.ConnectionError:
-                if debug:
-                    print("Connection lost while processing buildId %s, "
-                          "retrying..." % build_id,
-                          file=sys.stderr)
+                logger.debug(
+                    "Connection lost while processing buildId %s, "
+                    "retrying..." % build_id)
             else:
                 # Succeeded this time, so break out of the loop
                 break
-        if debug:
-            print("Processing %s:%s" % (build['package_name'], build['nvr']))
 
-        modulemds = Modulemd.objects_from_string(
-            build['extra']['typeinfo']['module']['modulemd_str'])
+        logger.info("Processing %s:%s" % (build['package_name'], build['nvr']))
+        ret, failures = index.update_from_string(
+            build['extra']['typeinfo']['module']['modulemd_str'], True)
 
-        # We should only get a single modulemd document from Koji
-        if len(modulemds) != 1:
-            raise ValueError("Koji build %s returned multiple modulemd YAML "
-                             "documents." % build['nvr'])
-
-        index = Modulemd.ModuleIndex.new()
-        ret, failures = index.update_from_string(modulemds, True)
-        
     return index
 
 
@@ -209,7 +199,7 @@ def get_modulemd_translations_from_catalog(catalogs, index):
         try:
             ret = index.add_translation(mmd_translation)
         except GLib.Error:
-            print(
+            logger.debug(
                 "The translation for this %s:%s could not be added",
                 module_name,
                 stream_name)
--- a/ModulemdTranslationHelpers/Utils.py
+++ b/ModulemdTranslationHelpers/Utils.py
@@ -11,6 +11,100 @@ except ImportError:
     sys.exit(77)
 
 
+def get_latest_modules_in_tag(session, tag, debug=False):
+    """
+    Get the most-recently built versions of each (module,stream) pair from
+    a Koji tag
+    :param session: A Koji session
+    :param tag: A koji tag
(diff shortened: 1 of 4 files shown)


ModulemdTranslationHelpers/Utils.py
```diff
@@ -11,6 +11,100 @@
...
+                    print("Connection lost while processing buildId %s, "
+                          "retrying..." % build_id,
+                          file=sys.stderr)
+            else:
+                # Succeeded this time, so break out of the loop
+                break
+        if debug:
+            print("Processing %s:%s" % (build['package_name'], build['nvr']))
+
+        modulemds = Modulemd.objects_from_string(
```
OrionStar25: I'm assuming this function is to retrieve a YAML document. But I don't know if this will work. Also, the `getBuild()` method returns a slightly different set of values. The available values as per the documentation is: 
```
A map will be returned containing the following keys*:
      id: build ID
      package_id: ID of the package built
      package_name: name of the package built
      name: same as package_name
      version
      release
      epoch
      nvr
      state
      task_id: ID of the task that kicked off the build
      owner_id: ID of the user who kicked off the build
      owner_name: name of the user who kicked off the build
      volume_id: ID of the storage volume
      volume_name: name of the storage volume
      creation_event_id: id of the create_event
      creation_time: time the build was created (text)
      creation_ts: time the build was created (epoch)
      start_time: time the build was started (may be null)
      start_ts: time the build was started (epoch, may be null)
      completion_time: time the build was completed (may be null)
      completion_ts: time the build was completed (epoch, may be null)
      source: the SCM URL of the sources used in the build
      extra: dictionary with extra data about the build
```

What values should I be using now?

sgallagher:
> @sgallagher, I reused most of the code you had written because the `listTagged()` method is still working the same way. There is a little change in the `getBuild()` method though.
> 
> As I have understood it this far, the flow is:
> 
> 1. Get Koji `tags`.
> 2. Query Koji using these tags and get tagged builds from Koji.
> 3. Retrieve `YAML` documents from each build.
> 4. Create a `ModulemdIndex` object from each `YAML` document.
> 5. Create a `Babel Catalog` from each `ModulemdIndex` Object.
> 6. Convert catalogs to `.po` files and send to for translation.

This is where you should stop for this portion of the work.

> 7. Retrieve translations from catalog and create `ModulemdTranslation` documents.

This will be part of another process and should be a different PR.

> 8. Modify `ModulemdIndex` object.
> 9. Modify `YAML`.
> 10. Push back to Koji.

Steps 8, 9 and 10 are incorrect. We don't push the translations up to Koji. We will inject them into the repodata when generating and publishing the Fedora repository. This is done as a release engineering task and is out of scope for this part of the project. So don't worry about this part yet.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -11,6 +11,100 @@
     sys.exit(77)
 
 
+def get_latest_modules_in_tag(session, tag, debug=False):
```
```suggestion
def get_latest_modules_in_tag(session, tag):
```

Drop the debug print statements in the code here and switch to using python's [logging](https://docs.python.org/3/library/logging.html) instead.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -11,6 +11,100 @@
...
+    """
+
+    # Koji sometimes disconnects for no apparent reason. Retry up to 5
+    # times before failing.
+    for attempt in range(5):
+        try:
+            tagged = session.listTagged(tag)
+        except requests.exceptions.ConnectionError:
+            if debug:
+                print("Connection lost while retrieving builds for tag %s, "
```
sgallagher: Instead of `if debug:` and `print()`, just use `debug()` from python logging.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -11,6 +11,100 @@
...
+
+    for build_id in unique_builds.keys():
+        # Koji sometimes disconnects for no apparent reason. Retry up to 5
+        # times before failing.
+        for attempt in range(5):
+            try:
+                build = session.getBuild(build_id)
+            except requests.exceptions.ConnectionError:
+                if debug:
+                    print("Connection lost while processing buildId %s, "
```
sgallagher: Instead of `if debug:` and `print()`, just use `debug()` from python logging.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -11,6 +11,100 @@
...
+    latest = []
+    for name in NSVs:
+        for stream in NSVs[name]:
+            version = sorted(list(NSVs[name][stream].keys()))[-1]
+            latest.extend(NSVs[name][stream][version])
+
+    return latest
+
+
+def get_index_from_tags(session, tags, debug=False):
```
```suggestion
def get_index_from_tags(session, tags):
```

Use the python logging facility instead.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -11,6 +11,100 @@
...
+    """
+    Construct a ModuleIndex object from the contents of the provided tags.
+    :param session: A Koji session
+    :param tags: A set of Koji tags from which module metadata should be pulled
+    :param debug: Whether to print debugging information to the console
```
```suggestion
```

ModulemdTranslationHelpers/Utils.py
```diff
@@ -11,6 +11,100 @@
...
+    :return: A ModuleIndex object. Raises an exception if any of the
+    retrieved modulemd is invalid.
+    """
+
+    tagged_builds = []
+    for tag in tags:
+        tagged_builds.extend(get_latest_modules_in_tag(session, tag, debug))
```
```suggestion
        tagged_builds.extend(get_latest_modules_in_tag(session, tag))
```

ModulemdTranslationHelpers/Utils.py
sgallagher: The modulemd YAML is all under the `extra` key and the existing way of accessing it (`build['extra']['typeinfo']['module']['modulemd_str']`) should be correct.

This used the old libmodulemd1 approach of getting the YAML, but it's completely out of date. You can actually simplify this immensely by just creating the index before this loop and then calling `index.update_from_string(build['extra']['typeinfo']['module']['modulemd_str'])` multiple times through the loop. Each `update_from_string()` will add the found content to the `ModulemdIndex`.

I think it's probably safe to just drop the `len(modulemds)` check, since it should be impossible and it will simplify things a lot here.

OrionStar25: Ohhh I got it now. 
`build['extra']['typeinfo']['module']['modulemd_str']` was accessing nested components of `extra`. I misunderstood them all as `keys` of the build.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -1,16 +1,100 @@
 import sys
+import logging
+import gi
 from babel.messages import Catalog, pofile
 from datetime import datetime
 from collections import defaultdict
 
 try:
```
sgallagher: This whole `try`/`except` block should be removed, actually. The `sys.exit(77)` thing is only intended for tests. If the translation helpers don't have python-libmodulemd, it should just outright fail.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -1,16 +1,100 @@
...
+    :return: A list of the most recent build of all modules in the tag.
+    """
+
+    # Koji sometimes disconnects for no apparent reason. Retry up to 5
+    # times before failing.
+    for attempt in range(5):
+        try:
+            tagged = session.listTagged(tag)
+        except requests.exceptions.ConnectionError:
+            logger.debug(
```
```suggestion
            logger.warning(
```

ModulemdTranslationHelpers/Utils.py
```diff
@@ -1,16 +1,100 @@
...
+
+    index = Modulemd.ModuleIndex.new()
+    for build_id in unique_builds.keys():
+        # Koji sometimes disconnects for no apparent reason. Retry up to 5
+        # times before failing.
+        for attempt in range(5):
+            try:
+                build = session.getBuild(build_id)
+            except requests.exceptions.ConnectionError:
+                logger.debug(
```
```suggestion
                logger.warning(
```

ModulemdTranslationHelpers/Utils.py
```diff
@@ -1,16 +1,100 @@
...
+                    "Connection lost while processing buildId %s, "
+                    "retrying..." % build_id)
+            else:
+                # Succeeded this time, so break out of the loop
+                break
+
+        logger.info("Processing %s:%s" % (build['package_name'], build['nvr']))
```
```suggestion
        logger.debug("Processing %s:%s" % (build['package_name'], build['nvr']))
```

ModulemdTranslationHelpers/Utils.py
```diff
@@ -115,7 +199,7 @@ def get_modulemd_translations_from_catalog(catalogs, index):
         try:
             ret = index.add_translation(mmd_translation)
         except GLib.Error:
-            print(
+            logger.debug(
```
```suggestion
            logger.warning(
```

OrionStar25:
@sgallagher I rebased atop master but it ended with a `v2` folder inside `modulemd` anyway :confused: Since I've made changes in just 1 file, I'm closing this PR and opening a new one with the suggested changes.

PR #339, opened 2019-07-18
Retrieve Modulemd documents from Koji

OrionStar25:
Continuing #335 here. 

List of tasks taken care of in this PR:

1. Get Koji tags.
2. Query Koji using these tags and get tagged builds from Koji.
3. Retrieve YAML documents from each build.
4. Create a ModulemdIndex object from each YAML document.

### DIFF
--- a/ModulemdTranslationHelpers/Utils.py
+++ b/ModulemdTranslationHelpers/Utils.py
@@ -1,14 +1,99 @@
 import sys
+import logging
+import gi
+import requests
+import koji
 from babel.messages import Catalog, pofile
 from datetime import datetime
 from collections import defaultdict
 
-try:
-    import gi
-    gi.require_version('Modulemd', '2.0')
-    from gi.repository import Modulemd
-except ImportError:
-    sys.exit(77)
+gi.require_version('Modulemd', '2.0')  # noqa
+from gi.repository import Modulemd
+
+
+def get_latest_modules_in_tag(session, tag):
+    """
+    Get the most-recently built versions of each (module,stream) pair from
+    a Koji tag
+    :param session: A Koji session
+    :param tag: A koji tag
+    :return: A list of the most recent build of all modules in the tag.
+    """
+
+    # Koji sometimes disconnects for no apparent reason. Retry up to 5
+    # times before failing.
+    for attempt in range(5):
+        try:
+            tagged = session.listTagged(tag)
+        except requests.exceptions.ConnectionError:
+            logging.warning(
+                "Connection lost while retrieving builds for tag %s, "
+                "retrying..." % tag)
+        else:
+            # Succeeded this time, so break out of the loop
+            break
+
+    # Find the latest, in module terms.  Pungi does this.
+    # Collect all contexts that share the same NSV.
+    NSVs = {}
+    for entry in tagged:
+        name, stream = entry['name'], entry['version']
+        version = entry['release'].rsplit('.', 1)[0]
+
+        NSVs[name] = NSVs.get(name, {})
+        NSVs[name][stream] = NSVs[name].get(stream, {})
+        NSVs[name][stream][version] = NSVs[name][stream].get(version, [])
+        NSVs[name][stream][version].append(entry)
+
+    latest = []
+    for name in NSVs:
+        for stream in NSVs[name]:
+            version = sorted(list(NSVs[name][stream].keys()))[-1]
+            latest.extend(NSVs[name][stream][version])
+
+    return latest
+
+
+def get_index_from_tags(session, tags):
+    """
+    Construct a ModuleIndex object from the contents of the provided tags.
+    :param session: A Koji session
+    :param tags: A set of Koji tags from which module metadata should be pulled
+    :return: A ModuleIndex object. Raises an exception if any of the
+    retrieved modulemd is invalid.
+    """
+
+    tagged_builds = []
+    for tag in tags:
+        tagged_builds.extend(get_latest_modules_in_tag(session, tag))
+
+    # Make the list unique since some modules may have multiple tags
+    unique_builds = {}
+    for build in tagged_builds:
+        unique_builds[build['id']] = build
+
+    index = Modulemd.ModuleIndex.new()
+    for build_id in unique_builds.keys():
+        # Koji sometimes disconnects for no apparent reason. Retry up to 5
+        # times before failing.
+        for attempt in range(5):
+            try:
+                build = session.getBuild(build_id)
+            except requests.exceptions.ConnectionError:
+                logging.warning(
+                    "Connection lost while processing buildId %s, "
+                    "retrying..." % build_id)
+            else:
+                # Succeeded this time, so break out of the loop
+                break
+
+        logging.debug(
+            "Processing %s:%s" %
+            (build['package_name'], build['nvr']))
+        ret, failures = index.update_from_string(
+            build['extra']['typeinfo']['module']['modulemd_str'], True)
+
+    return index
 
 
 def get_translation_catalog_from_index(index, project_name):
@@ -115,7 +200,7 @@ def get_modulemd_translations_from_catalog(catalogs, index):
         try:
             ret = index.add_translation(mmd_translation)
         except GLib.Error:
-            print(
+            logging.warning(
                 "The translation for this %s:%s could not be added",
                 module_name,
                 stream_name)
--- a/modulemd/tests/ModulemdTests/translation_helpers.py
+++ b/modulemd/tests/ModulemdTests/translation_helpers.py
@@ -14,12 +14,13 @@
 
 import os
 import sys
+import unittest
 from ModulemdTranslationHelpers import Utils
 from babel.messages import pofile
 from six import text_type
+from unittest import mock
 
 try:
-    import unittest
     import gi
     gi.require_version('Modulemd', '2.0')
     from gi.repository import Modulemd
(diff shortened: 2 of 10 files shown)


ModulemdTranslationHelpers/Utils.py
```diff
@@ -1,14 +1,97 @@
...
 from datetime import datetime
 from collections import defaultdict
 
-try:
-    import gi
-    gi.require_version('Modulemd', '2.0')
-    from gi.repository import Modulemd
-except ImportError:
-    sys.exit(77)
+gi.require_version('Modulemd', '2.0')
```
OrionStar25: > @sgallagher : This whole try/except block should be removed, actually. The sys.exit(77) thing is only intended for tests. If the translation helpers don't have python-libmodulemd, it should just outright fail.

I actually had done that to escape the error generated by `pycodestyle` saying all imports should be on top of the file.

sgallagher: You can fix that with a `# noqa` line, which is appropriate in this case.

```suggestion
gi.require_version('Modulemd', '2.0')  # noqa
```
(Note that it must be over by two spaces from the end of the line or pycodestyle will complain).

sgallagher:
I think this looks good, but I'd like to see you write some tests for it. Build the tests around the `f29` tag, because that release is "frozen" and will always have the same content.

OrionStar25:
- [ ] I'll have to look into how to cache the results (I still haven't gotten 1 successful call yet) [TODO]

- [ ] I was wondering if it's possible to have an `.equals()` for `ModulemdIndex` objects as well. Right now I've attempted to check if `ModulemdStream` objects are identical for each module.

sgallagher:
> * [ ]  I was wondering if it's possible to have an `.equals()` for `ModulemdIndex` objects as well. Right now I've attempted to check if `ModulemdStream` objects are identical for each module.

Probably the easiest way to do this would be:
```
self.assertEquals(idx.dump_to_string(), index.dump_to_string)
```
That said, this isn't going to work in this case. The YAML in `f29.yaml` contains modulemd-defaults objects, but the YAML in Koji will only contain module stream objects. So those two indices will never be equal.

If you want to check equality like that, you'll need to retrieve that YAML and save it to the test_data as a new filename and compare it that way.

OrionStar25:
@sgallagher Added the case for empty index. Not quite sure about the other case (commented code).

Conan-Kudo:
Doesn't this add Koji as a dependency to libmodulemd now?

sgallagher:
> Doesn't this add Koji as a dependency to libmodulemd now?

No, this will be a separate python library shipped out of the libmodulemd sources. libmodulemd itself (and python-libmodulemd) will *not* have a koji dependency.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -1,14 +1,111 @@
...
-except ImportError:
-    sys.exit(77)
+gi.require_version('Modulemd', '2.0')  # noqa
+from gi.repository import Modulemd
+
+
+class KojiMock:
+
+    def __init__(self, url):
+        self.session = koji.ClientSession(baseurl=url)
```
sgallagher: The whole point of creating a mock object is to *not* actually use `koji.ClientSession`. You instead create an object that simulates it.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -1,14 +1,111 @@
...
+
+    def listTagged(self, tag):
+        return self.session.listTagged(tag)
```
sgallagher: This shouldn't be calling out to koji. You just want to have this return a fixed list of objects in the format that `listTagged()` would return.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -1,14 +1,111 @@
...
+
+    def getBuild(self, build_id):
+        return self.session.getBuild(build_id)
```
sgallagher: Same here. What you want here is a `dict()` keyed on `build_id` that has values in the format of `getBuild()`'s return type.

OrionStar25:
@sgallagher 
My index is not being created and I'm not able to understand how to debug this. Is this approach correct though?

modulemd/tests/ModulemdTests/translation_helpers.py
```diff
@@ -137,6 +138,36 @@ def test_translations_from_catalog(self):
...
+
+        # Test when koji returns no modules
+        mock_session.listTagged.return_value = []
+        mock_session.getBuild.return_value = None
+
+        index = Utils.get_index_from_tags(mock_session, tags)
+        self.assertIsNotNone(index)
+        mock_session.listTagged.assert_called_once_with('f29')
+        module_names = index.get_module_names()
+        self.assertFalse(module_names)
```
sgallagher: What is this supposed to mean?

modulemd/tests/ModulemdTests/translation_helpers.py
```diff
@@ -137,6 +138,36 @@ def test_translations_from_catalog(self):
...
+        idx = Modulemd.ModuleIndex.new()
+        self.assertIsNotNone(idx)
+        ret, failures = idx.update_from_file(
+            "%s/f29.yaml" %
+            (os.getenv('TEST_DATA_PATH')), True)
+        self.assertTrue(ret)
+        self.assertTrue(len(failures) == 0)
+
+         # Test when koji returns proper modules
+        index = Utils.get_index_from_tags(mock_session, tags)
```
sgallagher: Before calling into `Utils.get_index_from_tags()`, you need to set up the mock functions for `mock_session`. For example:

```python3
        koji_session_mock = Utils.KojiSessionMock()

        mock_session.listTagged.side_effect = koji_session_mock.listTagged
        mock_session.getBuild.side_effect = koji_session_mock.getBuild
```

modulemd/tests/ModulemdTests/translation_helpers.py
OrionStar25: This test says that a `ModulemdIndex` object will me created but it would be empty.

OrionStar25: I added a test for a case when `listTagged` wouldn't return anything. In that case, there would be an empty `ModulemdIndex` object.

ModulemdTranslationHelpers/Utils.py
```diff
@@ -1,14 +1,136 @@
...
-except ImportError:
-    sys.exit(77)
+gi.require_version('Modulemd', '2.0')  # noqa
+from gi.repository import Modulemd
+
+
+class KojiSessionMock:
+
+    def __init__(self):
+        with open('../modulemd/tests/test_data/f29.yaml', 'r') as file:
```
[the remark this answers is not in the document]
OrionStar25: > A proper test here would return just a single module stream document per build_id

I didn't quite understand this bit. Am I not assigning a document only after receiving the `build_id`?

OrionStar25:
@sgallagher I tried to add `koji` as a dependency in each of the files, and it said no target found. So I removed the import. It then said `No module named requests`. so I added the corresponding module looking at the official package manager for each distro. Still, the module is not able to install only.

OrionStar25:
Okay, the error over here still says `No koji`, which is different from the local error :confused:

OrionStar25:
@sgallagher These are all the different errors:

Mageia: `No module name gi` How was this working before then?
Arch Linux: `target not found: koji`  Its [here](https://aur.archlinux.org/packages/koji/)
OpenSuse: `No provider of 'koji' found.` 
Fedora 30: `cannot import name mock` (for python2) 
Fedora 29: `Failed to synchronize cache for repo 'fedora-modular'`
Fedora Rawhide: `Unable to find a match: python2-koji` Its [here](https://apps.fedoraproject.org/packages/python2-koji)

sgallagher:
> ~Okay, the error over here still says `No koji`, which is different from the locally-run docker error~
> I'm getting this same error locally. And opensuse apparently doesn't have any package like `koji/python koji`

Koji is a Fedora tool. It is not expected to exist on other platforms; none of the Translation tests should be run anywhere but Fedora and CentOS. This is part of why I want to move the TranslationHelpers code into its own section in the source hierarchy.

sgallagher:
> @sgallagher These are all the different errors:
> 
> Mageia: `No module name gi` How was this working before then?
> Arch Linux: `target not found: koji` Its [here](https://aur.archlinux.org/packages/koji/)
> OpenSuse: `No provider of 'koji' found.`

The above are not expected to have Koji and shouldn't try to run these tests (or build this python module).

> Fedora 30: `cannot import name mock` (for python2)
Python 2 is dead on Fedora 30+. Only Python 3 is supported.

> Fedora 29: `Failed to synchronize cache for repo 'fedora-modular'`
That's a temporary failure; restarting the test should complete fine.

> Fedora Rawhide: `Unable to find a match: python2-koji` Its [here](https://apps.fedoraproject.org/packages/python2-koji)

Rawhide doesn't build python 2 packages anymore except for a very small number of critical ones.

ModulemdTranslationHelpers/Utils.py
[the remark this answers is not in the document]
sgallagher: I mean that you shouldn't use f29.yaml for this test, because we actually want the lookups to give us back individual module stream objects (so a single-document YAML) rather than the complete repodata.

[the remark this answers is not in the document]
OrionStar25: @sgallagher 
So I have created a new test file which has the sum of 2 individual module stream objects. I hope this is what you meant!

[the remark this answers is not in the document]
OrionStar25: I've added 2 module stream objects of `f29` that happen to be the first 2 strings in all the `.po` files.

OrionStar25:
@sgallagher Travis is failing and the logs aren't showing the error again. `./make_rpms.sh` worked perfectly in mine.

Conan-Kudo:
@sgallagher Koji is packaged in Mageia, we just only ship the Python 3 variant. But there's no Koji instance to target yet...

sgallagher:
> @sgallagher Travis is failing and the logs aren't showing the error again. `./make_rpms.sh` worked perfectly in mine.

Edit `make_rpms.sh.in` and drop the `> /dev/null` from the `ninja dist` line. Right now it's suppressing the output.

I need to rework that so it prints the output if the process fails; I'll send a PR.

OrionStar25:
@sgallagher Everything is passing now, except for Rawhide like expected.

OrionStar25:
@sgallagher Error for Fedora 29:

> Error: Failed to synchronize cache for repo 'updates'

sgallagher:
> @sgallagher Error for Fedora 29:
> 
> > Error: Failed to synchronize cache for repo 'updates'

That's a transient error. I restarted the test.

PR #342, opened 2019-07-19
Emitter: Throw validation error for empty ModuleIndex

sgallagher:
Attempting to emit an empty `ModuleIndex` (one containing no modules) currently returns NULL for `dump_to_string()` or writes no data into `dump_to_file()` or `dump_to_custom()`. With this patch, it will return a `GError` indicating a validation error.

### DIFF
--- a/modulemd/tests/test-modulemd-moduleindex.c
+++ b/modulemd/tests/test-modulemd-moduleindex.c
@@ -1002,6 +1002,22 @@ module_index_test_get_default_streams (void)
 }
 
 
+static void
+module_index_test_dump_empty_index (void)
+{
+  g_autoptr (GError) error = NULL;
+  g_autofree gchar *yaml = NULL;
+  g_autoptr (ModulemdModuleIndex) index = modulemd_module_index_new ();
+
+  g_assert_nonnull (index);
+  g_assert_true (MODULEMD_IS_MODULE_INDEX (index));
+
+  yaml = modulemd_module_index_dump_to_string (index, &error);
+  g_assert_null (yaml);
+  g_assert_error (error, MODULEMD_ERROR, MODULEMD_ERROR_VALIDATE);
+}
+
+
 int
 main (int argc, char *argv[])
 {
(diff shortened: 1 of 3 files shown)


sgallagher:
The alternative would be to make strings behave like the file/custom output and return zero-length, but that feels to me like it would eventually cause more problems than it fixes.

modulemd/tests/test-modulemd-moduleindex.c
```diff
@@ -1002,6 +1002,22 @@ module_index_test_get_default_streams (void)
...
 
+static void
+module_index_test_dump_empty_index (void)
+{
+  g_autoptr (GError) error = NULL;
+  g_autofree gchar *yaml = NULL;
+  g_autoptr (ModulemdModuleIndex) index = modulemd_module_index_new ();
+
+  g_assert_nonnull (index);
+  g_assert_true (MODULEMD_IS_MODULE_INDEX);
```
mmathesius: The compiler is complaining about this assertion:
```
../modulemd/tests/test-modulemd-moduleindex.c: In function 'module_index_test_dump_empty_index':
../modulemd/tests/test-modulemd-moduleindex.c:1013:18: warning: the address of 'MODULEMD_IS_MODULE_INDEX' will always evaluate as 'true' [-Waddress]
 1013 |   g_assert_true (MODULEMD_IS_MODULE_INDEX);
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~
/usr/include/glib-2.0/glib/gmacros.h:900:25: note: in definition of macro 'G_LIKELY'
  900 | #define G_LIKELY(expr) (expr)
      |                         ^~~~
../modulemd/tests/test-modulemd-moduleindex.c:1013:3: note: in expansion of macro 'g_assert_true'
 1013 |   g_assert_true (MODULEMD_IS_MODULE_INDEX);
      |   ^~~~~~~~~~~~~
```

mmathesius: Shouldn't it be `g_assert_true (MODULEMD_IS_MODULE_INDEX (index));`?

sgallagher: Yes, it should be. Weird that it wasn't a compilation error though.

PR #350, opened 2019-07-29
Component buildonly property fix

mmathesius:
Correct bugs--likely the result of copy & paste errors--where component `buildonly` and `buildorder` properties are getting mixed up. Add python tests for same.


sgallagher:
> Great catch! Please merge the two patches, though. I don't want there to be a commit in the history that will fail tests, as it will make it harder to use `git bisect` in the future.

By "merge" I of course meant "squash together".

Issue #288, opened 2019-05-07
RFE: add flags to include packages in (srpm-)buildroot


ignatenkobrain:
Imagine that I have 2 components, `rust-srpm-macros` which defines `%{rust_arches}` and `rust` which is using it in `ExclusiveArch`. The former one has to be in srpm-buildroot in order for latter to build. That means, at this point I have only option to use 2 module (rust and rust-bootstrap).. But rust needs its own bootstrapping, so I would need to have 3 modules.

Having something like:

```yaml
rpms:
  rust-srpm-macros:
    rationale: Main component.
    srpm-buildroot: true
  rust:
    rationale: Main component.
    buildafter: [rust-srpm-macros]
```

Would be good. Also for completeness, adding `buildroot` flag.

ignatenkobrain:
As discussed with @contyk in a pub :)

ignatenkobrain:
@mprahl @jkaluza how hard would this be to implement in MBS?

mprahl:
@ignatenkobrain this shouldn't be too hard in MBS. If this becomes part of the libmodulemd spec, we can support it in MBS.

Adding @lcarva for visibility.

mprahl:
@ignatenkobrain I noticed that in your example that you use `buildafter`. Would this feature still be useful to you even without `buildafter`? I ask because supporting `buildafter` is a lot more work, and the earliest we'll be able to work on that is in June (likely later).

sgallagher:
@ignatenkobrain @mprahl This slipped off my radar. Is this still something that is desired? I can get it into 2.7.0 if so.

ignatenkobrain:
Yes, please.

sgallagher:
@mmathesius Would you please add a new boolean attribute to `ModulemdComponentRpm`, along with the appropriate getter, setter and tests?

mmathesius:
Just to make sure I've got the semantics right, is https://github.com/mmathesius/libmodulemd/commit/f035b2754e4abdab502ac0e46f0c2aeb3b7a98a4 an accurate representation of the spec change corresponding to this RFE?

sgallagher:
Hmm, I'm not sure exactly what the point of the `buildroot` argument would be here. The `srpm-buildroot` makes sense, but `buildroot` should just be handled by the `buildorder/buildafter` functionality, I think.

@ignatenkobrain Can you explain why you think `buildroot` is useful?

ignatenkobrain:
[...]

PR #351, opened 2019-07-30
Fix modulemd_yaml_parse_bool()

mmathesius:
`modulemd_yaml_parse_bool()` currently returns `TRUE` when parsing a YAML `false` value, and `FALSE` when parsing `true`.

This PR fixes that issue, as well as corrects the defective test that was enforcing the backwards parsing behavior.


mmathesius:
> Ouch, that's a painful bug. Good catch!

Yes, thanks. (As I was trying to figure out why my WIP for #288 wasn't working as expected. :confused:)

PR #352, opened 2019-07-30
Stop using deprecated build_always

sgallagher:
Since we now require meson 0.47+ for building, we no longer need to
use the deprecated `build_always` argument for the specfile template
script. Switch to using the supported `build_always_stale` and
`build_by_default` arguments instead.


sgallagher:
For a time, we had been supporting 0.46 because that was what had been in RHEL 7, but since that got updated and we started using other 0.47+ features, it was time to clean up this warning.

PR #357, opened 2019-08-02
Print `ninja dist` logs if it fails

sgallagher:

### DIFF
--- a/make_rpms.sh.in
+++ b/make_rpms.sh.in
@@ -7,7 +7,14 @@ set -e
 
 echo "Creating tarball. This may take a while."
 
-MMD_SKIP_VALGRIND=True ninja dist > /dev/null
+set +e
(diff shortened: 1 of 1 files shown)


sgallagher:
The `ninja dist` output can be long and noisy on success, but we probably want to see those logs when it fails.

sgallagher:
> Why is Rawhide showing error for `translation_helpers`? Does this mean its also failing in the master branch? Btw, this is the same error I was getting in my PR #339 for Rawhide.

That looks like someone broke pytz in Rawhide. It's missing a critical file.

OrionStar25:
> That looks like someone broke pytz in Rawhide. It's missing a critical file.

So should we be ignoring this?

sgallagher:
> > That looks like someone broke pytz in Rawhide. It's missing a critical file.
> 
> So should we be ignoring this?

Yes, you can ignore that for now.

PR #358, opened 2019-08-05
Drop ModulemdTranslationHelpers

sgallagher:
This is being moved to its own project. See
https://github.com/fedora-modularity/ModulemdTranslationHelpers/pull/1


sgallagher:
> I think we also need to remove `python3-koji, python3-babel, python3-requests` from the `fedora.deps` file.

I think you're right. I just pushed an amended patch to do so.

PR #362, opened 2019-08-30
Add support for reading compressed files into ModuleIndex

sgallagher:
With this patch, `Modulemd.ModuleIndex.update_from_file()` will now automatically detect if it's accessing a compressed file and pass it through a decompression wrapper (if compiled with an optional dependency on librpm). To disable this functionality (such as on systems without librpm), pass `-Drpmio=disabled` to `meson`. The supported compression types are `gz`, `bzip2` and `xz`. Support for `zck` may come in the future, but it wasn't trivially usable from librpm, so I deferred it.

It adds a second optional dependency on libmagic (to autodetect the compression type if the suffix doesn't make it obvious).

This adds read support only; I don't think right now it makes sense to have libmodulemd support outputting it to a compressed format; that is usually handled by createrepo_c.

Also includes a minor patch for the Valgrind test to work around a race-condition causing occasional failures.

Fixes: https://github.com/fedora-modularity/libmodulemd/issues/208


Conan-Kudo:
@sgallagher Looks like this code broke on EL7...?

sgallagher:
@Conan_Kudo: Hmm, you're right. Something went wrong with the bzip2 support on EL7 it seems. It's possible that librpm doesn't support bzip2 on that platform...

modulemd/include/modulemd-2.0/modulemd-compression.h
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * SECTION: modulemd-compression
+ * @title: Modulemd Compression Helpers
+ * @stability: stable
+ * @short_description: Utility functions for working with compressed files.
+ */
+
+
+/**
+ * ModulemdCompressionTypeEnum
```
mmathesius:
```suggestion
 * ModulemdCompressionTypeEnum:
```

Conan-Kudo:
@sgallagher I'm pretty sure bzip2 payload support is supposed to work on EL7, so it's probably a bug.

sgallagher:
> @sgallagher I'm pretty sure bzip2 payload support is supposed to work on EL7, so it's probably a bug.

Yeah, I'm chasing this down the rabbit-hole. It's surprisingly complicated.

modulemd/modulemd-compression.c
```diff
@@ -0,0 +1,277 @@
...
+
+ModulemdCompressionTypeEnum
+modulemd_compression_type (const gchar *name)
+{
+  if (!name)
+    return MODULEMD_COMPRESSION_TYPE_UNKNOWN_COMPRESSION;
+
+  int type = MODULEMD_COMPRESSION_TYPE_UNKNOWN_COMPRESSION;
+
+  if (!g_strcmp0 (name, "gz") || !g_strcmp0 (name, "gzip"))
```
mmathesius: `modulemd_detect_compression()` also recognizes "gunzip" as a valid suffix for `MODULEMD_COMPRESSION_TYPE_GZ_COMPRESSION`. Should that suffix be included here as well for consistency?

sgallagher:
So, it turned out to be a subtle difference in libmagic on EL7: newer versions of `magic_descriptor()` will reset the file pointer back to the start of the file before returning, whereas the version on EL7 did not do this. So I just added `lseek (fd, 0, SEEK_SET);` to the end of the `modulemd_detect_compression()` function.

I also fixed the doc issue that @mmathesius pointed out as well as fixing a test to work correctly when `-Dlibmagic=disabled` and `-Drpmio=enabled`.

Also, it may be that the seek issue was the root of my problems with zchunk earlier, so I'm going to have another look at it now.

modulemd/modulemd-compression.c
```diff
@@ -0,0 +1,277 @@
...
+    {
+      g_set_error (error,
+                   MODULEMD_ERROR,
+                   MODULEMD_ERROR_MAGIC,
+                   "mime_type() detection failed: %s",
+                   magic_error (magic));
+      return MODULEMD_COMPRESSION_TYPE_DETECTION_FAILED;
+    }
+  /* Reset the file descriptor to the start of the file, if it has moved */
+  lseek (fd, 0, SEEK_SET);
```
mmathesius: I'd suggest moving this to right after `close(magic_fd);` above--in case the caller decides to proceed with reading from `fd` (perhaps treating it as uncompressed) even when `MODULEMD_COMPRESSION_TYPE_DETECTION_FAILED` is returned.

modulemd/modulemd-compression.c
sgallagher: Agreed

modulemd/include/modulemd-2.0/modulemd-compression.h
```diff
@@ -0,0 +1,69 @@
...
+ * @MODULEMD_COMPRESSION_TYPE_AUTO_DETECT_COMPRESSION: Autodetection
```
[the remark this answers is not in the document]
sgallagher: Whoops; vestige of an earlier thought that I revised.

modulemd/include/modulemd-2.0/modulemd-compression.h
```diff
@@ -0,0 +1,69 @@
...
+typedef enum
+{
+  MODULEMD_COMPRESSION_TYPE_DETECTION_FAILED = -2,
+  MODULEMD_COMPRESSION_TYPE_UNKNOWN_COMPRESSION,
+  MODULEMD_COMPRESSION_TYPE_NO_COMPRESSION,
+  MODULEMD_COMPRESSION_TYPE_GZ_COMPRESSION,
+  MODULEMD_COMPRESSION_TYPE_BZ2_COMPRESSION,
+  MODULEMD_COMPRESSION_TYPE_XZ_COMPRESSION,
+  MODULEMD_COMPRESSION_TYPE_ZCK_COMPRESSION,
+  MODULEMD_COMPRESSION_TYPE_SENTINEL,
```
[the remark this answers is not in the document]
sgallagher: I excluded `MODULEMD_COMPRESSION_TYPE_SENTINEL` intentionally, since it's not intended for public use, but I can note that.

modulemd/include/modulemd-2.0/modulemd-errors.h
```diff
@@ -47,7 +49,9 @@ typedef enum
   MODULEMD_ERROR_VALIDATE,
   MODULEMD_ERROR_FILE_ACCESS,
   MODULEMD_ERROR_NO_MATCHES,
-  MODULEMD_ERROR_TOO_MANY_MATCHES
+  MODULEMD_ERROR_TOO_MANY_MATCHES,
+  MODULEMD_ERROR_MAGIC,
+  MODULEMD_ERROR_NOT_IMPLEMENTED
```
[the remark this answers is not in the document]
sgallagher: Done

modulemd/include/private/modulemd-compression-private.h
```diff
@@ -0,0 +1,95 @@
...
+ */
+
+
+/**
+ * modulemd_detect_compression:
+ * @filename: (in): The original filename that matches @fd.
+ * @fd: (in): An open file descriptor pointing at a real file.
+ * @error: (out): A #GError containing the reason this function failed.
+ *
+ * Returns: #The ModulemdCompressionTypeEnum detected from this file
```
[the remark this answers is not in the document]
sgallagher: Done

modulemd/include/private/modulemd-compression-private.h
```diff
@@ -0,0 +1,95 @@
...
+ * Since: 2.8
+ */
+gchar *
+modulemd_get_rpmio_fmode (const gchar *mode,
+                          ModulemdCompressionTypeEnum comtype);
+
+
+/**
+ * compressed_stream_read_fn:
+ *
```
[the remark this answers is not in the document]
sgallagher: Done

modulemd/include/private/modulemd-compression-private.h
```diff
@@ -0,0 +1,95 @@
...
+#include <glib.h>
+
+#include "config.h"
+#include "modulemd-compression.h"
+
+#ifdef HAVE_RPMIO
+#include <rpm/rpmio.h>
+void
+mmd_Fclose (FD_t fd);
+G_DEFINE_AUTO_CLEANUP_FREE_FUNC (FD_t, mmd_Fclose, NULL);
```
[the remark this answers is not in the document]
sgallagher: Done

modulemd/modulemd-compression.c
sgallagher: Good idea, done.

modulemd/include/modulemd-2.0/modulemd-compression.h
```diff
@@ -0,0 +1,69 @@
...
+ * ModulemdCompressionTypeEnum:
+ * @MODULEMD_COMPRESSION_TYPE_DETECTION_FAILED: Autodetection failure
+ * @MODULEMD_COMPRESSION_TYPE_UNKNOWN_COMPRESSION: Unknown compression
+ * @MODULEMD_COMPRESSION_TYPE_NO_COMPRESSION: No compression
+ * @MODULEMD_COMPRESSION_TYPE_GZ_COMPRESSION: gzip compression
+ * @MODULEMD_COMPRESSION_TYPE_BZ2_COMPRESSION: bzip2 compression
+ * @MODULEMD_COMPRESSION_TYPE_XZ_COMPRESSION: LZMA compression
+ * @MODULEMD_COMPRESSION_TYPE_ZCK_COMPRESSION: zchunk compression
+ * @MODULEMD_COMPRESSION_TYPE_SENTINEL: Enum list terminator
+ * Since: 2.8
```
mmathesius: Please add a blank line before the "Since: 2.8" so it doesn't get concatenated to the last enum's description.

sgallagher: Fixed

sgallagher:
@Conan-Kudo Would you mind also reviewing this before I merge it upstream? You were the original requester of this feature after all.

Conan-Kudo:
@sgallagher What about support zchunk? I see stubs for it, but no implementation... You mentioned you had problems initially...?

Also, what about that `libsquishy` library of yours you've made... Wouldn't that be better than using librpmio for this?

sgallagher:
> @sgallagher What about support zchunk? I see stubs for it, but no implementation... You mentioned you had problems initially...?

Right, I haven't tracked it down entirely yet, but whenever I try to pass a .zck file to rpmio using `"r.zstdio"`, it enters and infinite loop. I'm working on building a minimal reproducer to submit to RPM.

> Also, what about that `libsquishy` library of yours you've made... Wouldn't that be better than using librpmio for this?

That was my original plan, but I had a conversation with @kontura and he indicated that now that rpmio has support for this stuff, he was planning to drop createrepo_c's wrapper anyway and just use that. I took a look at how rpmio does things and realized it would be pretty easy to just do the same, so I went that route too. It'll be better than having to maintain libsquishy in the long run.

Conan-Kudo:
@sgallagher That'd be problematic, since librpmio has zero support for zchunk right now.

sgallagher:
@Conan-Kudo I'm willing to entertain the idea (or patches...) to support Zchunk directly in libmodulemd until librpmio can do it. I don't think it's worthwhile for me to maintain libsquishy to support all the others that rpmio can do for me.

Conan-Kudo:
Well, it's better than nothing, I guess.

sgallagher:
@Conan-Kudo Thanks. I'll reach out to @jdieter and see if I can convince him to help me support the zchunk format in libmodulemd.

PR #363, opened 2019-09-03
Add ModuleIndex.update_from_defaults_directory()

sgallagher:
This function can be used by pungi and MBS to generate a buildroot for modular and
non-modular builds. It will read all files suffixed with .yaml from the provided
directory and add them to the index. If an overrides_path is provided, those
contents will also be merged in (and preferred in the event of a conflict).

### DIFF
--- a/modulemd/include/modulemd-2.0/modulemd-module-index.h
+++ b/modulemd/include/modulemd-2.0/modulemd-module-index.h
@@ -247,6 +247,42 @@ modulemd_module_index_update_from_custom (ModulemdModuleIndex *self,
 
 
 /**
+ * modulemd_module_index_update_from_defaults_directory:
+ * @self: This #ModulemdModuleIndex object.
+ * @path: (in): The path to a directory containing defaults documents.
+ * @strict: (in): Whether the parser should return failure if it encounters an
+ * unknown mapping key or if it should ignore it.
+ * @overrides_path: (in) (nullable): If non-NULL, the path to a directory
+ * containing defaults documents that should override those in @path.
+ * @error: (out): A #GError indicating why this function failed.
+ *
+ * This function will open the directory at @path and iterate through it,
+ * adding any files with the suffix ".yaml" to @self. If @overrides_path is
+ * non-NULL, it will also open any files with the suffix ".yaml" in that
+ * location and import them, overriding any conflicts with the existing
+ * defaults.
+ *
+ * Note: If you need detailed information about what failed and why, it is
+ * better to implement the directory traversal yourself and use the
+ * modulemd_module_index_update_from_file() function, as it will return the
+ * failures information.
+ *
+ * Returns: TRUE if all ".yaml" files in the directory were imported
+ * successfully (this includes if no ".yaml" files were present). FALSE if one
+ * or more files could not be read successfully and sets @error appropriately.
+ *
+ * Since: 2.8
+ */
+gboolean
+modulemd_module_index_update_from_defaults_directory (
+  ModulemdModuleIndex *self,
+  const gchar *path,
+  gboolean strict,
+  const gchar *overrides_path,
+  GError **error);
+
+
+/**
  * modulemd_module_index_dump_to_string:
  * @self: This #ModulemdModuleIndex object.
  * @error: (out): A #GError containing the reason the function failed, NULL if
--- a/modulemd/modulemd-module-index.c
+++ b/modulemd/modulemd-module-index.c
@@ -39,6 +39,10 @@
 #include "private/modulemd-util.h"
 #include "private/modulemd-yaml.h"
 
+
+#define MMD_YAML_SUFFIX ".yaml"
+
+
 struct _ModulemdModuleIndex
 {
   GObject parent_instance;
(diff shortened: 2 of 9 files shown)


modulemd/include/modulemd-2.0/modulemd-module-index.h
```diff
@@ -246,6 +246,40 @@ modulemd_module_index_update_from_custom (ModulemdModuleIndex *self,
...
+ * or more files could not be read successfully.
+ *
+ * Since: 2.8
+ */
+gboolean
+modulemd_module_index_update_from_defaults_directory (
+  ModulemdModuleIndex *self,
+  const gchar *path,
+  gboolean strict,
+  const gchar *overrides_subdir,
```
mmathesius: What is the reasoning behind forcing overrides to be in a subdirectory of the provided `path`? Wouldn't it be more flexible to require the caller to provide an absolute path that could be anywhere? Or, perhaps, if the overrides began with `/` treat it as absolute, and anything else would be relative to `path`.

modulemd/modulemd-module-index.c
```diff
@@ -667,6 +671,83 @@ modulemd_module_index_update_from_custom (ModulemdModuleIndex *self,
...
+          return FALSE;
+        }
+
+      override_idx = modulemd_module_index_new ();
+      while ((filename = g_dir_read_name (overridedir)) != NULL)
+        {
+          if (g_str_has_suffix (filename, MMD_YAML_SUFFIX))
+            {
+              filepath = g_build_path ("/", dirpath, filename, NULL);
+              g_debug ("Reading defaults from %s", filepath);
```
mmathesius:
```suggestion
              g_debug ("Reading default overrides from %s", filepath);
```
Or similar.

modulemd/tests/test-modulemd-moduleindex.c
```diff
@@ -1186,6 +1186,116 @@ test_module_index_read_compressed (void)
...
+  const gchar *names[] = { "meson", "ninja", "nodejs", NULL };
+  for (i = 0; names[i]; i++)
+    {
+      for (j = 0; module_names[j]; j++)
+        {
+          if (g_str_equal (names[i], module_names[j]))
+            break;
+        }
+      g_assert_cmpint (j, <=, name_len);
+    }
```
mmathesius: Wouldn't it be much simpler to do the above using `g_strv_equal()`? `modulemd_module_index_get_module_names_as_strv()` already returns the names sorted.

modulemd/include/modulemd-2.0/modulemd-module-index.h
sgallagher: I was basing it around our expected usage, but I suppose you're right: it would be more sensible to make the `overrides_subdir` be a full path. Making it relative to `path` seems like unnecessary complexity. If a relative path is provided, it should be relative to `$CWD` to avoid confusion.

modulemd/modulemd-module-index.c
sgallagher: Good idea

modulemd/tests/test-modulemd-moduleindex.c
sgallagher: Of course it would. I just completely forgot that `modulemd_module_index_get_module_names_as_strv()` guarantees the sorting. Thanks, that'll simplify things quite a bit.

I had written the Python test first and then basically copied it to the C tests. That was my ham-handed attempt at recreating the `in` keyword...

sgallagher: Hmm, actually I think I'm going to use `g_strv_contains()` instead, because I'd like it to be more visible. But that's still quite a lot cleaner to read than the current implementation.

sgallagher:
Made the requested edits and force-pushed.

modulemd/include/modulemd-2.0/modulemd-module-index.h
```diff
@@ -246,6 +246,40 @@ modulemd_module_index_update_from_custom (ModulemdModuleIndex *self,
...
+ * @self: This #ModulemdModuleIndex object.
+ * @path: (in): The path to a directory containing defaults documents.
+ * @strict: (in): Whether the parser should return failure if it encounters an
+ * unknown mapping key or if it should ignore it.
+ * @overrides_path: (in) (nullable): If non-NULL, the path to a directory
+ * containing defaults documents that should override those in @path.
+ *
+ * This function will open the directory at @path and iterate through it,
+ * adding any files with the suffix ".yaml" to @self. If @overrides_subdir is
+ * non-NULL, it will also open the named subdirectory of @path and include its
```
mmathesius: The above two lines also need adjusted for the @overrides_path change.

modulemd/tests/test-modulemd-moduleindex.c
```diff
@@ -1186,6 +1186,100 @@ test_module_index_read_compressed (void)
...
+   * - meson
+   * - ninja
+   * - nodejs
+   */
+  module_names = modulemd_module_index_get_module_names_as_strv (idx);
+  g_assert_nonnull (module_names);
+
+  g_assert_cmpint (g_strv_length (module_names), ==, 3);
+
+  g_strv_contains ((const gchar *const *)module_names, "meson");
```
mmathesius:
```suggestion
  g_assert_true (g_strv_contains ((const gchar *const *)module_names, "meson"));
```
And the next two lines...

modulemd/tests/test-modulemd-moduleindex.c
```diff
@@ -1186,6 +1186,100 @@ test_module_index_read_compressed (void)
...
+   * - ninja
+   * - nodejs
+   * - testmodule
+   */
+  module_names = modulemd_module_index_get_module_names_as_strv (idx);
+  g_assert_nonnull (module_names);
+
+  g_assert_cmpint (g_strv_length (module_names), ==, 4);
+
+  g_strv_contains ((const gchar *const *)module_names, "meson");
```
mmathesius: As above, wrap these four lines `g_assert_true ()`.

sgallagher:
Thanks for catching those. Fixed and pushed again.

modulemd/include/modulemd-2.0/modulemd-module-index.h
```diff
@@ -246,6 +246,41 @@ modulemd_module_index_update_from_custom (ModulemdModuleIndex *self,
...
 
+/**
+ * modulemd_module_index_update_from_defaults_directory:
+ * @self: This #ModulemdModuleIndex object.
+ * @path: (in): The path to a directory containing defaults documents.
+ * @strict: (in): Whether the parser should return failure if it encounters an
+ * unknown mapping key or if it should ignore it.
+ * @overrides_path: (in) (nullable): If non-NULL, the path to a directory
+ * containing defaults documents that should override those in @path.
+ *
```
mmathesius: Sorry for not noticing this until now, but parameter @error isn't listed above.

sgallagher: Thanks, fixed.

Issue #366, opened 2019-09-18
update_from_defaults_directory does not check for conflicting streams


lubomir:
If I have a directory with two defaults files that are conflicting (same module name, different default stream), this function will load the files and not produce any error even with `strict=True`.

This would be handy in Pungi. There was [a change](https://github.com/fedora-modularity/libmodulemd/commit/3087e4a5c38a331041fec9b6b8f1a372f9ffe64d) to make this error reporting possible, but with this new API for loading defaults it is not usable.

sgallagher:
@lubomir Do you think it makes sense to have the `strict` variable control this or would you prefer if I created a `Modulemd.update_from_defaults_directory_ext()` function with a `strict_default_streams` argument like `Modulemd.ModuleIndexMerger.resolve_ext()` has?

lubomir:
I think just reusing `strict` would be sufficient.

sgallagher:
@lubomir OK, I can do that. Should have something today or tomorrow.

lubomir:
No pressure, I can work around this very easily.

sgallagher:
@lubomir I just pushed the patch for this into all released branches in Fedora: https://bodhi.fedoraproject.org/updates/?like=libmodulemd

Please test and give karma.

PR #367, opened 2019-09-18
Improvements to ModuleIndex.update_from_defaults_dir()

sgallagher:
- Import each file in the directory as a merge rather than an an overwrite so we
can detect conflicts.
- Modify the meaning of the 'strict' argument to mean that it should fail if the
merge would result in a conflict in the default stream setting for a module.

Fixes: https://github.com/fedora-modularity/libmodulemd/issues/366

Attn: @lubomir


sgallagher:
It could be argued that this is a backwards-incompatible ABI change, since some directories that would have been read successfully before will throw an error now, but I think this behavior is what people would have expected from the function to begin with.

sgallagher:
@mmathesius Thanks for the review, updated and force-pushed.

Issue #368, opened 2019-09-20
[RFE] More lenient defaults merging


sgallagher:
The current behavior of the defaults merging logic is located at https://fedora-modularity.github.io/libmodulemd/latest/modulemd-2.0-Modulemd.ModuleIndexMerger.html

In particular, this part of the logic: "When merging ModuleIndexes, if two or more indexes contain Defaults for the same module, but different modified values, the one with the highest modified value will be used and the others discarded."

I realized today that this is going to be problematic for EPEL, because it will mean that EPEL will interact negatively with the RHEL defaults. If EPEL provides a defaults object (such as to set default profiles for streams present in EPEL), it may override the RHEL defaults object entirely (or, if RHEL is newer, the EPEL default profiles will be dropped).

I propose to change the merge logic to be such that instead of the `modified` field resulting in it being an override, it should instead act only in the event of conflict resolution. So if there are two objects whose merge would otherwise conflict in one or more places, the one with the higher version for that item would be used, rather than using the entire object.

@contyk @mmathesius Can you give me your opinion on this? It's arguably an incompatible change, but without it we can't have third-party repositories providing defaults at all, which I think is pretty fundamental.

contyk:
Yes, that makes sense to me. +1

mmathesius:
Agreed. (Although it is strictly an incompatible change, it's a new situation that wasn't previously considered.)

PR #372, opened 2019-10-01
Rework defaults merging logic

sgallagher:
Updates documentation of ModuleIndexMerger with the new algorithm.

Also fixes a memory leak in DefaultsV1.set_default_stream()

Fixes: https://github.com/fedora-modularity/libmodulemd/issues/368

This is the 2.x implementation of the new merge logic. I will be sending the 1.x implementation in a separate PR for the 1.x-maint branch once this is reviewed and merged.

### DIFF
--- a/modulemd/modulemd-defaults-v1.c
+++ b/modulemd/modulemd-defaults-v1.c
@@ -205,6 +205,7 @@ modulemd_defaults_v1_set_default_stream (ModulemdDefaultsV1 *self,
       else
         {
           /* This is the fallback default for non-specific intents */
+          g_clear_pointer (&self->default_stream, g_free);
           self->default_stream = g_strdup (default_stream);
         }
     }
@@ -1193,17 +1194,18 @@ modulemd_defaults_v1_emit_intents (ModulemdDefaultsV1 *self,
 }
 
 static gboolean
-modulemd_defaults_v1_merge_intent_profiles (
+modulemd_defaults_v1_merge_default_profiles (
   GHashTable *from_profile_defaults,
   GHashTable *merged_profile_defaults,
+  guint64 from_modified,
+  guint64 into_modified,
   GError **error);
 static GHashTable *
 modulemd_defaults_v1_copy_intent_profiles (GHashTable *intent_profiles);
 
 
 ModulemdDefaults *
-modulemd_defaults_v1_merge (const gchar *module_name,
-                            ModulemdDefaultsV1 *from,
+modulemd_defaults_v1_merge (ModulemdDefaultsV1 *from,
                             ModulemdDefaultsV1 *into,
                             gboolean strict_default_streams,
                             GError **error)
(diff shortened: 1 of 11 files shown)


modulemd/modulemd-defaults-v1.c
```diff
@@ -1213,60 +1215,69 @@ modulemd_defaults_v1_merge (const gchar *module_name,
...
+    MODULEMD_DEFAULTS_V1 (modulemd_defaults_copy (MODULEMD_DEFAULTS (into)));
 
   /* Merge the default streams */
-  if (into->default_stream && !from->default_stream)
-    {
-      modulemd_defaults_v1_set_default_stream (
-        merged, into->default_stream, NULL);
-    }
-  else if (from->default_stream && !into->default_stream)
+  if (from->default_stream && !into->default_stream)
```
mmathesius:
```suggestion
  if (from->default_stream && !merged->default_stream)
```

Since you seemed to be deliberately changing the code to refer to `merged` (as a copy of `into`) throughout the block below, it should be changed here as well for consistency.

modulemd/modulemd-defaults-v1.c
```diff
@@ -1213,60 +1215,69 @@ modulemd_defaults_v1_merge (const gchar *module_name,
...
+                      from->default_stream);
+              if (strict_default_streams)
+                {
+                  g_set_error (
+                    error,
+                    MODULEMD_ERROR,
+                    MODULEMD_ERROR_VALIDATE,
+                    "Default stream mismatch in module %s: %s != %s",
+                    module_name,
+                    into->default_stream,
```
mmathesius:
```suggestion
                    merged->default_stream,
```

For consistency...

modulemd/modulemd-defaults-v1.c
```diff
@@ -1213,60 +1215,69 @@ modulemd_defaults_v1_merge (const gchar *module_name,
...
-                           from->default_stream);
-              return NULL;
+              modulemd_defaults_v1_set_default_stream (
+                merged, from->default_stream, NULL);
+            }
+          else if (from_modified == into_modified)
+            {
+              /* They have conflicting default streams */
+              g_info ("Module stream mismatch in merge: %s != %s",
+                      into->default_stream,
```
mmathesius:
```suggestion
                      merged->default_stream,
```
For consistency...

modulemd/modulemd-defaults-v1.c
```diff
@@ -1322,45 +1316,43 @@ modulemd_defaults_v1_merge (const gchar *module_name,
       merged_default_stream =
         g_hash_table_lookup (merged->intent_default_streams, intent_name);
 
-      if (!merged_default_stream)
+      /* If there is no new default stream, just jump to the next item */
+      if (!intent_default_stream)
+        continue;
+
+      else if (!merged_default_stream && intent_default_stream)
```
mmathesius:
```suggestion
      else if (!merged_default_stream)
```
The `if` above already made sure `intent_default_stream` isn't NULL.

modulemd/modulemd-defaults-v1.c
```diff
@@ -1322,45 +1316,43 @@ modulemd_defaults_v1_merge (const gchar *module_name,
...
-
-      intent_profile_defaults =
-        modulemd_defaults_v1_copy_intent_profiles (intent_profiles);
-
-      g_hash_table_insert (merged->intent_default_profiles,
-                           g_strdup (intent_name),
-                           g_steal_pointer (&intent_profile_defaults));
-    }
-
   /* Now copy 'from" into merged, checking for conflicts */
```
mmathesius:
```suggestion
   /* Now copy 'from' into merged, checking for conflicts */
```
That was an interesting mixture of quotes!

modulemd/modulemd-defaults-v1.c
```diff
@@ -1213,60 +1215,69 @@ modulemd_defaults_v1_merge (const gchar *module_name,
...
         }
       else
         {
-          /* They're the same, so store that */
-          modulemd_defaults_v1_set_default_stream (
-            merged, into->default_stream, NULL);
+          /* They're the same, so change nothing */
         }
     }
   else
```
mmathesius: The comment below should be changed. If it's reached, it only means that the "from" default stream value is NULL, so there is nothing to do.

modulemd/modulemd-defaults-v1.c
```diff
@@ -1375,7 +1367,7 @@ modulemd_defaults_v1_merge (const gchar *module_name,
           intent_profile_defaults =
             modulemd_defaults_v1_copy_intent_profiles (intent_profiles);
 
-          /* This wasn't in 'into', so just add it */
+          /* This wasn't in 'merged', so just add it */
```
mmathesius: Nice catch!

modulemd/modulemd-defaults-v1.c
sgallagher: This one is actually correct. It's a proof against future code refactoring resulting in `merged->default_stream` not matching `into->default_stream` anymore. We want to be sure to log what the inputs were that failed.

sgallagher:
Amended and pushed.

modulemd/modulemd-defaults-v1.c
mmathesius: Gotcha. I just find it bothersome that this is within a conditional block checking for equality against `merged->default_stream` while the error messages log `into->default_stream`.

PR #374, opened 2019-10-03
Refactor common test routines

sgallagher:
Additional cleanups:
* Handle cleanup with a bash trap.
* Drop an unneeded ARG from the deps Dockerfiles
* Drop unneeded JOB_NAME parsing from travis-tasks.sh


.travis/travis-mageia.sh
```diff
@@ -2,48 +2,20 @@
...
-set -e
 
+source $SCRIPT_DIR/travis-common.inc
+
+set -e
 set -x
 
 
-JOB_NAME=${TRAVIS_JOB_NAME:- Mageia 7}
+JOB_NAME=${TRAVIS_JOB_NAME:-Mageia 7}
```
mmathesius: Thank you. That extra space _was_ bothering me.

.travis/travis-common.inc
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+set -x
+
+# Global variables to support finalize()
+MMD_OS=
+MMD_RELEASE=
+MMD_TARBALL_PATH=
+
```
mmathesius: Optional: consider adding `: ${SCRIPT_DIR:=$(pwd)}` in here, or throw an error if `SCRIPT_DIR` isn't set, since things heavily depend upon it.

.travis/travis-common.inc
```diff
@@ -0,0 +1,68 @@
...
+    return $exitcode
+}
+
+trap common_finalize EXIT
+
+
+function mmd_run_docker_tests {
+    local os release repository image
+    local "${@}"
```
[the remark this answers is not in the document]
sgallagher: I hadn't either. I stumbled upon it yesterday on a [stackoverflow](https://stackoverflow.com/questions/12128296/is-there-a-way-to-avoid-positional-arguments-in-bash) post and decided it was really nice.

.travis/travis-common.inc
sgallagher: Hmm, good point. A lot of stuff could break in that case. I'll add it inside the function call, though. I want it to be possible to source the `travis-common.inc` before `$SCRIPT_DIR` is set. It just needs to fail if it's used without it.

mmathesius:
The additional fixups look good. Thanks!

PR #375, opened 2019-10-03
Made tests non-mandatory

KOLANICH:


KOLANICH:
Not tests, but docs, of course. The same mistake, but with tests, is in another package.

sgallagher:
@KOLANICH Would you mind force-pushing with a corrected commit message before I merge it, please?

PR #377, opened 2019-10-04
Make doc-build configurable

sgallagher:

@KOLANICH


meson.build
```diff
@@ -46,12 +46,17 @@ foreach cflag: test_cflags
...
 gobject = dependency('gobject-2.0')
 yaml = dependency('yaml-0.1')
-gtkdoc = dependency('gtk-doc')
+
+if with_docs
+    gtkdoc = dependency('gtk-doc')
+endif
 
 glib_prefix = dependency('glib-2.0').get_pkgconfig_variable('prefix')
 glib_docpath = join_paths(glib_prefix, 'share', 'gtk-doc', 'html')
```
mmathesius: Should these lines and the tests for the presence of GLib and GObject documentation (a few lines below) also be included within an `if with_docs` block?

sgallagher: Yes, they should. Fixing.

meson.build
```diff
@@ -46,28 +46,34 @@ foreach cflag: test_cflags
   endif
 endforeach
 
+with_docs = get_option('with_docs')
```
mmathesius: Totally optional: given the way things lay out now, it would make more sense to put this line right before the `if with_docs` block.

sgallagher: Done

mmathesius:
LTGM!

KOLANICH:
Thank you!

PR #380, opened 2019-10-11
Use more comprehensive auto-formatting tools for Python and C

sgallagher:
These patches switch our tests from running `autopep8` to running [`black`](https://black.readthedocs.io) instead. This autoformatter includes some niceties as enforcing the same style of quotes everywhere and does a better job of aligning on the column boundaries.

I separated the patch that updates all of the syntax from the patch that adds the auto-formatter to make review simpler.

I also noticed that `.pyc` and `.pyo` files weren't included in .gitignore while working on this, so there's a patch for that as well.

### DIFF
--- a/modulemd/include/private/modulemd-subdocument-info-private.h
+++ b/modulemd/include/private/modulemd-subdocument-info-private.h
@@ -103,13 +103,13 @@ modulemd_subdocument_info_get_mdversion (ModulemdSubdocumentInfo *self);
 /**
  * modulemd_subdocument_info_set_yaml:
  * @self: This #ModulemdSubdocumentInfo object.
- * @contents: The contents of the document.
+ * @yaml: The contents of the document.
  *
  * Since: 2.0
  */
 void
 modulemd_subdocument_info_set_yaml (ModulemdSubdocumentInfo *self,
-                                    const gchar *contents);
+                                    const gchar *yaml);
 
 
 /**
--- a/modulemd/meson.build
+++ b/modulemd/meson.build
@@ -16,9 +16,15 @@ test_dirty_git = get_option('test_dirty_git')
 test_installed_lib = get_option('test_installed_lib')
 skip_introspection = get_option('skip_introspection')
 
-clang_format = find_program('clang-format', required: developer_build)
-if not clang_format.found()
+clang_tidy = find_program('clang-tidy', required: false)
+if clang_tidy.found()
     clang_format = disabler()
+else
+    clang_tidy = disabler()
+    clang_format = find_program('clang-format', required: developer_build)
+    if not clang_format.found()
+        clang_format = disabler()
+    endif
 endif
 
 python_formatter = find_program('black', required : developer_build)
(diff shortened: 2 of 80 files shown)


sgallagher:
I just pushed an additional couple of patches to add `clang-tidy` on systems that support it to do a more thorough code cleanup of C files and headers.

sgallagher:
That, and figuring out why the tests are failing in Travis but working when I run it locally...

modulemd/include/private/modulemd-subdocument-info-private.h
```diff
@@ -109,7 +109,7 @@ modulemd_subdocument_info_get_mdversion (ModulemdSubdocumentInfo *self);
  */
```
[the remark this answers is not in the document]
sgallagher: Interesting; looks like those didn't match between the `.h` and `.c` files. The formatter made the prototype match the implementation.

mmathesius:
I could have sworn I saw Travis succeeded--but I must have had my PRs crossed.

modulemd/meson.build
```diff
@@ -381,35 +386,46 @@ endforeach
...
-            modulemd_srcs +
-            modulemd_hdrs +
-            modulemd_priv_hdrs +
-            modulemd_validator_srcs +
-            test_srcs +
-            test_priv_hdrs)
+     args : clang_args + clang_files)
+
+clang_tidy_args = [ '-p', meson.build_root(),
+                    '--checks=*,-llvm-header-guard,-bugprone-macro-parentheses',
```
mmathesius:
```suggestion
                    '--checks=*,-llvm-header-guard,-bugprone-macro-parentheses,-misc-unused-parameters',
```
Upon doing some investigation, it looks like this change should help with the travis tests. Without it, `clang-tidy` is making some undesirable changes.

sgallagher: Would you mind giving me an example of such a case where an undesirable change is being made?

mmathesius: Exactly what seems to have caused the latest Fedora 29 travis run to fail, such as can be seen at the end of the raw log file: https://api.travis-ci.org/v3/job/596767803/log.txt

When I run the clang_tidy test locally without disabling the misc-unused-parameters check, I get dozens of undesirable changes along the lines of:

```
--- a/modulemd/modulemd-defaults-v1.c
+++ b/modulemd/modulemd-defaults-v1.c
@@ -461,7 +461,7 @@ modulemd_defaults_v1_get_default_profiles_for_stream_as_strv (
 static void
 modulemd_defaults_v1_get_property (GObject *object,
                                    guint prop_id,
-                                   GValue *value,
+                                   GValue * /*value*/,
                                    GParamSpec *pspec)
 {
   switch (prop_id)
```

sgallagher: So, it seems that clang-tidy 7.0.0 has a bug with that particular function. The 9.0.0 I have on my F31 system does not do that. I've modified the meson build to just require at least 9.0.0 for clang-tidy or else fall back to clang-format. That should make life more bearable. I'm going to cancel this PR and push a new branch (atop my #383 PR, since it fixes a bug that the clang-tidy stuff revealed).

sgallagher:
Obsoleted by PR #384

Issue #382, opened 2019-10-17
Modulemd.ModuleStream.read_*() doesn't honor name overrides


sgallagher:
The `Modulemd.ModuleStream.read_*()` functions have optional fields `module_name` and `module_stream` that are intended to override the names read from the file (or not read, if they were unspecified on disk). This is mostly intended for use with packager modulemd documents that should get their name from the git directory and their stream name from the git branch.

However, we never actually implemented that override. I noticed this when the work I was doing on #380 caused those two fields to be removed from some internal functions, tipping me off that they weren't actually being used.

PR #383, opened 2019-10-17
ModuleStream: Properly override names when asked to

sgallagher:
Fixes: https://github.com/fedora-modularity/libmodulemd/issues/382

This PR also includes two other minor fixes, a missing error return (that shouldn't be reachable, but better safe than sorry) and moves the validation of the ModuleStream YAML constructors to the common parent.

### DIFF
--- a/modulemd/modulemd-module-stream.c
+++ b/modulemd/modulemd-module-stream.c
@@ -278,6 +278,13 @@ modulemd_module_stream_read_yaml (yaml_parser_t *parser,
     }
   yaml_event_delete (&event);
 
+
+  if (!modulemd_module_stream_validate (stream, &nested_error))
+    {
+      g_propagate_error (error, g_steal_pointer (&nested_error));
+      return NULL;
+    }
+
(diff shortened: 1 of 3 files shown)


modulemd/modulemd-module-stream.c
```diff
@@ -254,6 +254,7 @@ modulemd_module_stream_read_yaml (yaml_parser_t *parser,
                    MODULEMD_YAML_ERROR_PARSE,
                    "Unknown ModuleStream version: %" PRIu64,
                    modulemd_subdocument_info_get_mdversion (subdoc));
+      return NULL;
       break;
```
mmathesius: Good catch!

But what are your thoughts on removing this `break` immediately following the `return`? I'm almost surprised the clang auto-formatting tools don't remove it or warn that it's unreachable code.

sgallagher: I'd prefer to leave it in. A missing `break` can cause all sorts of havoc on a refactor. Better to have it and not need it than to need it and not have it.

PR #385, opened 2019-10-18
Travis: Cache deps for packagers, add F31 test target, and run tests on aarch64

sgallagher:
With the recent [patch](https://github.com/fedora-modularity/libmodulemd/commit/c42bb7266f56b59cc31bda9a1c6cf692d1b78355) to enable building with `buildah` and `podman`, local runs of the test suite using the OCI tools was no longer caching the `deps` intermediate container, so test runs were much longer than they needed to be. This is because `buildah` defaults to not caching. This patch adds the `--layers=true` flag for `buildah` so that it will cache the layers and vastly speed up local travis runs.


sgallagher:
The latest version of this PR now also includes running all Fedora and CentOS tests on both x86_64 and aarch64 platforms. I reordered some of the tests so that the documentation is built first and then CentOS builds are run which will serve as a canary for whether the other ones are likely to succeed. This way a contributor doesn't necessarily have to wait for all of the checks to complete before identifying that there is an issue.

.travis.yml
```diff
@@ -27,21 +27,59 @@ notifications:
...
+      script: ./.travis/travis-fedora.sh
+      arch: arm64
+    - name: "Fedora rawhide"
+      script: ./.travis/travis-fedora.sh
+      arch: arm64
+
+
     - name: "Coverity Scan"
       script: 'if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then ./.travis/travis-coverity.sh; fi'
+
```
mmathesius: Should the Coverity scan get an explicit `arch: amd64`, too?

sgallagher: Yes, I missed that.

.travis.yml
```diff
@@ -27,21 +27,60 @@ notifications:
...
+      script: 'if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then ./.travis/travis-docs.sh; fi'
+      arch: amd64
+
+
+    - name: "CentOS 7"
+      script: ./.travis/travis-centos.sh
+      arch: amd64
+
+
     - name: "Fedora 29"
```
mmathesius: One more suggestion... When looking at the travis results summary, it would be good to know which builds were x86_64 vs. aarch64. You should be able to add the platform to the Fedora job names (eg., "Fedora 29 x86_64" here and ""Fedora 29 aarch64" below). The `travis-fedora.sh` script already does the right thing by setting the release to `${arr[1]}` and ignoring other tokens in the job name.

sgallagher: Interesting; the Travis view when you aren't signed in doesn't show the architecture field in the build overview. It shows up for me while I'm signed in.

mmathesius: Yes, interesting. With the build stages you just added (:+1:, btw), the architecture field now shows up without being signed in--in addition to appearing in the job name.

mmathesius:
Looks good to me. It's unfortunate that `clang-tidy` has such an urge to segfault as to require temporary disablement.

Approved--pending successful completion of the remaining CI.

Issue #387, opened 2019-11-06
Could not execute command "/builddir/build/BUILD/modulemd-2.8.2/modulemd/clang_simple_version.sh /usr/bin/clang-tidy


tpgxyz:
Hi,

i'm trying to build modulemd-1.8.2 with LLVM/clang-9.0.1 and i got this error. Full build log can be found here https://abf.openmandriva.org/build_lists/640727

```
Configuring make_rpms.sh using configuration
Program clang_simple_version.sh found: YES (/builddir/build/BUILD/modulemd-2.8.2/modulemd/clang_simple_version.sh)
Program clang-tidy found: YES (/usr/bin/clang-tidy)
Running command: /builddir/build/BUILD/modulemd-2.8.2/modulemd/clang_simple_version.sh /usr/bin/clang-tidy
modulemd/meson.build:37:8: ERROR: Could not execute command "/builddir/build/BUILD/modulemd-2.8.2/modulemd/clang_simple_version.sh /usr/bin/clang-tidy".
```

Looks like problem lies here:
https://github.com/fedora-modularity/libmodulemd/blob/89031dbcc98fee3696c51b48f8e665ade4695d05/modulemd/clang_simple_version.sh#L1

tpgxyz:
fixed

PR #388, opened 2019-11-06
fix interpreter path

tpgxyz:
This fixes issue reported here https://github.com/fedora-modularity/libmodulemd/issues/387

### DIFF
--- a/modulemd/clang_simple_version.sh
+++ b/modulemd/clang_simple_version.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/sh
+#!/bin/sh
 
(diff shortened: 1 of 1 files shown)


sgallagher:
This looks fine to me and I'll merge it once the CI tests are complete.

Would you be willing to contribute a CI test for OpenMandriva so I can catch issues like this upstream in the future? See https://github.com/fedora-modularity/libmodulemd/tree/master/.travis/mageia and https://github.com/fedora-modularity/libmodulemd/blob/master/.travis/travis-mageia.sh for an example.

tpgxyz:
@sgallagher sure will do this

PR #389, opened 2019-11-08
add OpenMandriva travis test

tpgxyz:


.travis/travis-openmandriva.sh
```diff
@@ -0,0 +1,20 @@
...
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+pushd $SCRIPT_DIR
+
+source $SCRIPT_DIR/travis-common.inc
+
+set -e
+set -x
+
+JOB_NAME=${TRAVIS_JOB_NAME:-OpenMandriva Lx cooker }
```
sgallagher:
```suggestion
JOB_NAME=${TRAVIS_JOB_NAME:-OpenMandriva cooker }
```

I'm not sure if that was a typo, but the code value below expects the second word to be the release value.

.travis/travis-openmandriva.sh
```diff
@@ -0,0 +1,20 @@
...
+
+arr=($JOB_NAME)
+release=${arr[1]:-cooker}
+
+mmd_run_docker_tests \
+    os=openmandriva \
```
sgallagher: Take a look at the [opensuse](https://github.com/fedora-modularity/libmodulemd/blob/master/.travis/travis-opensuse.sh#L20) test.

When the image name has a forward slash in it, you need to add `image=openmandriva/$release` to the call and it should work.

.travis/travis-openmandriva.sh
tpgxyz: Fixed

.travis/travis-openmandriva.sh
```diff
@@ -0,0 +1,20 @@
...
+
+JOB_NAME=${TRAVIS_JOB_NAME:-OpenMandriva cooker}
+
+arr=($JOB_NAME)
+release=${arr[1]:-cooker}
+
+mmd_run_docker_tests \
+    os=openmandriva \
+    release=$release \
+    repository=hub.docker.com/r \
```
sgallagher: This address gives a 404. Are you sure you don't mean
```suggestion
    repository=docker.io \
```

Because that pulls an image successfully.

tpgxyz: There is something strange here as OpenMandriva docker's images lives on hub.docker.com/r/openmandriva previously I've set docker.io and it did not work, while docker pull openmandriva/cooker works

sgallagher:
OK, the latest failure seems to be because we're running the tests on an Ubuntu 14.04 test runner. I'll push an update to switch the test runner to 16.04 and see if that passes.

I submitted it as PR #390 

Good news: It successfully boots the container now.
Bad news: Still getting errors in that container:
```
determining the fastest mirror (13 hosts).. done.
OpenMandriva Cooker - x86_64                    6.4 MB/s |  11 MB     00:01    
OpenMandriva cooker - x86_64                    5.6 MB/s |  11 MB     00:02    
OpenMandriva cooker - x86_64 - Updates           14  B/s | 153  B     00:10    
Failed to synchronize cache for repo 'updates-x86_64'
Ignoring repositories: updates-x86_64
Last metadata expiration check: 0:00:04 ago on Thu Nov 14 13:04:48 2019.
No match for argument: file-devel
No match for argument: glib2-devel
No match for argument: /usr/share/gtk-doc/html/glib/index.html
No match for argument: libyaml-devel
Package openssl-1.1.1c-1.x86_64 is already installed.
No match for argument: python3-autopep8
No match for argument: python3-gitpython
No match for argument: python3-gobject3
No match for argument: python3-pycodestyle
No match for argument: rpm-mageia-setup-build
Package wget-1.20.3-2.x86_64 is already installed.
Error: Unable to find a match: file-devel glib2-devel /usr/share/gtk-doc/html/glib/index.html libyaml-devel python3-autopep8 python3-gitpython python3-gobject3 python3-pycodestyle rpm-mageia-setup-build
The command '/bin/sh -c dnf -y --setopt=install_weak_deps=False --setopt=tsflags='' install 	clang 	createrepo_c 	curl 	file-devel 	git-core 	glib2-devel 	/usr/share/gtk-doc/html/glib/index.html 	gobject-introspection-devel 	gtk-doc 	libyaml-devel 	cmake 	meson 	ninja 	openssl 	pkgconf 	python2-six 	python3-autopep8 	python3-devel 	python3-gitpython 	python3-gobject3 	python3-pycodestyle 	rpm-mageia-setup-build 	rpm-build 	rpm-devel 	rpmdevtools 	ruby 	"rubygem(json)" 	rubygems 	sudo 	valgrind 	wget     && dnf -y clean all' returned a non-zero code: 1
+common_finalize
+exitcode=1
```

tpgxyz:
> OK, the latest failure seems to be because we're running the tests on an Ubuntu 14.04 test runner. I'll push an update to switch the test runner to 16.04 and see if that passes.
> 
Well that issues came from *older* kernel inside that Ubuntu. OpenMandriva's glibc does not support kernels older than 4.14 on ix86

List of dnf packages will be fixed this week, as i do not have much time these days :(

sgallagher:
I just pushed a fix to move to the newer Ubuntu test host.

tpgxyz:
@sgallagher Looks like things are going into good way. LLVM/clang is used on building. Now travis-ci times out on tests. Probably because of `-Dskip_clang_tidy=false` which i've set by default.

sgallagher:
@tpgxyz It's more likely to be the Valgrind tests taking too long. Try setting `MMD_SKIP_VALGRIND=True` before calling `ninja test` and see if that works.

sgallagher:
You either need to include GitPython or drop `-Dtest_dirty_git` in `COMMON_MESON_ARGS` from travis-tasks.sh.

sgallagher:
The `popd` error here looks like it's a problem with my common include file. I'll fix that and push it momentarily, then you can rebase atop it and try again. Sorry about that.

tpgxyz:
@sgallagher well, best option would be if you drop all non POSIX shell comand. Try https://www.shellcheck.net/

sgallagher:
> @sgallagher well, best option would be if you drop all non POSIX shell comand. Try https://www.shellcheck.net/

I'll happily take patches to de-bashify the tools.

I just squashed your patches into a single patch in my https://github.com/fedora-modularity/libmodulemd/pull/395 PR and merged them upstream. Thanks for the contribution!

PR #392, opened 2019-11-27
WIP: Validate artifact list

tbaederr:
Hi,

the artifacts list in the modulemd currently accepts everything, including empty list entries or entries containing slashes, etc. I know this is not being written by humans in the usual workflow but validating the list is still worthwhile for tools that generate the modulemd file.

The commit adds a few rudimentary checks (and I'm not 100% certain all of them are correct). Not sure if they should care more about utf8 though (or if utf8 is even permitted here...).

### DIFF
--- a/modulemd/modulemd-util.c
+++ b/modulemd/modulemd-util.c
@@ -352,12 +352,18 @@ modulemd_validate_nevra (const gchar *nevra)
 
   /* No need to validate Release; it's fairly arbitrary */
 
+  i--;
   /* Process the version */
   while (i >= tmp)
     {
       if (*i == ':')
         {
           break;
+        }
+      else if (*i == '-')
+        {
+          /* '-' between version and epoch is not allowed */
+          return FALSE;
         }
       i--;
     }


sgallagher:
I’m on vacation this week, so I’ll do a full review on Monday. UTF-8 is not permitted in Fedora/RHEL package names at the moment, but it’s not prevented by RPM, so we should handle it.

tbaederr:
Adding them to `modulemd_validate_nevra()` makes sense, but that is not being called at all when running `modulemd-validator`, although there is no way of returning a good error message.

It just creates a `ModulemdModuleIndex` and then calls `modulemd_module_index_update_from_parser()`

sgallagher:
> Adding them to `modulemd_validate_nevra()` makes sense, but that is not being called at all when running `modulemd-validator`, although there is no way of returning a good error message.

You're right. I need to figure out what code path is missing the validation, because it should happen for any case where we read the YAML in. Thanks for spotting that.

tbaederr:
I updated this PR anyway with a much smaller change, does that look good? That would fix the error I ran into.

PR #401, opened 2019-12-10
support to set module build architectures in the modulemd file

### LINKED ISSUE #399: Add the ability to set the architectures to build for in the modulemd file
Currently, there is a way to override the arches per component, but it'd be nice to be able to set the default arches at the module level. In practice, MBS would set these arches on the build tags it creates in Koji.

This should likely be a field under `buildopts`.

mmathesius:
Initial changes to resolve #399.

No validation or tests yet.

### DIFF
--- a/modulemd/modulemd-buildopts.c
+++ b/modulemd/modulemd-buildopts.c
@@ -95,8 +95,7 @@ modulemd_buildopts_copy (ModulemdBuildopts *self)
   modulemd_buildopts_set_rpm_macros (b,
                                      modulemd_buildopts_get_rpm_macros (self));
 
-  g_hash_table_unref (b->whitelist);
-  b->whitelist = g_hash_table_ref (self->whitelist);
+  MODULEMD_REPLACE_SET (b->whitelist, self->whitelist);
 
   return g_steal_pointer (&b);
 }
--- a/modulemd/modulemd-buildopts.c
+++ b/modulemd/modulemd-buildopts.c
@@ -29,6 +29,7 @@ struct _ModulemdBuildopts
   gchar *rpm_macros;
 
   GHashTable *whitelist;
+  GHashTable *arches;
 };
 
 G_DEFINE_TYPE (ModulemdBuildopts, modulemd_buildopts, G_TYPE_OBJECT)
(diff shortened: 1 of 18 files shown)


modulemd/modulemd-buildopts.c
```diff
@@ -87,18 +93,20 @@ modulemd_buildopts_new (void)
...
+  copy = modulemd_buildopts_new ();
 
-  modulemd_buildopts_set_rpm_macros (b,
+  modulemd_buildopts_set_rpm_macros (copy,
                                      modulemd_buildopts_get_rpm_macros (self));
 
-  g_hash_table_unref (b->whitelist);
-  b->whitelist = g_hash_table_ref (self->whitelist);
+  g_hash_table_unref (copy->whitelist);
+  copy->whitelist = g_hash_table_ref (self->whitelist);
```
sgallagher: The original code here is buggy. We need to be making a deep copy of the hash table, not taking a reference on it. Otherwise, editing it in the new copy will modify it in the original copy (and vice-versa).

Mind fixing that for both whitelist and arches while you're here?

modulemd/modulemd-buildopts.c
```diff
@@ -288,6 +331,20 @@ modulemd_buildopts_parse_yaml (yaml_parser_t *parser,
...
+            {
+              g_hash_table_unref (buildopts->arches);
+              buildopts->arches =
+                modulemd_yaml_parse_string_set (parser, &nested_error);
+              if (buildopts->arches == NULL)
+                {
+                  MMD_YAML_ERROR_EVENT_EXIT_BOOL (
+                    error,
+                    event,
+                    "Failed to parse arches list in buildopts rpms: %s",
```
sgallagher:
```suggestion
                    "Failed to parse arches list in buildopts: %s",
```

modulemd/modulemd-buildopts.c
mmathesius: Fixed.

mmathesius:
> Before merging, we're going to want to merge logical changes together. We can't update the spec in a separate patch from the implementation and test fixes or else we break the ability to do a `git bisect` later.

Right. Since this is a work in progress, I was planning to do that later. But I went ahead and did that now. Tests should pass at each commit--except for the OpenMandriva tests that currently seem to be failing.

I've added tests with my latest push, but still no validation.

sgallagher:
Looks good to me so far. Thanks!

mmathesius:
@sgallagher I added validation checks, but something is blowing up in the tests and I haven't figured it out yet. However, I figured I'd push this so you can take a look while I'm trying to sort it out.

Yes, I know I should rebase to the latest as well.

mmathesius:
Found it. If there are no module buildopts.arches specified, then validation of the rpm component arches automatically succeeds. I had added that simple check in the validation function earlier, but it somehow got clobbered...

mmathesius:
Now rebased to the latest master...

sgallagher:
> The validation function should not be public. It should be an internal implementation detail of `modulemd_module_stream_validate()`
> 
> What we want is for `modulemd_module_stream_validate()` to include this in its checks (so we're always running this validation when loading into an Index and before writing out to a file.

I misread which header that was in. It's just fine.

mmathesius:
Now with the memory leak fixed.

mmathesius:
Sorry about the extra noise. That last force push was to line wrap some long comments in the validation tests.

mmathesius:
If only I had actually added that to the commit. Now the comment wrapping is fixed.

PR #403, opened 2019-12-11
Correct typo in MODULEMD_REPLACE_SET()

mmathesius:
It has been working only because every existing use of it has the second argument named `set`...


mmathesius:
Closing this PR and including it as part of #401. But CI ran successfully--until it failed for unrelated reasons.

PR #407, opened 2019-12-14
Add tests for ModulemdModuleStream description

Norem80:
Copied python tests for Modulemd.ModuleStream.description into C as a Google Code-in 2019 Task.

### DIFF
--- a/modulemd/tests/test-modulemd-modulestream.c
+++ b/modulemd/tests/test-modulemd-modulestream.c
@@ -21,6 +21,8 @@
   "À϶￥🌭∮⇒⇔¬β∀₂⌀ıəˈ⍳⍴V)"                           \
   "═€ίζησθლბშიнстемองจึองታሽ።ደለᚢᛞᚦᚹ⠳⠞⠊⠎▉▒▒▓😃"
 
+#define MMD_TEST_DESC_TEXT "A different description"
+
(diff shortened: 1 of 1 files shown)


modulemd/tests/test-modulemd-modulestream.c
```diff
@@ -21,6 +21,9 @@
   "À϶￥🌭∮⇒⇔¬β∀₂⌀ıəˈ⍳⍴V)"                           \
   "═€ίζησθლბშიнстемองจึองታሽ።ደለᚢᛞᚦᚹ⠳⠞⠊⠎▉▒▒▓😃"
 
+#define MMD_TEST_DESC_LOCALE "C"
```
OrionStar25: I don't think this is necessary as "C" refers to the default locale. It's unlikely to change. I personally think the readability is being lost by such a huge macro, but I think we'll let this stay. 

@sgallagher What do you think?

Norem80: So should I just write "C" as an argument rather then create macro?

Norem80:
Added that.

sgallagher:
There are conflicts with the current master branch. Would you mind doing a `git rebase` and resolving them, please?

sgallagher:
@Norem80 Please rebase this pull request.

PR #409, opened 2019-12-14
Copy python tests for Modulemd.ModuleStream.summary into C

Norem80:
Copy python tests for Modulemd.ModuleStream.summary into C as Google Code-in 2019 task.

### DIFF
--- a/modulemd/tests/test-modulemd-modulestream.c
+++ b/modulemd/tests/test-modulemd-modulestream.c
@@ -21,6 +21,8 @@
   "À϶￥🌭∮⇒⇔¬β∀₂⌀ıəˈ⍳⍴V)"                           \
   "═€ίζησθლბშიнстемองจึองታሽ።ደለᚢᛞᚦᚹ⠳⠞⠊⠎▉▒▒▓😃"
 
+#define MMD_TEST_SUM_TEXT "A different summary"
+
 typedef struct _ModuleStreamFixture
 {
 } ModuleStreamFixture;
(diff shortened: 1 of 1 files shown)


modulemd/tests/test-modulemd-modulestream.c
```diff
@@ -205,6 +207,63 @@ module_stream_v2_test_profiles (ModuleStreamFixture *fixture,
...
+  // Test setting it back to NULL
+  modulemd_module_stream_v1_set_summary (stream, NULL);
+  summary = modulemd_module_stream_v1_get_summary (stream, "C");
+  g_assert_null (summary);
+
+  // Test setting unicode characters
+  modulemd_module_stream_v1_set_summary (stream, MMD_TEST_DOC_UNICODE_TEXT);
+  summary = modulemd_module_stream_v1_get_summary (stream, "C");
+  g_assert_cmpstr (summary, ==, MMD_TEST_DOC_UNICODE_TEXT);
+}
```
OrionStar25: You have declared `summary` but you also have to clear them so that it doesn't result in memory leakage. See https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/tests/test-modulemd-modulestream.c#L329

modulemd/tests/test-modulemd-modulestream.c
```diff
@@ -205,6 +207,63 @@ module_stream_v2_test_profiles (ModuleStreamFixture *fixture,
...
+  // Test setting it back to NULL
+  modulemd_module_stream_v2_set_summary (stream, NULL);
+  summary = modulemd_module_stream_v2_get_summary (stream, "C");
+  g_assert_null (summary);
+
+  // Test setting unicode characters
+  modulemd_module_stream_v2_set_summary (stream, MMD_TEST_DOC_UNICODE_TEXT);
+  summary = modulemd_module_stream_v2_get_summary (stream, "C");
+  g_assert_cmpstr (summary, ==, MMD_TEST_DOC_UNICODE_TEXT);
+}
```
OrionStar25: Same as above.

Norem80: There's clear of documentation_prop which is different from documentation which is analogical to summary here I can't clear ```const gchar``` but I surely can clean stream.

OrionStar25:
You're right. Pointers confuse me big time 🙈 
`auto_ptr` has inbuilt garbage collection. That is, it deletes it's instance after the object has escaped its scope. So there's no need to clear summary too. 

Great job!

PR #410, opened 2019-12-15
Copy Python tests for Modulemd.ModuleStream.tracker into C

srikavin:
This is related to #199. I also added Unicode tests.

### DIFF
--- a/modulemd/tests/test-modulemd-modulestream.c
+++ b/modulemd/tests/test-modulemd-modulestream.c
@@ -20,6 +20,7 @@
 #define MMD_TEST_DOC_UNICODE_TEXT                                             \
   "À϶￥🌭∮⇒⇔¬β∀₂⌀ıəˈ⍳⍴V)"                           \
   "═€ίζησθლბშიнстемองจึองታሽ።ደለᚢᛞᚦᚹ⠳⠞⠊⠎▉▒▒▓😃"
+#define MMD_TEST_TRACKER_PROP "tracker"
 
 typedef struct _ModuleStreamFixture
 {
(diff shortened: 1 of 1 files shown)


modulemd/tests/test-modulemd-modulestream.c
```diff
@@ -331,6 +332,136 @@ module_stream_v2_test_documentation (ModuleStreamFixture *fixture,
...
+
+  g_object_get (stream, MMD_TEST_TRACKER_PROP, &tracker_prop, NULL);
+  tracker = modulemd_module_stream_v1_get_tracker (stream);
+
+  g_assert_null (tracker);
+  g_assert_null (tracker_prop);
+
+  g_clear_pointer (&tracker_prop, g_free);
+
+  // Test Unicode values
```
OrionStar25: According to the docs, the `tracker` is:
`The website address of the upstream bug tracker for this module.`

So it doesn't really make sense to have Unicode characters in a website URL. Now we should either remove this test assuming no one is ever going to add a URL containing Unicode or handle this case in the main function by returning `NULL` whenever we encounter an invalid URL.

@sgallagher What do you suggest?

modulemd/tests/test-modulemd-modulestream.c
```diff
@@ -331,6 +332,136 @@ module_stream_v2_test_documentation (ModuleStreamFixture *fixture,
...
+
+  g_object_get (stream, MMD_TEST_TRACKER_PROP, &tracker_prop, NULL);
+  tracker = modulemd_module_stream_v2_get_tracker (stream);
+
+  g_assert_null (tracker);
+  g_assert_null (tracker_prop);
+
+  g_clear_pointer (&tracker_prop, g_free);
+
+  // Test Unicode values
```
OrionStar25: Same as above.

modulemd/tests/test-modulemd-modulestream.c
sgallagher: Leave it as-is for now. There's nothing *stopping* someone right now from putting unicode into the metadata here. We should probably look at using http-parser or llhttp to validate that we have validly-formatted URLs though. Mind opening a ticket to track that?

srikavin:
@sgallagher I've rebased this branch onto master.

PR #412, opened 2019-12-28
Modulemd.ModuleIndexMerger.test_merger_with_real_world_data into C

MR-M1M3:
Copied python test to C for https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/tests/ModulemdTests/merger.py#L238-254

### DIFF
--- a/modulemd/tests/test-modulemd-merger.c
+++ b/modulemd/tests/test-modulemd-merger.c
@@ -237,6 +237,45 @@ merger_test_add_conflicting_stream_and_profile_modified (void)
                    "8.2");
 }
 
+static void
+merger_test_with_real_world_data (void)
+{
+  g_autoptr (ModulemdModuleIndex) f29 = NULL;
+  g_autoptr (ModulemdModuleIndex) f29_updates = NULL;
+  g_autoptr (ModulemdModuleIndex) index = NULL;
+  g_autoptr (ModulemdModuleIndexMerger) merger = NULL;
+  g_autoptr (GError) error = NULL;
+  g_autoptr (GPtrArray) failures = NULL;
+  g_autofree gchar *yaml_path = NULL;
+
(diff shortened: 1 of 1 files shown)


modulemd/tests/test-modulemd-merger.c
```diff
@@ -273,3 +273,31 @@ main (int argc, char *argv[])
 
   return g_test_run ();
 }
+
+test_merger_with_real_world_data (ModuleStreamFixture *fixture,
```
OrionStar25: Every function in C needs a return type unlike in Python. Here the return type would be `static void`.

MR-M1M3: I've updated the file again and fixed some errors.

MR-M1M3: Here are the errors.
https://paste.centos.org/view/bc4bbcc9

OrionStar25: Please push your changes to this PR so that I can see them.

modulemd/modulemd-util.c
```diff
@@ -360,7 +360,7 @@ modulemd_validate_nevra (const gchar *nevra)
         {
           break;
         }
-      else if (*i == '-')
+      if (*i == '-')
```
OrionStar25: Why this?

sgallagher: The code formatter that we run during `ninja test` recently started to drop unneeded `else` statements. It's unrelated to this change. I've pushed this to the current master branch so it doesn't keep showing up.

OrionStar25:
@MR-M1M3 Please rebase and run `ninja test` before committing.

MR-M1M3:
I ran ninja test before commiting the changes, still the error exists

PR #414, opened 2020-01-03
Fix GDate autoptr cleanup build errors that started appearing with GLib 2.63.3

mmathesius:
The GDate autoptr cleanup was finally officially added in GLib 2.63.3--which is now available in Fedora Rawhide--so stop defining it ourselves.

https://gitlab.gnome.org/GNOME/glib/merge_requests/1256

### DIFF
--- a/modulemd/include/private/glib-extensions.h
+++ b/modulemd/include/private/glib-extensions.h
@@ -13,4 +13,7 @@
 
 #include <glib.h>
 
-G_DEFINE_AUTOPTR_CLEANUP_FUNC (GDate, g_date_free);
+/* GDate autoptr cleanup was finally added in GLib 2.63.3. */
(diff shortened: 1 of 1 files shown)


mmathesius:
> Actually, on further reflection, I would rather that we check `#ifndef GDate_autoptr` which is always defined if `G_DEFINE_AUTOPTR_CLEANUP_FUNC (GDate)` has been called. That way if someone decides to backport this patch to earlier glib2 releases, it won't break there.

@sgallagher I searched for something exactly like that `#define` that could be checked, but didn't find it--so I reverted to the version check. Unfortunately, the CI for Rawhide for #415 just blew up with the old GDate errors.

sgallagher:
I'm closing this PR in favor of #415 which uses a compile-check during the meson configuration to determine if we need to set the cleanup function.

PR #415, opened 2020-01-06
The GDate autoptr cleanup was finally officially added in GLib 2.63.3…

sgallagher:
…--which is now available in Fedora Rawhide--so stop defining it ourselves.

https://gitlab.gnome.org/GNOME/glib/merge_requests/1256

Fixes: https://github.com/fedora-modularity/libmodulemd/pull/414

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,11 +40,7 @@ jobs:
 
 
     - stage: Fedora x86_64
-      name: "Fedora 29 x86_64"
-      script: ./.travis/travis-fedora.sh
-      arch: amd64
-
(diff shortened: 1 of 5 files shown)


sgallagher:
Hmm, it seems that this test was insufficient. I am not sure why.

I left the title as @mmathesius had written it, but yeah... it's long.

sgallagher:
Okay, this version isn't going to work because `GDate_autoptr` is a typedef.

sgallagher:
I changed it so that we do a compile-test during meson configuration to detect whether we can use `g_autofree(GDate)` without defining the cleanup function.

PR #420, opened 2020-01-08
Copy Python tests for Modulemd.ModuleStream.rpm_api into C

srikavin:
This is related to #199.

### DIFF
--- a/modulemd/tests/test-modulemd-modulestream.c
+++ b/modulemd/tests/test-modulemd-modulestream.c
@@ -280,6 +280,63 @@ module_stream_v2_test_profiles (ModuleStreamFixture *fixture,
   g_clear_pointer (&rpms, g_strfreev);
 }
 
+
+static void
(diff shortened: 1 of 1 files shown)


modulemd/tests/test-modulemd-modulestream.c
```diff
@@ -2179,6 +2236,20 @@ main (int argc, char *argv[])
               module_stream_v2_test_profiles,
               NULL);
 
+  g_test_add ("/modulemd/v2/modulestream/v1/rpm_api",
```
OrionStar25: Please convert this to `g_test_add_func()` [see the GLib manual on how to use that] since we are going to replace it in the whole codebase with that anyway for simplicity.

sgallagher: We will do that in a separate request later. I'm going to merge this as-is.

PR #422, opened 2020-01-08
converted from g_test_add() to g_test_add_func()

sashreek1:
corrected the previous syntax and also modified the required functions to enable compatibility with g_test_add_func()

### DIFF
--- a/modulemd/tests/test-modulemd-buildopts.c
+++ b/modulemd/tests/test-modulemd-buildopts.c
@@ -27,7 +27,7 @@ typedef struct _BuildoptsFixture
 } BuildoptsFixture;
 
 static void
-buildopts_test_construct (BuildoptsFixture *fixture, gconstpointer user_data)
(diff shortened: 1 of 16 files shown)


sashreek1:
I dida run ```ninja test ``` before submitting so the code should have been formatted right? Moreover all the tests had passed.

modulemd/tests/test-dirty.py
```diff
@@ -11,6 +11,7 @@
 # For more information on the license, see COPYING.
 # For more information on free software, see
 # <https://www.gnu.org/philosophy/free-sw.en.html>.
+#comment
```
sgallagher: What is the point of this?

sashreek1: I just removed it i will  be pushing it

sashreek1:
I have been running ```ninja test``` before every push, yet it returns a failure

Issue #198, opened 2019-02-26
Copy python tests for Modulemd.ModuleIndexMerger into C


sgallagher:
Currently, most of the tests for the merge logic are run under Python. The valgrind tests for identifying memory leaks and errors can only run against tests in the C code. We need to replicate the python tests as C tests.

This would be a good task for a newcomer to the project. Look at the tests being run in [`modulemd/v2/tests/ModulemdTests/merger.py`](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/tests/ModulemdTests/merger.py) and copy them into [`modulemd/v2/tests/test-modulemd-merger.c`](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/v2/tests/test-modulemd-merger.c)

This will ensure that these code-paths are scanned with valgrind for leaks when `ninja test` is run (and when the CI test suite is executed).

SunitiJain:
Hi! I am an outreachy applicant. Please guide me how to contribute.

sgallagher:
@SunitiJain Please see https://github.com/fedora-modularity/libmodulemd/blob/master/README.md#getting-started-with-developing for a basic introduction into developing libmodulemd.

Madhumitha:
Can I work on this issue? I'm an outreachy applicant.

AarushiSingh09:
Can you please assign me this issue. I have tried to make some updates. I am outreachy applicant.

athira-selvam:
Is anyone working on this issue?

AarushiSingh09:
I am working on this issue.

SunitiJain:
Thanks .

yashpupneja:
@sgallagher I am Outreachy applicant and I have already started working on this issue . Can this issued be assigned to me as well?

OrionStar25:
Only [strict_default_streams](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/tests/ModulemdTests/merger.py#L259) is left to copy in this.

PR #425, opened 2020-01-11
Copy python tests for Modulemd.ModuleIndexMerger.merger into C

chenliTW:
For issue #198

### DIFF
--- a/modulemd/tests/test-modulemd-merger.c
+++ b/modulemd/tests/test-modulemd-merger.c
@@ -103,6 +103,227 @@ merger_test_deduplicate (CommonMmdTestFixture *fixture,
 
 
 static void
+merger_test_merger (CommonMmdTestFixture *fixture, gconstpointer user_data)
+{
+  g_autoptr (ModulemdModuleIndex) base_index = NULL;
+  g_autofree gchar *yaml_path = NULL;
+  g_autofree gchar *module_name = NULL;
+  ModulemdModule *httpd = NULL;
+  ModulemdDefaultsV1 *httpd_defaults = NULL;
+  GStrv httpd_profile_streams = NULL;
+  g_autoptr (ModulemdModuleIndex) override_nodejs_index = NULL;
+  g_autoptr (ModulemdModuleIndexMerger) merger = NULL;
+  g_autoptr (ModulemdModuleIndex) merged_index = NULL;
+  ModulemdModule *nodejs = NULL;
+  ModulemdDefaultsV1 *nodejs_defaults = NULL;
+  ModulemdDefaultsV1 *merged_httpd_defaults = NULL;
+  g_autoptr (ModulemdModuleIndex) override_index = NULL;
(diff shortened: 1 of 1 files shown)


modulemd/tests/test-modulemd-merger.c
```diff
@@ -20,6 +21,7 @@
 #include "modulemd-module-index.h"
 #include "private/test-utils.h"
 
+typedef int gint;
```
OrionStar25: You could also use `guint64` and not use this at all. This is all defined in `glib` already.

OrionStar25: You could also use `guint64` and not use this at all. This is all defined in `glib` already.

modulemd/tests/test-modulemd-merger.c
```diff
@@ -102,6 +104,164 @@ merger_test_deduplicate (CommonMmdTestFixture *fixture,
...
+
+  /*
+   * Test that override_index at a higher priority level succeeds
+   * Test that adding both of these at the same priority level fails
+   * with a merge conflict.
+   * Use randomly-selected high and low values to make sure we don't have
+   * sorting issues.
+   */
+  merger = modulemd_module_index_merger_new ();
+  random_low = rand()%100+1;
```
OrionStar25: Please look at this on how to use `random` with `GLib`. https://developer.gnome.org/glib/stable/glib-Random-Numbers.html

Declare your variable datatypes accordingly.

OrionStar25: Please look at this on how to use `random` with `GLib`. https://developer.gnome.org/glib/stable/glib-Random-Numbers.html

Declare your variable datatypes accordingly.

modulemd/tests/test-modulemd-merger.c
```diff
@@ -12,6 +12,7 @@
  */
 
 #include <glib.h>
+#include <stdlib.h>
```
OrionStar25: I'm assuming you added this for `rand`. This can be eliminated by using the functionality in `glib`

chenliTW:
I have run ```ninja test``` and change the ```rand()``` into ``g_random_int_range()```

OrionStar25:
Please see the log from the failing CI. You have a bunch of warnings because of incompatible data types. Either typecast your parameters or use functions from the `glib` manual accordingly. We need to eliminate all the warnings. 

Also, you must run `ninja test` *right* before committing. Running `ninja test` after committing or before making another change would only fail the CI.

chenliTW:
I have typecast all the parameters and eliminate all the warnings. But the CI test seems failing while I  run ```ninja test``` right before committing . Are there something that I missed to correct?

chenliTW:
Th

> I have typecast all the parameters and eliminate all the warnings. But the CI test seems failing while I run `ninja test` right before committing . Are there something that I missed to correct?

The CI test passed after I use clang-format .

Issue #429, opened 2020-01-21
Search ModuleIndex for streams containing particular NEVRA


kontura:
It should also support fnmatching.

We can provide separated strings of NEVRA as different arguments.
Filters for: name, version, epoch, release, architecture and also the full NEVRA string.

sgallagher:
@kontura 
Is it acceptible to give you just the full nevra string matching for now? We don't actually maintain the split NEVRA in the module metadata, so providing the split lookup would be difficult. Even if you have the components separated already, merging them into a NEVRA format should be a trivial `sprintf()`.

I'm thinking we'd have: `modulemd_module_index.search_rpms_by_nevra(nevra_glob)` as the interface, returning a `GPtrArray` of `ModulemdModuleStream` objects that match the provided NEVRA. Does that seem reasonable to you?

PR #431, opened 2020-01-23
Add ModuleIndex.search_rpms()

sgallagher:
Resolves: https://github.com/fedora-modularity/libmodulemd/issues/429

This patch adds the ability to query the `ModuleIndex` for which `ModuleStream` object(s) provide a given RPM nevra (or glob). This will make some DNF lookups easier (which may translate into repoquery enhancements).

There's also a second, minor patch to provide a new macro: `UNUSED()` which allows us to indicate that we are intentionally not using a function argument in the body (such as when we're defining the implementation of a callback function and don't need to use some of the data passed in.)

This set of patches builds atop the patches from PR #430

### DIFF
--- a/modulemd/include/private/modulemd-util.h
+++ b/modulemd/include/private/modulemd-util.h
@@ -310,6 +310,48 @@ modulemd_validate_nevra (const gchar *nevra);
 gboolean
 modulemd_boolean_equals (gboolean a, gboolean b);
 
+
+/**
+ * modulemd_is_glob_pattern:
+ * @pattern: (nullable) A string to check for glob patterns as defined by
+ * [glob(7)[(https://www.mankier.com/7/glob)
+ *
+ * Returns: TRUE if @pattern contains any globbing characters. FALSE
+ * otherwise.
+ *
+ * Since: 2.9
+ */
+gboolean
+modulemd_is_glob_pattern (const char *pattern);
+
+
+/**
+ * compare_streams:
+ * @a: The first #ModulemdModuleStream to sort
+ * @b: The second #ModulemdModuleStream to sort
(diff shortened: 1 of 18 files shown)


modulemd/include/private/modulemd-util.h
```diff
@@ -352,6 +358,23 @@ gboolean
 modulemd_fnmatch (const gchar *pattern, const gchar *string);
 
 
+/**
+ * modulemd_rpm_match:
+ * @key: The key in the RPM artifacts hash table.
+ * @value: The value in the RPM artifacts hash table (should be the same
+ * as @key
```
mmathesius:
```suggestion
 * as @key).
```

modulemd/include/private/modulemd-util.h
```diff
@@ -352,6 +358,23 @@ gboolean
 modulemd_fnmatch (const gchar *pattern, const gchar *string);
 
 
+/**
```
mmathesius: This `modulemd_rpm_match()` definition seems to have been added to this header file by a different commit that the one adding the implementation. That might cause problems with bisection...

modulemd/include/private/modulemd-util.h
sgallagher: Good catch.

modulemd/include/private/modulemd-util.h
sgallagher: Oops. I'll fix that.

sgallagher:
@mmathesius Thanks for the feedback. I made the requested changes and pushed it.

mmathesius:
LGTM. Thanks!

Issue #386, opened 2019-10-24
Implement retry logic in CI tests


sgallagher:
There are a few places in the CI tests that require external resources (`dnf install` and `docker pull`, mainly). Occasionally one of these hits a network hiccup and fails, causing the test to fail (and all subsequent stages to be canceled).

We should implement a retry mechanism for these calls to make the tests more reliable. It's probably sufficient to try three times before giving up.

mmathesius:
Fix by #434.

PR #434, opened 2020-01-28
Implement retry logic in CI tests

mmathesius:
An initial test for Fedora only to see that it works in the CI environment.

@sgallagher Is this what you had in mind for #386?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -94,3 +94,5 @@ jobs:
 
   allow_failures:
     - name: "OpenMandriva cooker"
+    - name: "Fedora rawhide x86_64"
+    - name: "Fedora rawhide aarch64"
--- a/.travis/docs/travis-tasks.sh
+++ b/.travis/docs/travis-tasks.sh
@@ -4,6 +4,7 @@
 set -e
 set -x
 
+RETRY_CMD=/builddir/.travis/retry-command.sh
 
 pushd /builddir/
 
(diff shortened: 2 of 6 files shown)


mmathesius:
@sgallagher Glad you like it! I presume you're OK with the defaults I chose of 3 tries with a 2 second delay between attempts.

Do you think it's worth having the command retry loop emit messages when it encounters a failure and cycles to try again?

Also, do you think the `git` commands in `.travis/docs/travis-tasks.sh` should be wrapped to retry?

sgallagher:
> @sgallagher Glad you like it! I presume you're OK with the defaults I chose of 3 tries with a 2 second delay between attempts.

Yes, that's fine.

> Do you think it's worth having the command retry loop emit messages when it encounters a failure and cycles to try again?

I don't think so; it would just clutter the logs (which we only look at when things fail, so...)

> Also, do you think the `git` commands in `.travis/docs/travis-tasks.sh` should be wrapped to retry?
I assume you mean just the clone and the push. Probably a good idea. Maybe the rsync as well.

mmathesius:
Sorry for the flurry of force pushes. I think I've got everything appropriate wrapped now except the `rsync` in `./docs/travis-tasks.sh`. The `eval` in the retry script is having issues parsing the parenthesis in paths such as `.../modulemd-2.0-Modulemd.Module-(Private).html`. I'll see if I can figure out a workaround for that. But that can be a subsequent enhancement if you want to get this in sooner.

sgallagher:
> The `eval` in the retry script is having issues parsing the parenthesis in paths such as `.../modulemd-2.0-Modulemd.Module-(Private).html`. I'll see if I can figure out a workaround for that. 

Can you just put the rsync command into its own script file and just have the retry execute that? That should work around the problem reasonably well.

mmathesius:
@sgallagher Actually, upon closer evaluation, the `rsync` isn't doing any remote operations. It's just copying files from one local directory to another. The subsequent `git push` is what does the remote copy and it's already been wrapped.

sgallagher:
> @sgallagher Actually, upon closer evaluation, the `rsync` isn't doing any remote operations. It's just copying files from one local directory to another. The subsequent `git push` is what does the remote copy and it's already been wrapped.

Oh right. I completely forgot that... well, great!

Issue #364, opened 2019-09-05
Create modulemd-packager YAML document type


sgallagher:
Right now, we have a complicated libmodulemd metadata format that is both the input and output for the build system. The input that packagers are expected to provide is a very small subset of the total specification. It would be simpler on packagers if they could write a document in a simplified format and use that as input to the build system.

I propose that we create a specification as follows:
```yaml
# Document type identifier
document: modulemd-packager
# Module metadata format version
version: 2
data:
    # A short summary describing the module, required
    summary: An example module
    # A verbose description of the module, required
    description: >-
        A module for the demonstration of the metadata format. Also,
        the obligatory lorem ipsum dolor sit amet goes right here.
    # Module and content licenses in the Fedora license identifier
    # format, required
    license:
        # Module license, required
        # This list covers licenses used for the module metadata and
        # possibly other files involved in the creation of this specific
        # module.
        module:
            - MIT
        # Content license, optional
        # A list of licenses used by the packages in the module.
        # This should be populated by build tools, not the module author.
        content:
            - Beerware
            - GPLv2+
            - zlib
    # Module dependencies, if any.  Optional.
    # A list of dictionaries describing build and runtime dependencies
    # of this module.  Each list item describes a combination of dependencies
    # this module can be built or run against.
    # Dependency keys are module names, dependency values are lists of
    # required streams.  The lists can be both inclusive (listing compatible
    # streams) or exclusive (accepting every stream except for those listed).
    # An empty list implies all active existing streams are supported.
    # Requiring multiple streams at build time will result in multiple
    # builds.  Requiring multiple streams at runtime implies the module
    # is compatible with all of them.  If the same module streams are listed
    # in both the build time and the runtime block, the build tools translate
    # the runtime block so that it matches the stream the module was built
    # against.  Multiple builds result in multiple output modulemd files.
    # See below for an example.
    # TODO: Provides, conflicts, obsoletes, recommends, etc.
    #       Do we even need those?
    # The example below illustrates how to build the same module in four
    # different ways, with varying build time and runtime dependencies.
    dependencies:
        # Build on all available platforms except for f27, f28 and epel7
        # After build, the runtime dependency will match the one used for
        # the build.
        - buildrequires:
              platform: [-f27, -f28, -epel7]
          requires:
              platform: [-f27, -f28, -epel7]
        # For platform:f27 perform two builds, one with buildtools:v1, another
        # with buildtools:v2 in the buildroot.  Both will also utilize
[...]

PR #435, opened 2020-01-29
Add support for modulemd-packager documents

sgallagher:
This also moves the YAML spec documents into a common location under `yaml_specs/` and fixes a small inconsistency with error-setting. While doing so, I removed some unused YAML files from the `mod-defaults` directory.

Resolves: https://github.com/fedora-modularity/libmodulemd/issues/364

~~Marked as WIP because it doesn't yet support actually restricting the available options to only the packager ones. That will be added in a subsequent patch, but I wanted to get the first part out for review sooner.~~


yaml_specs/modulemd_packager_v2.yaml
```diff
@@ -0,0 +1,271 @@
+# Document type identifier
```
mmathesius: Perhaps a line in here somewhere that states what this document type represents--and that it is a simplified subset of the full 'modulemd' type?

Actually, none of the existing specs seem to contain a self-explanatory description of what they represent. They just jump right into detailed descriptions of each individual key.

modulemd/tests/ModulemdTests/moduleindex.py
```diff
@@ -42,15 +42,18 @@ def test_constructors(self):
...
             self.assertTrue(res)
             self.assertListEqual(failures, [])
 
         for fname in [
-            "spec.v2.yaml",
-            "translations/spec.v1.yaml",
-            "mod-defaults/spec.v1.yaml",
+            "yaml_specs/modulemd_stream_v2.yaml",
+            "yaml_specs/modulemd_translations_v1.yaml",
+            "yaml_specs/modulemd_defaults_v1.yaml",
```
mmathesius: This would be a good time to add a test here that reads `"yaml_specs/modulemd_packager_v2.yaml"`. Although not in this loop...

modulemd/tests/test-modulemd-modulestream.c
```diff
@@ -1021,6 +1021,21 @@ module_stream_test_v2_yaml (void)
...
+  g_autoptr (ModulemdModuleStream) stream = NULL;
+  g_autoptr (GError) error = NULL;
+
+  g_autofree gchar *specV2Path = g_strdup_printf (
+    "%s/yaml_specs/modulemd_stream_v2.yaml", g_getenv ("MESON_SOURCE_ROOT"));
+  stream =
+    modulemd_module_stream_read_file (specV2Path, TRUE, NULL, NULL, &error);
+  g_assert_no_error (error);
+  g_assert_nonnull (stream);
+}
```
mmathesius: `g_clear_object (&stream); g_clear_pointer (&specV2Path, g_free);` for good measure?

modulemd/tests/test-modulemd-modulestream.c
```diff
@@ -1021,6 +1021,21 @@ module_stream_test_v2_yaml (void)
...
 
+
+static void
+module_packager_v2_sanity (void)
+{
+  g_autoptr (ModulemdModuleStream) stream = NULL;
+  g_autoptr (GError) error = NULL;
+
+  g_autofree gchar *specV2Path = g_strdup_printf (
+    "%s/yaml_specs/modulemd_stream_v2.yaml", g_getenv ("MESON_SOURCE_ROOT"));
```
mmathesius: Did you intend this to be `modulemd_packager_v2.yaml`? Otherwise it's completely redundant with the very last check in the function above.

modulemd/include/private/modulemd-module-stream-v2-private.h
```diff
@@ -75,7 +75,10 @@ struct _ModulemdModuleStreamV2
  * document.
  * @strict: (in): Whether the parser should return failure if it encounters an
  * unknown mapping key or if it should ignore it.
- * @error: (out): A #GError that will return the reason for a parsing or
+ * @only_packager: (in): Whether the parser should return failure if it
+ * encounters a mapping key that is not intended for packager use. Ignored
+ * if @strict is #FALSE
+ * @error: (out): A #GError that will reparse_yturn the reason for a parsing or
```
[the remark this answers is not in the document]
sgallagher: Probably a middle-click while scrolling.

modulemd/include/modulemd-2.0/modulemd-errors.h
```diff
@@ -54,7 +54,8 @@ typedef enum
   MODULEMD_ERROR_NO_MATCHES,
   MODULEMD_ERROR_TOO_MANY_MATCHES,
   MODULEMD_ERROR_MAGIC,
-  MODULEMD_ERROR_NOT_IMPLEMENTED
+  MODULEMD_ERROR_NOT_IMPLEMENTED,
+  MODULEMD_ERROR_MISSING_REQUIRED
```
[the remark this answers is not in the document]
sgallagher: Whoops, missed that. Will do.

modulemd/include/private/modulemd-yaml.h
```diff
@@ -68,7 +68,8 @@ typedef enum
   MODULEMD_YAML_DOC_UNKNOWN = 0,
   MODULEMD_YAML_DOC_MODULESTREAM,
   MODULEMD_YAML_DOC_DEFAULTS,
-  MODULEMD_YAML_DOC_TRANSLATIONS
+  MODULEMD_YAML_DOC_TRANSLATIONS,
+  MODULEMD_YAML_DOC_PACKAGER
```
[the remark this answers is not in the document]
sgallagher: Ditto

modulemd/tests/ModulemdTests/moduleindex.py
sgallagher: It cannot actually be read into a ModuleIndex because it has no `name` attribute to use as a lookup key. It must be read by `Modulemd.ModuleStream.read_*()` instead.

modulemd/tests/test-modulemd-modulestream.c
sgallagher: Probably good hygiene.

modulemd/tests/test-modulemd-modulestream.c
sgallagher: Whoops... yes I did.

modulemd/tests/ModulemdTests/moduleindex.py
mmathesius: I was aware of that. What I overlooked was that this is `module**index**.py`, so it doesn't belong here. Such a python test belongs in `module**stream**.py`--which is exactly what your added `test_packager_sanity()` does. Nevermind.  :)

yaml_specs/modulemd_packager_v2.yaml
mmathesius: Thank you for adding the self-descriptions in each of the specs!

sgallagher:
@mmathesius OK, I just pushed the remaining series of patches. There's quite a bit of stuff in there, but I tried to break them up into smaller patches to make review simpler.

mmathesius:
> @mmathesius OK, I just pushed the remaining series of patches. There's quite a bit of stuff in there, but I tried to break them up into smaller patches to make review simpler.

@sgallagher Sounds good. I'll dig through it tomorrow.

sgallagher:
I just pushed an additional patch that excludes the use of `buildorder` in the `document: modulemd-packager`. This isn't yet supported in MBS, but I think we want to encourage folks to migrate to this much-easier approach as soon as it is. So I think it's a good idea to land this in the initial version and just not update the packager docs in Fedora until the MBS can handle it.

Worst case scenario: if we decide we do need `buildorder` after all, it's not an ABI break to lessen the restriction.

yaml_specs/modulemd_packager_v2.yaml
```diff
@@ -0,0 +1,255 @@
...
+                    - xyz
+            xyz:
+                rationale: xyz is a bundled dependency of xxx.
+        # Module content of this module
+        # Included modules are built in the shared buildroot, together with
+        # other included content.  Keys are module names, values additional
+        # component information.  Note this only includes components and their
+        # properties from the referenced module and doesn't inherit any
+        # additional module metadata such as the module's dependencies or
+        # component buildopts.
```
mmathesius: Since the packager spec drops `buildopts`, should this reference to `buildopts` also be dropped?

modulemd/tests/test-modulemd-component-rpm.c
```diff
@@ -418,7 +418,7 @@ component_rpm_test_parse_yaml (void)
   g_assert_cmpint (event.type, ==, YAML_SCALAR_EVENT);
   yaml_event_delete (&event);
 
-  r = modulemd_component_rpm_parse_yaml (&parser, "bar", TRUE, &error);
+  r = modulemd_component_rpm_parse_yaml (&parser, "bar", TRUE, FALSE, &error);
```
mmathesius: Should a another C test be added using the same input file but with `only_packager = TRUE` to make sure it grumbles about the presence of `buildorder`? I do see you added a Python test for that, so it would be someone redundant.

modulemd/tests/ModulemdTests/modulepackager.py
```diff
@@ -0,0 +1,191 @@
...
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+class TestModulemdPackager(TestBase):
+    def test_sanity(self):
+        # Validate that the specification parses correctly
+        stream = Modulemd.ModuleStream.read_file(
+            os.path.join(
+                self.source_root, "yaml_specs/modulemd_stream_v2.yaml"
```
mmathesius:
```suggestion
                self.source_root, "yaml_specs/modulemd_packager_v2.yaml"
```
???

modulemd/include/modulemd-2.0/modulemd-errors.h
```diff
@@ -25,6 +25,62 @@ G_BEGIN_DECLS
...
+ * metadata version of a module stream or module defaults.
+ * @MMD_ERROR_VALIDATE: Represents an error encountered while validating
+ * module metadata.
+ * @MMD_ERROR_FILE_ACCESS: Represents an error encountered when attempting
+ * to access a file.
+ * @MMD_ERROR_NO_MATCHES: Represents an error indicating that no streams
+ * matched when searching for a specific module stream. Since: 2.2
+ * @MMD_ERROR_TOO_MANY_MATCHES: Represents an error indicating that
+ * multiple streams matched when searching for a specific module
+ * stream. Since: 2.2
```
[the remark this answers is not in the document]
sgallagher: Yes; I forgot to change that. Thanks.

modulemd/include/modulemd-2.0/modulemd-errors.h
```diff
@@ -37,14 +93,18 @@ G_BEGIN_DECLS
  * matched when searching for a specific module stream. Since: 2.2
  * @MODULEMD_ERROR_TOO_MANY_MATCHES: Represents an error indicating that
  * multiple streams matched when searching for a specific module
- * stream. Since: 2.2
+ * stream.
```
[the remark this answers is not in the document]
sgallagher: I think I deleted that from the wrong enum doc section.

modulemd/include/modulemd-2.0/modulemd-errors.h
```diff
@@ -37,14 +93,18 @@ G_BEGIN_DECLS
...
  * @MODULEMD_ERROR_MAGIC: Could not detect the mime type of a file for
- * automatic detection of compression format. Since: 2.8
+ * automatic detection of compression format.
  * @MODULEMD_ERROR_NOT_IMPLEMENTED: The requested function is not implemented
  * on this platform, likely due to needing a newer version of a dependency
- * library. Since: 2.8
+ * library.
+ * @MODULEMD_ERROR_MISSING_REQUIRED: The object is missing some data necessary
```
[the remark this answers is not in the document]
sgallagher: Yeah, I can do that. I hadn't originally expected to deprecate that enum.

yaml_specs/modulemd_packager_v2.yaml
sgallagher: Agreed

modulemd/tests/test-modulemd-component-rpm.c
sgallagher: That's probably a good idea.

modulemd/tests/ModulemdTests/modulepackager.py
sgallagher: :man_facepalming:

sgallagher:
OK, I think I've made all the requested changes now.

modulemd/include/modulemd-2.0/modulemd-errors.h
```diff
@@ -45,6 +101,8 @@ G_BEGIN_DECLS
  * library. Since: 2.8
  *
  * Since: 2.0
+ * Deprecated: 2.9
+ * Use #ModulemdError instead.
```
mmathesius: Two things I just dug up...

On the gtk-doc doc page explaining the use of "Deprecated:" at https://developer.gnome.org/gtk-doc-manual/stable/documenting_symbols.html.en, the provided examples show the format to use is a single line: "Deprecated: 2.9: Use #ModulemdError instead."

I also noticed there is a `MMD_DEPRECATED_TYPE_FOR` macro defined in `modulemd-deprecated.h`. `MMD_DEPRECATED_TYPE_FOR (ModulemdError)` should be placed below.

modulemd/include/modulemd-2.0/modulemd-errors.h
```diff
@@ -78,7 +195,10 @@ typedef enum
  * encountered while parsing a YAML document.
  *
  * Since: 2.0
+ * Deprecated: 2.9
+ * Use #ModulemdYamlError instead.
```
mmathesius: As noted above...

A single line "Deprecated: 2.9: Use #ModulemdYamlError instead."

And `MMD_DEPRECATED_TYPE_FOR (ModulemdYamlError)` below.

modulemd/include/modulemd-2.0/modulemd-errors.h
mmathesius: Of course, that also means `#include "modulemd-deprecated.h"` is needed above.

mmathesius: Experimentation shows that gtk-doc also picks up the text on the lines immediately following the "Deprecated:" line and properly includes it in the generated documentation. So there's no need to change that.

But please do add `MMD_DEPRECATED_TYPE_FOR()`. Thanks!

sgallagher: It turns out that `MMD_DEPRECATED_TYPE_FOR()` does not work for the enum as a whole, but I can specify it for the individual enumerations within it instead.

sgallagher:
@mmathesius OK, I changed up the deprecation functionality. I couldn't apply a deprecation to the enum as a whole, but I can apply it to the individual names, so I did that.

mmathesius:
> @mmathesius OK, I changed up the deprecation functionality. I couldn't apply a deprecation to the enum as a whole, but I can apply it to the individual names, so I did that.

@sgallagher Huh. Deprecating the entire type works for me. I tested it out starting with your latest commit from last Friday and I made the following local changes:

```
diff --git a/modulemd/include/modulemd-2.0/modulemd-errors.h b/modulemd/include/modulemd-2.0/modulemd-errors.h                                                                
index 82328f9..ed34851 100644                                                                                                                                                 
--- a/modulemd/include/modulemd-2.0/modulemd-errors.h                                  
+++ b/modulemd/include/modulemd-2.0/modulemd-errors.h                                  
@@ -13,6 +13,7 @@                                                                      
                                  
 #pragma once                                                                          
                                           
+#include "modulemd-deprecated.h"          
 #include <glib.h>              
                                                                                                                                                                              
 G_BEGIN_DECLS                                                                                                                                                                
@@ -104,6 +105,7 @@ typedef enum      
  * Deprecated: 2.9                                                                                                                                                           
  * Use #ModulemdError instead.
  */                                                                                   
+MMD_DEPRECATED_TYPE_FOR (ModulemdError)                                               
 typedef enum    
 {
   MODULEMD_ERROR_UPGRADE,
@@ -198,7 +200,7 @@ typedef enum
  * Deprecated: 2.9              
  * Use #ModulemdYamlError instead.
  */
-             
+MMD_DEPRECATED_TYPE_FOR (ModulemdYamlError)                                           
 typedef enum      
 {                                                                                                                                                                            
   MODULEMD_YAML_ERROR_OPEN, 

```

Then I used meson/ninja to build that into build directory `debugbuild`.

Then I compiled a very short C program:

```
#include "modulemd.h"

int
main (int argc, char *argv[])
{
  ModulemdErrorEnum dep1;
  ModulemdError dep2;
  ModulemdYamlErrorEnum dep3;
  ModulemdYamlError dep4;
}
```

Using:

```
gcc -g -I./modulemd/include/modulemd-2.0 -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -L./debugbuild/modulemd -lmodulemd -lgobject-2.0 -lglib-2.0 deptest.c
deptest.c: In function ‘main’:
deptest.c:6:3: warning: ‘ModulemdErrorEnum’ is deprecated: Use 'ModulemdError' instead [-Wdeprecated-declarations]
    6 |   ModulemdErrorEnum dep1;
      |   ^~~~~~~~~~~~~~~~~
In file included from ./modulemd/include/modulemd-2.0/modulemd.h:25,
                 from deptest.c:1:
./modulemd/include/modulemd-2.0/modulemd-errors.h:110:1: note: declared here
  110 | {
      | ^
deptest.c:8:3: warning: ‘ModulemdYamlErrorEnum’ is deprecated: Use 'ModulemdYamlError' instead [-Wdeprecated-declarations]
    8 |   ModulemdYamlErrorEnum dep3;
      |   ^~~~~~~~~~~~~~~~~~~~~
In file included from ./modulemd/include/modulemd-2.0/modulemd.h:25,
                 from deptest.c:1:
./modulemd/include/modulemd-2.0/modulemd-errors.h:205:1: note: declared here
  205 | {
      | ^
```

Adding the `-Wno-deprecated-declarations` compiler flag silences the warnings.

Have I oversimplified my test?

mmathesius:
OK. I do see where simply deprecating the type doesn't catch the cases where the deprecated enums are used directly without declaring a variable...

sgallagher:
> OK. I do see where simply deprecating the type doesn't catch the cases where the deprecated enums are used directly without declaring a variable...

@mmathesius 
It turns out that this is a lot more subtle than it first appeared. The `glib-mkenums` command doesn't properly generate the enum types if I include the deprecations on the individual enum values, resulting in them not appearing in the documentation or the language bindings. So I'm going to drop those and go back to just deprecating the enum type itself.

sgallagher:
I dropped the fancy attempt to deprecate the individual enum values and then just merged the type deprecation into the original patch to add the new enums. This should also have the side-effect of unbreaking the CentOS 7 build.

PR #440, opened 2020-02-05
CI retry backoff

mmathesius:
To help with the intermittent failures of the libmodulemd CI tests...

### DIFF
--- a/.travis/retry-command.sh
+++ b/.travis/retry-command.sh
@@ -1,17 +1,20 @@
 #!/bin/bash
 
 function retry_command {
-    local usage="Usage: ${FUNCNAME[0]} [-n numtries] [-d delay]"
+    local usage="Usage: ${FUNCNAME[0]} [-b backoff-factor] [-d delay] [-n numtries]"
     local OPTIND OPTION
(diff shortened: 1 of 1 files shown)


mmathesius:
And CI still failed due to persistent HTTP `503 Service Temporarily Unavailable` problems...

https://api.travis-ci.org/v3/job/646482492/log.txt

.travis/retry-command.sh
```diff
@@ -1,18 +1,21 @@
...
-    local numtries=3 delay=2
+    local backoff_factor=2 delay=3 numtries=4
 
-    while getopts ":n:d:" OPTION; do
+    while getopts ":bd:n:" OPTION; do
         case "${OPTION}" in
-        n)
-            numtries=${OPTARG}
+        b)
+            backoff_factor=1
```
sgallagher: So, by default we double the time between each retry, but if we pass `-b` we don't back off? That seems less than obvious. Maybe make `-b` take an option for the factor?

mmathesius: I considered doing it that way originally. But I've done it now.

PR #441, opened 2020-02-05
CI: Get Fedora images from Quay.io

sgallagher:
The container registry at registry.fedoraproject.org  is unreliable
and the one at Quay.io is better maintained. Quay.io also updates
its rawhide images at each successful rawhide compose, so we won't
lose our fast notification of build breakage.

### DIFF
--- a/.travis/get_rawhide_version.py
+++ b/.travis/get_rawhide_version.py
@@ -1,0 +1,45 @@
+#!/usr/bin/python3
+# -*- coding: utf-8 -*-
+# This file is part of libmodulemd
+# Copyright (C) 2020 Stephen Gallagher
+#
+# Fedora-License-Identifier: MIT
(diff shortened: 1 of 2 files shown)


.travis/get_rawhide_version.py
```diff
@@ -0,0 +1,40 @@
...
+
+def get_fedora_rawhide_version(session):
+    # Koji sometimes disconnects for no apparent reason. Retry up to 5
+    # times before failing.
+    for attempt in range(5):
+        try:
+            build_targets = session.getBuildTargets('rawhide')
+        except requests.exceptions.ConnectionError:
+            logging.warning(
+                "Connection lost while retrieving rawhide branch, retrying...")
```
mmathesius: Perhaps a `time.sleep()` for a few seconds here?

Issue #333, opened 2019-07-15
Create valgrind suppression file for Python tests


sgallagher:
When running valgrind against the Python tests for libmodulemd, we get a lot of noise. We need to analyze it and identify

* What is a false positive
* What is real but caused by GObject and not us
* What we need to address in libmodulemd

@mmathesius Would you mind looking into this? http://valgrind.org/docs/manual/manual-core.html#manual-core.suppress has notes on how to generate a suppression file.

sgallagher:
The ideal outcome here of course is to be able to add the python tests to the valgrind test that runs as part of our CI. Right now they are excluded and only the C tests are run.

It'll also make it easier when writing new tests, as we won't have to duplicate them in both Python and C.

sgallagher:
This was merged for 2.9.0

PR #442, opened 2020-02-05
valgrind suppressions file for Python tests

mmathesius:
Progress towards #333.

Valgrind failures for Python scripts are currently considered non-fatal if `MMD_IGNORE_PYTHON_VALGRIND` is set in the environment--which is currently done in `modulemd/meson.build`.

The big question is... what should be included in the suppression file? For starters, I included a few from the `valgrind-python.supp` file included in the Python 2 and 3 installs that were prevalent when running the tests.

Also, am I allowed to include `contrib/valgrind/grindmerge` here? It's a handy script I found for extracting and merging suppression entries from the test logfile.

### DIFF
--- a/modulemd/meson.build
+++ b/modulemd/meson.build
@@ -413,12 +413,16 @@ foreach name, script : python_tests
           env : py_test_release_env,
           args : files(script))
 
+    valgrind_tests += name + '_python3_debug'
+
     test (name + '_python2_debug', python2,
           env : py_test_env,
           args : files(script))
     test (name + '_python2_release', python2,
           env : py_test_release_env,
           args : files(script))
+
+    valgrind_tests += name + '_python2_debug'
 endforeach
 
 
--- a/modulemd/tests/test-valgrind.py
+++ b/modulemd/tests/test-valgrind.py
@@ -23,6 +23,12 @@ from multiprocessing import Pool, TimeoutError
 if os.getenv("MMD_SKIP_VALGRIND"):
     sys.exit(77)
 
+supp_file = os.path.join(
+    os.getenv("MESON_SOURCE_ROOT", "."),
+    "contrib",
+    "valgrind",
+    "libmodulemd-python.supp",
+)
 
 failed = False
 
@@ -41,8 +47,9 @@ with tempfile.TemporaryDirectory(prefix="libmodulemd_valgrind_") as tmpdirname:
             "/usr/bin/valgrind "
             "--leak-check=full "
             "--suppressions=/usr/share/glib-2.0/valgrind/glib.supp "
+            "--suppressions=%s "
             "--xml=yes "
-            "--xml-file=%s/%s.xml " % (tmpdirname, test)
+            "--xml-file=%s/%s.xml " % (supp_file, tmpdirname, test)
         )
         proc_result = subprocess.run(
             [
@@ -70,6 +77,8 @@ with tempfile.TemporaryDirectory(prefix="libmodulemd_valgrind_") as tmpdirname:
                 failed = True
                 continue
 
+            test_failed = False
+
             # Process the XML for leaks
             tree = ET.parse("%s/%s.xml" % (tmpdirname, test))
             root = tree.getroot()
@@ -83,21 +92,21 @@ with tempfile.TemporaryDirectory(prefix="libmodulemd_valgrind_") as tmpdirname:
                                     "Memory leak detected in %s" % test,
                                     file=sys.stderr,
                                 )
-                                failed = True
+                                test_failed = True
 
                             elif error_child.text == "InvalidFree":
                                 print(
                                     "Invalid free() detected in %s" % test,
                                     file=sys.stderr,
                                 )
-                                failed = True
+                                test_failed = True
 
                             elif error_child.text == "InvalidRead":
                                 print(
                                     "Invalid read detected in %s" % test,
                                     file=sys.stderr,
                                 )
-                                failed = True
+                                test_failed = True
 
                             elif error_child.text == "UninitCondition":
                                 print(
@@ -105,11 +114,12 @@ with tempfile.TemporaryDirectory(prefix="libmodulemd_valgrind_") as tmpdirname:
                                     % test,
                                     file=sys.stderr,
                                 )
-                                failed = True
-            if failed:
+                                test_failed = True
+
+            if test_failed:
                 with open("%s/%s.xml" % (tmpdirname, test), "r") as xml:
                     print(xml.read())
-
+                failed = True
 
 if failed:
     sys.exit(1)
--- a/contrib/valgrind/libmodulemd-python.supp
+++ b/contrib/valgrind/libmodulemd-python.supp
@@ -1,0 +1,80 @@
+# libmodulemd valgrind suppressions file
+#
+# This provides a list of suppressions for libmodulemd for valgrind for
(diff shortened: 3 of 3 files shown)


mmathesius:
FWIW, when including the following four very specific valgrind suppression entries (in addition to the three general entries already included), the valgrind test has no complaints. But I haven't been able to figure out if they're real problems or not.
```
{
   <insert_a_suppression_name_here>
   Memcheck:Leak
   match-leak-kinds: definite
   fun:malloc
   fun:g_malloc
   fun:g_slice_alloc
   fun:g_hash_table_new_full
   fun:modulemd_yaml_parse_string_set
   fun:modulemd_module_stream_v1_parse_licenses
   fun:modulemd_module_stream_v1_parse_yaml
   fun:modulemd_module_stream_read_yaml
   fun:modulemd_module_stream_read_string
   fun:ffi_call_unix64
   fun:ffi_call
   obj:/usr/lib64/python3.7/site-packages/gi/_gi.cpython-37m-x86_64-linux-gnu.so
}
{
   <insert_a_suppression_name_here>
   Memcheck:Leak
   match-leak-kinds: definite
   fun:malloc
   fun:_PyMem_RawWcsdup
   obj:/usr/lib64/libpython3.7m.so.1.0
   fun:_Py_InitializeCore
   obj:/usr/lib64/libpython3.7m.so.1.0
   obj:/usr/lib64/libpython3.7m.so.1.0
   fun:_Py_UnixMain
   fun:(below main)
}
{
   <insert_a_suppression_name_here>
   Memcheck:Cond
   fun:PyUnicode_Decode
   fun:PyUnicode_FromEncodedObject
   obj:/usr/lib64/libpython3.7m.so.1.0
   fun:_PyObject_FastCallKeywords
   obj:/usr/lib64/libpython3.7m.so.1.0
   fun:_PyEval_EvalFrameDefault
   fun:_PyFunction_FastCallDict
   fun:_PyObject_Call_Prepend
   obj:/usr/lib64/libpython3.7m.so.1.0
   fun:_PyObject_FastCallKeywords
   obj:/usr/lib64/libpython3.7m.so.1.0
   fun:_PyEval_EvalFrameDefault
}
{
   <insert_a_suppression_name_here>
   Memcheck:Leak
   match-leak-kinds: definite
   fun:malloc
   fun:g_malloc
   fun:g_strdup
   obj:/usr/lib64/python3.7/site-packages/gi/_gi.cpython-37m-x86_64-linux-gnu.so
   obj:/usr/lib64/python3.7/site-packages/gi/_gi.cpython-37m-x86_64-linux-gnu.so
   obj:/usr/lib64/python3.7/site-packages/gi/_gi.cpython-37m-x86_64-linux-gnu.so
   obj:/usr/lib64/python3.7/site-packages/gi/_gi.cpython-37m-x86_64-linux-gnu.so
   obj:/usr/lib64/python3.7/site-packages/gi/_gi.cpython-37m-x86_64-linux-gnu.so
   fun:PyObject_SetAttr
   fun:_PyEval_EvalFrameDefault
   fun:_PyFunction_FastCallKeywords
   obj:/usr/lib64/libpython3.7m.so.1.0
}
```

sgallagher:
> Progress towards #333.
> 
> Valgrind failures for Python scripts are currently considered non-fatal if `MMD_IGNORE_PYTHON_VALGRIND` is set in the environment--which is currently done in `modulemd/meson.build`.

The whole point of adding the python suppressions is so that we *should* be treating new failures as fatal. I'm not sure why this option is useful, much less enabled by default.

> The big question is... what should be included in the suppression file? For starters, I included a few from the `valgrind-python.supp` file included in the Python 2 and 3 installs that were prevalent when running the tests.

Short answer: we need to filter out anything that is coming from one of our dependencies (python itself, pyGObject, etc.). I see in your other comment that you are unsure whether those four remaining suppressions would be correct; we need to investigate that further.

That said, I'd be satisfied with being overzealous and including them if only because it would still allow us to have a baseline to notice if anything new appeared.

> Also, am I allowed to include `contrib/valgrind/grindmerge` here? It's a handy script I found for extracting and merging suppression entries from the test logfile.

The license header says nothing about permission to modify it (it only says "used and copied"), so I'm not sure whether including it in the upstream sources would be acceptable.

sgallagher:
> FWIW, when including the following four very specific valgrind suppression entries (in addition to the three general entries already included), the valgrind test has no complaints. But I haven't been able to figure out if they're real problems or not.
> 
> ```
> {
>    <insert_a_suppression_name_here>
>    Memcheck:Leak
>    match-leak-kinds: definite
>    fun:malloc
>    fun:g_malloc
>    fun:g_slice_alloc
>    fun:g_hash_table_new_full
>    fun:modulemd_yaml_parse_string_set
>    fun:modulemd_module_stream_v1_parse_licenses
>    fun:modulemd_module_stream_v1_parse_yaml
>    fun:modulemd_module_stream_read_yaml
>    fun:modulemd_module_stream_read_string
>    fun:ffi_call_unix64
>    fun:ffi_call
>    obj:/usr/lib64/python3.7/site-packages/gi/_gi.cpython-37m-x86_64-linux-gnu.so
> }

This one might be real (though it might be in pyGObject rather than libmodulemd). I'll take a look.

I'm comfortable with suppressing the other three, since the traceback involved doesn't include libmodulemd, pyGObject or FFI in any way, so it's almost certainly not our fault. As long as we aren't adding any *new* leaks going forward, that will still be a win.

sgallagher:
@mmathesius Yup, that first one was indeed real. See #443 

The other three are safe to suppress.

modulemd/tests/test-valgrind.py
```diff
@@ -41,8 +47,10 @@ def exec_valgrind(test):
             "/usr/bin/valgrind "
             "--leak-check=full "
             "--suppressions=/usr/share/glib-2.0/valgrind/glib.supp "
+            "--suppressions=%s "
+            "--gen-suppressions=all "
```
sgallagher: I don't think we want to be generating suppressions for every run.

modulemd/meson.build
```diff
@@ -306,6 +306,7 @@ test_release_env.set ('TEST_DATA_PATH', meson.source_root() + '/modulemd/tests/t
 test_env = test_release_env
 test_env.set('G_DEBUG', 'fatal-warnings,fatal-criticals')
 test_env.set('G_MESSAGES_DEBUG', 'all')
+test_env.set('MMD_IGNORE_PYTHON_VALGRIND', 'True')
```
sgallagher: Please drop this environment variable; we want valgrind failures to be fatal.

modulemd/tests/test-valgrind.py
mmathesius: > I don't think we want to be generating suppressions for every run.

I don't think there's much additional overhead for valgrind, and the XML only gets dumped to the logfile when one of the fatal error types is identified by `test-valgrind.py`. (Which, by the way, is a bug I just fixed. Previously, once any of the valgrind tests failed, the XML for _all_ remaining tests was dumped to the log!) Nevertheless, I removed the line to generate the suppressions.

mmathesius:
> @mmathesius Yup, that first one was indeed real. See #443
> 
> The other three are safe to suppress.

@sgallagher Thanks for the feedback, and for tracking that real one down!

We can expect CI to fail with my latest push until #443 is merged.

PR #443, opened 2020-02-05
ModuleStreamV1: Fix memory leak

sgallagher:
We didn't catch this in the C tests because it only appears when the `data.license.content` field appears before the `data.license.module` field in the module metadata and only for ModuleStreamV1. In all of the input files used for V1 in the C tests, `data.license.content` was either omitted or came after `data.license.module`.


sgallagher:
Actually, for completeness' sake, I'm going to add a C test that validates this change as well.

PR #444, opened 2020-02-06
Assorted cleanups for 2.9.0

sgallagher:
Reconcile differences between upstream and Fedora/EPEL packaging.

### DIFF
--- a/.travis/openmandriva/travis-tasks.sh
+++ b/.travis/openmandriva/travis-tasks.sh
@@ -4,7 +4,7 @@
 set -e
 set -x
 
-COMMON_MESON_ARGS="-Dtest_dirty_git=false -Ddeveloper_build=false -Dskip_clang_tidy=false -Dwith_py2_overrides=false -Dwith_docs=false"
+COMMON_MESON_ARGS="-Dtest_dirty_git=false -Ddeveloper_build=false -Dskip_clang_tidy=false -Dwith_docs=false"
 
 pushd /builddir/
 
--- a/.travis/opensuse/travis-tasks.sh
+++ b/.travis/opensuse/travis-tasks.sh
@@ -15,6 +15,6 @@ meson --buildtype=debug \
       $COMMON_MESON_ARGS \
       travis
 
-ninja -C travis test
+MMD_SKIP_VALGRIND=TRUE ninja -C travis test
 
 popd #builddir
--- a/bindings/python/meson.build
+++ b/bindings/python/meson.build
@@ -21,20 +21,18 @@ print(overridedir)
 '''
 
 # Python 3
-if (get_option('with_py3_overrides'))
-  ret = run_command([python3, '-c', get_overridedir])
+ret = run_command([python3, '-c', get_overridedir])
 
-  if ret.returncode() != 0
-    error('Failed to determine Python 3 pygobject overridedir')
-  else
-    pygobject_override_dir = join_paths(get_option('libdir'), ret.stdout().strip())
-  endif
-
-  install_data('gi/overrides/Modulemd.py', install_dir: pygobject_override_dir)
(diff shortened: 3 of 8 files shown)


sgallagher:
I just pushed an additional patch adding two more valgrind suppressions that we missed somehow.

mmathesius:
> I just pushed an additional patch adding two more valgrind suppressions that we missed somehow.

Sorry about that. There seems to be some more. The containerized environment is catching other things.  And I completely forgot about python2. :-(

sgallagher:
> > I just pushed an additional patch adding two more valgrind suppressions that we missed somehow.
> 
> Sorry about that. There seems to be some more. The containerized environment is catching other things. And I completely forgot about python2. :-(

No problem. I've got it nearly sorted out now. We need to skip attempting to scan the python2 on Rawhide anyway, since it doesn't exist.

sgallagher:
OK, I reworked this patch series a fair bit. There are now three patches:

1. This patch changes the default for building python 2 overrides (and running python 2 tests) to be false, since python 2 is now EOL upstream. It also renames the configuration options to `-Dwith_py{23}`, dropping the `_overrides` suffix since it affects test running as well. This will need to be called out in the release notes for 2.9.0
2. This patch cleans up the valgrind suppression file to support Python 2 as well as Python 3.8 (on Rawhide) and other versions in the future.
3. This patch syncs the specfile from Fedora/EPEL with upstream so that we will be able to take advantage of Packit's release handling.

@mmathesius I think this is the last patch set before we release 2.9.0, do you agree?

mmathesius:
> @mmathesius I think this is the last patch set before we release 2.9.0, do you agree?

@sgallagher I just created #445 to add a man page for `modulemd-validator`--if you wanted to wait for me to sort that out so it can be included.

sgallagher:
Hmm, I see the valgrind complaints on OpenSUSE and Mageia... @Conan-Kudo do you have any idea why those platforms would be reporting different valgrind output than Fedora?

sgallagher:
@mmathesius The previous patch-set passed CI on the alternative distributions this time. I canceled it because I wanted to add one more patch to this PR to include descriptions for the meson_options.txt. Running `meson configure` in a build directory will now produce:
```
...
Project options:
  Option             Current Value Possible Values           Description                                                                                                                                                                 
  ------             ------------- ---------------           -----------                                                                                                                                                                 
  developer_build    true          [true, false]             Enables automatic code formatters and memory leak checks. This option should be set to FALSE for release builds.                                                            
  libmagic           enabled       [enabled, disabled, auto] Enables the use of libmagic to detect compression of YAML files.                                                                                                            
  python_name                                                The name of the Python 3 interpreter to use for generating Python bindings and running tests. If left blank, it defaults to the version of Python 3 being used to run meson.
  rpmio              enabled       [enabled, disabled, auto] Use the rpmio library to automatically decompress gzip, bzip2 and xz YAML streams.                                                                                          
  skip_clang_tidy    true          [true, false]             Do not use the clang-tidy utility for code formatting.                                                                                                                      
  skip_formatters    false         [true, false]             Do not do any automatic formatting of source code.                                                                                                                          
  skip_introspection false         [true, false]             Do not generate GObject Introspection data.                                                                                                                                 
  test_dirty_git     false         [true, false]             Check whether the automatic formatting tools have made changes. Used during CI to verify coding style compliance                                                            
  test_installed_lib false         [true, false]             Build only the test suite and run it against a copy of libmodulemd installed on the local system.                                                                           
  with_docs          true          [true, false]             Build API documentation.                                                                                                                                                    
  with_py2           false         [true, false]             Build Python 2 language bindings and run Python 2 tests                                                                                                                     
...
```

sgallagher:
Assuming CI passes, I think this is finally ready.

PR #445, opened 2020-02-06
Add modulemd-validator man page

### LINKED ISSUE #365: modulemd-validator: missing man page
User-visible binaries in `{,/usr}/{,s}bin` should have documentation associated with them, at the very least a man page.

This is currently not the case.

mmathesius:
Resolves #365.

I'm not sure if the packaging of the generated man page will work. I also suspect CI will blow up with missing dependencies for non-Fedora distributions.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -82,6 +82,9 @@ if with_docs
   endif
 endif
 
+with_manpages = get_option('with_manpages')
+help2man = find_program('help2man', required: with_manpages)
+
 
 # Check whether this version of glib has the GDate autoptr defined
 gdate_check = '''#include <glib.h>
--- a/meson.build
+++ b/meson.build
@@ -112,8 +112,8 @@ else
     python3 = pymod.find_installation()
 endif
 
-with_py2_overrides = get_option('with_py2_overrides')
-if with_py2_overrides
+with_py2 = get_option('with_py2')
+if with_py2
     python2 = pymod.find_installation('python2')
 else
     python2 = disabler()
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -9,15 +9,36 @@
 # For more information on the license, see COPYING.
 # For more information on free software, see <https://www.gnu.org/philosophy/free-sw.en.html>.
 
-option('developer_build', type : 'boolean', value : true)
-option('libmagic', type : 'feature', value : 'enabled')
-option('python_name', type : 'string')
-option('rpmio', type : 'feature', value : 'enabled')
-option('skip_clang_tidy', type : 'boolean', value : false)
-option('skip_formatters', type : 'boolean', value : false)
-option('skip_introspection', type : 'boolean', value : false)
-option('test_dirty_git', type : 'boolean', value : false)
-option('test_installed_lib', type : 'boolean', value : false)
-option('with_docs', type : 'boolean', value : true)
-option('with_py2', type : 'boolean', value : false)
+option('developer_build', type : 'boolean', value : true,
+       description : 'Enables automatic code formatters and memory leak checks. This option should be set to FALSE for release builds.')
+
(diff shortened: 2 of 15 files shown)


meson_options.txt
```diff
@@ -19,5 +19,6 @@ option('skip_introspection', type : 'boolean', value : false)
 option('test_dirty_git', type : 'boolean', value : false)
 option('test_installed_lib', type : 'boolean', value : false)
 option('with_docs', type : 'boolean', value : true)
+option('with_manpage', type : 'boolean', value : true)
```
sgallagher:
```suggestion
option('with_manpage', type : 'feature', value : 'auto')
```
If we make this a `feature` option, you can just use it in the `required:` argument to `find_program()` and it will Do The Right Thing. (In the case of 'auto', it will return the program or a `disabler()` object if it cannot be found. It can also be passed `'enabled'` or `'disabled'` to force the result as appropriate.

meson.build
```diff
@@ -82,6 +82,12 @@ if with_docs
   endif
 endif
 
+with_manpage = get_option('with_manpage')
+help2man = find_program('help2man', required: with_manpage)
+if not help2man.found()
+  help2man = disabler()
+endif
```
sgallagher: These three lines are not necessary if you make `with_manpage` a `feature` option instead of a boolean.

modulemd/meson.build
```diff
@@ -527,3 +527,18 @@ if with_docs
         install : true,
     )
 endif
+
+if help2man.found()
```
sgallagher: You don't need this `if` block. If `help2man` is a `disabler()` object, meson will automatically skip any target that uses it as a dependency.

mmathesius:
@sgallagher  Thanks for the feedback. I'll take care of that and sort through any missing dependencies shortly.

One other thing that came to mind regarding the spec file... Should the man page be moved into the main package alongside with the `modulemd-validator` executable (instead of the devel package)?

sgallagher:
> One other thing that came to mind regarding the spec file... Should the man page be moved into the main package alongside with the `modulemd-validator` executable (instead of the devel package)?

Oh, I completely missed that. Yes, the manpage belongs with the executable.

modulemd/meson.build
```diff
@@ -529,3 +529,16 @@ if with_docs
         install : true,
     )
 endif
+
+help2man_opts = [
+  '--no-info',
+  '--section=1',
+]
+custom_target(
```
jon-turney: I think this target needs to be under the with_manpages condition, as it can't be generated when help2man is not found.

sgallagher: @jon-turney

from [`meson.build`](https://github.com/fedora-modularity/libmodulemd/blob/master/meson.build#L86)
```
with_manpages = get_option('with_manpages')
help2man = find_program('help2man', required: with_manpages)
```

This means that if `with_manpages` is set to `enabled`, the `meson configure` step will fail with an error (since the program is required) if `help2man` cannot be located. If `with_manpages` is `auto` (the default), it means that it will be built if `help2man` happens to be available, but if it's not, the `help2man` variable will be a [`disabler()`](https://mesonbuild.com/Disabler.html#page-description) object. A `disabler()` object means that any target that requires it will simply be skipped. If `with_manpages` is `disabled`, a `disabler()` object is always returned.

jon-turney: Thanks.

But I'm not sure if that's quite right.   Don't you need to write `find_program('help2man', required: with_manpages, disabler: true)` to get a dsabler object?

I any event, this doesn't appear to be currently working as you expect: see https://travis-ci.org/jon-turney/meson-corpus-test/jobs/648623888.

Is this a meson regression (i.e. you've seen this work as expected for some meson version?).  Or a problem with the meson documentation?

PR #447, opened 2020-02-11
Clarify v2 stream specification

sgallagher:
Related to [BZ 1799036](https://bugzilla.redhat.com/show_bug.cgi?id=1799036)

Several fields (notably those that provide `N:S:V:C:A`) were marked as being "optional" when that was really only because the packager should not be setting them. 

Also separates the different spec sections with blank lines for readability.


yaml_specs/modulemd_packager_v2.yaml
```diff
@@ -36,8 +39,6 @@ data:
     # the runtime block so that it matches the stream the module was built
     # against.  Multiple builds result in multiple output modulemd files.
     # See below for an example.
-    # TODO: Provides, conflicts, obsoletes, recommends, etc.
-    #       Do we even need those?
```
Conan-Kudo: We are probably going to need these...

yaml_specs/modulemd_stream_v2.yaml
```diff
@@ -91,8 +126,6 @@ data:
     # the runtime block so that it matches the stream the module was built
     # against.  Multiple builds result in multiple output modulemd files.
     # See below for an example.
-    # TODO: Provides, conflicts, obsoletes, recommends, etc.
-    #       Do we even need those?
```
Conan-Kudo: As said above, this will likely become necessary soon. Not sure why we don't have it yet, actually...

sgallagher: I removed these from the specification because I didn't think *proposed* concepts belong in the spec (it confuses the reader). Those should be filed as tickets in the issue tracker instead.

yaml_specs/modulemd_packager_v2.yaml
```diff
@@ -3,15 +3,18 @@
...
 version: 2
 data:
     # A short summary describing the module, required
     summary: An example module
+
     # A verbose description of the module, required
     description: >-
         A module for the demonstration of the metadata format. Also,
         the obligatory lorem ipsum dolor sit amet goes right here.
+
```
mmathesius: While adding blank lines for clarity, add one between `license` and `dependencies` below.

yaml_specs/modulemd_stream_v2.yaml
Conan-Kudo: If you move them to the issue tracker, then this will look good to merge.

yaml_specs/modulemd_stream_v2.yaml
```diff
@@ -4,40 +4,66 @@ document: modulemd
...
-    # timestamp.  Module version defines upgrade path for the particular
-    # update stream.
+
+    # Module version, integer, cannot be negative
+    # Filled in by the buildsystem, using the VCS commit timestamp.  Module
+    # version defines upgrade path for the particular update stream.
+    #
+    # When included as input to the build system, the behavior is unspecified.
+    #
+    # Mandatory for module metadata in a yum/dnf repository.
```
mmathesius: I think this needs something noted about what the value `0` means. Since `version` is an integer, I think `0` is not allowed as a module version identifier, and gets treated treated the same as if it were unspecified.

yaml_specs/modulemd_stream_v2.yaml
```diff
@@ -4,40 +4,66 @@ document: modulemd
...
-    # Module context flag, optional
+
+    # Module context flag
     # The context flag serves to distinguish module builds with the
     # same name, stream and version and plays an important role in
     # future automatic module stream name expansion.
     # Filled in by the buildsystem.  A short hash of the module's name,
-    # stream, version and its expanded runtime dependencies.
+    # stream, version and its expanded runtime dependencies. The exact
+    # mechanism of generating the has is unspecified except that it must be
```
mmathesius:
```suggestion
    # mechanism for generating the hash is unspecified except that it must be
```

goosemania:
:+1:  Thanks for the clarification, @sgallagher !

yaml_specs/modulemd_stream_v2.yaml
```diff
@@ -4,40 +4,66 @@ document: modulemd
...
+    # stable.
+    # When included as input to the build system, the behavior is unspecified.
```
mmathesius: Please empty comment lines above and below for clarity as was done for `name`, `stream`, and `version`.

yaml_specs/modulemd_stream_v2.yaml
sgallagher: https://github.com/fedora-modularity/libmodulemd/issues/448

yaml_specs/modulemd_stream_v2.yaml
sgallagher: It's actually the other way around. Unspecified is the same as `version==0` which is a valid version. The special-case here is that if you input `0`, you won't see it in the output. I'll try to clear that up. I guess that means this field isn't *actually* mandatory, though.

yaml_specs/modulemd_stream_v2.yaml
```diff
@@ -150,198 +264,291 @@ data:
...
+
+                # buildafter:
                 # Use the "buildafter" value to specify that this component
                 # must be be ordered later than some other entries in this map.
                 # The values of this array come from the keys of this map and
                 # not the real component name to enable bootstrapping.
                 # Use of both buildafter and buildorder in the same document is
                 # prohibited, as they will conflict.
+                #
+                #
```
mmathesius: Remove extraneous empty line.

yaml_specs/modulemd_stream_v2.yaml
```diff
@@ -350,29 +557,49 @@ data:
...
             includedmodule:
+
+                # rationale:
                 # Why is this module included?
-                # Required
+                #
+                # Type: MANDATORY
                 rationale: Included in the stack, just because.
+
                 # Link to VCS repository that contains the modulemd file
```
mmathesius: Add `# repository:` line here.

yaml_specs/modulemd_stream_v2.yaml
```diff
@@ -350,29 +557,49 @@ data:
...
+                # Type: AUTOMATIC
                 repository: https://pagure.io/includedmodule.git
+
+                # ref:
                 # See the rpms ref.
+                #
+                # Type: AUTOMATIC
                 ref: somecoolbranchname
+
+                # buildorder
```
mmathesius:
```suggestion
                # buildorder:
```

yaml_specs/modulemd_stream_v2.yaml
```diff
@@ -381,24 +608,50 @@ data:
             - xxx-0:1-1.module_deadbeef.x86_64
             - xxx-0:1-1.module_deadbeef.i686
             - xyz-0:1-1.module_deadbeef.x86_64
+
         # The rpm-map exists to link checksums from repomd to specific
```
mmathesius: Add `# rpm-map:` here, and add other hard-coded keys throughout the rest of this block.

yaml_specs/modulemd_stream_v2.yaml
```diff
@@ -134,14 +231,31 @@ data:
...
         documentation: http://www.example.com/
-        # Upstream bug tracker, if it exists, optional
+
+        # tracker:
+        # Upstream bug tracker, if it exists
+        #
+        # Type: OPTIONAL
         tracker: http://www.example.com/
+
+    # profiles:
```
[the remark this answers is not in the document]
sgallagher: I'm going to defer to @contyk here. Petr, how much of this is accurate?

yaml_specs/modulemd_packager_v2.yaml
```diff
@@ -3,15 +3,18 @@
 # for a module stream. It is always a proper subset of a `document: modulemd`
 # of the same `version`.
 document: modulemd-packager
```
mmathesius: One more comment... Once the `modulemd_stream_v2.yaml` text is settled, it would be best to tediously replicate the applicable pieces here in `modulemd_packager_v2.yaml` as well.

yaml_specs/modulemd_stream_v2.yaml
```diff
@@ -1,50 +1,125 @@
...
+# preserved and provided in the output metadata produced by the build        #
+# system for inclusion into a repository.                                    #
+#                                                                            #
+# AUTOMATIC: Attributes of this type must be present in the the repository   #
+# metadata, but they may be left unspecified by the packager. In this case,  #
+# the build system is responsible for generating an appropriate value for    #
+# the attribute and including it in the repository metadata. If the packager #
+# specifies this attribute explicitly, it must be preserved and provided in  #
+# the output metadata for inclusion into a repository.                       #
+##############################################################################
```
contyk: There are cases where the build system overrides supplied values; for instance, MBS has APIs that let you override the NSVC or the dependencies. I'm not sure if the language here isn't too strict.

While we normally want to it preserved, the abovementioned behavior shouldn't be considered broken or as violating the standard, I'd say.

yaml_specs/modulemd_packager_v2.yaml
sgallagher: Yeah, it probably would. :sob:

yaml_specs/modulemd_stream_v2.yaml
sgallagher: @contyk 
Does the following line sound right if added at the end of the glossary?

"The definitions above describe the expected behavior of the build system operating in its default configuration. It is permissible for the build system to override user-provided entries through non-default operating modes. If such changes are made, all items indicated as being required for the output repository must still be present."

contyk: Yeah, that'd be fine.

yaml_specs/modulemd_stream_v2.yaml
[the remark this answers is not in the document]
contyk: Unfortunately, "default" has no special meaning nowadays. "buildroot" and "srpm-buildroot" are the only snowflakes we have. I'm not sure what youy mean by the component rpm flags.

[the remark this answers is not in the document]
sgallagher: @contyk He's asking if we should include a pointer to `data.components.rpms.<name>.[srpm]buildroot` boolean attributes. I think the answer is yes.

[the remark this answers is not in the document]
mmathesius: Yes, that's what I meant.

sgallagher:
OK, I updated both the `document: modulemd` and `document: modulemd-packager` specifications to keep them in sync. I think I cleaned up the language of the `buildroot` options to make their purpose more clear.

In the process of reworking them to make them clearer, I had to update one of the tests which also revealed a minor bug in the output. We were emitting `buildorder` to the YAML as a 64-bit unsigned integer instead of as a 64-bit *signed* integer. This caused the test to fail when I set `buildorder: -1` in one place (since it was comparing `-1` to `MAX_UINT64`)

Another round of review would be helpful, please.

sgallagher:
@mmathesius Changes made

sgallagher:
Merged in 80b253eebee9caab151dc7e5b482b5fd89a32da1

PR #449, opened 2020-02-11
Add ModuleStreamV2.search_profiles()

sgallagher:

attn: @kontura

### DIFF
--- a/modulemd/modulemd-module-stream-v2.c
+++ b/modulemd/modulemd-module-stream-v2.c
@@ -734,6 +734,52 @@ modulemd_module_stream_v2_get_profile (ModulemdModuleStreamV2 *self,
 }
 
 
+struct profile_match_ctx
+{
+  GHashTable *profiles;
+  GPtrArray *found;
+  const gchar *pattern;
+};
+
+static void
+profile_match (gpointer data, gpointer user_data)
+{
+  struct profile_match_ctx *match_ctx = (struct profile_match_ctx *)user_data;
+
+  /* Add it to the found list if it matches the pattern */
+  if (modulemd_fnmatch (match_ctx->pattern, (const gchar *)data))
+    {
+      g_ptr_array_add (match_ctx->found,
+                       g_object_ref (g_hash_table_lookup (
+                         match_ctx->profiles, (const gchar *)data)));
+    }
+}
+
+GPtrArray *
+modulemd_module_stream_v2_search_profiles (ModulemdModuleStreamV2 *self,
+                                           const gchar *profile_pattern)
+{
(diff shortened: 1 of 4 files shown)


modulemd/tests/ModulemdTests/modulestream.py
```diff
@@ -1311,6 +1311,93 @@ def test_xmd_issue_290(self):
...
+        profiles = stream.search_profiles("*")
+        self.assertIsNotNone(profiles)
+        self.assertEqual(len(profiles), 0)
+
+        profiles = stream.search_profiles("thefirstprofile")
+        self.assertIsNotNone(profiles)
+        self.assertEqual(len(profiles), 0)
+
+        # Now add three profiles, conveniently alphabetical
+        stream.add_profile(Modulemd.Profile.new("thefirstprofile"))
```
mmathesius: Might I suggest adding the profiles _not_ in alphabetical order--to demonstrate that `search_profiles()` sorts the returned list as documented.

modulemd/modulemd-module-stream-v2.c
```diff
@@ -734,6 +734,47 @@ modulemd_module_stream_v2_get_profile (ModulemdModuleStreamV2 *self,
...
+    }
+}
+
+GPtrArray *
+modulemd_module_stream_v2_search_profiles (ModulemdModuleStreamV2 *self,
+                                           const gchar *profile_pattern)
+{
+  /* We will commonly only match a single profile, so allocate for that */
+  GPtrArray *found = g_ptr_array_new_full (1, g_object_unref);
+
```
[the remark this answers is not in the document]
sgallagher: > I[t] could be argued if the common use case will match only a single profile, or all profiles.

Hmm, given that it's unlikely for the number of profiles to grow very large, you're probably right that it's better to assume it matches all of them.

> Also, might it be a good idea to add `g_return_val_if_fail (MODULEMD_IS_MODULE_STREAM_V2 (self), found);` here? Or is it better to let it (hopefully) blow up with an exception?

No, you're right. I forgot to add it. Since the function description says "this function cannot fail", we should return the empty list on a programming error. (It will assert in debug builds.)

modulemd/tests/ModulemdTests/modulestream.py
sgallagher: Yes, that's a good idea. I'll do that.

sgallagher:
@mmathesius Thanks for the feedback. I just pushed the requested fixes, along with another patch to temporarily allow Rawhide to fail in CI because we just crossed the branch point and there are no Fedora 33 container base images yet.

kontura:
Great, thank you for adding this.

PR #450, opened 2020-02-11
Add Release script

sgallagher:
Add a script to perform upstream releases.

Also includes a new test to ensure that we update the version in `meson.build` after a release to avoid the occasional issue where CI ends up performing the installed-library tests against the Fedora version rather than the upstream version.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -2,6 +2,8 @@ specfile_path: libmodulemd.spec
 upstream_package_name: libmodulemd
 downstream_package_name: libmodulemd
 upstream_tag_template: libmodulemd-{version}
+current_version_command:
+  - ./.packit_version.sh
 
 synced_files:
   - .packit.yml
--- a/modulemd/tests/test-version.sh
+++ b/modulemd/tests/test-version.sh
@@ -1,0 +1,36 @@
+#!/usr/bin/bash
+
+set -x
+
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+pushd $MESON_BUILD_ROOT
+
+function error_out {
+    local code message
+    local "${@}"
+
+    echo $message 1>&2
+    exit $code
+}
+
+function common_finalize {
+    exitcode=$?
+
+    return $exitcode
+}
+
+trap common_finalize EXIT
+
+jq --version >/dev/null || error_out code=127 message="Install 'jq' to use this script"
+
+# Get the previous tag for this branch
+OLDTAG=$(git describe --first-parent --abbrev=0)
+
+# Get the version that will be tagged
+NEWVERSION=$(meson introspect $MESON_BUILD_ROOT --projectinfo |jq -r .version)
+NEWTAG=libmodulemd-$NEWVERSION
+
+if [ $NEWTAG = $OLDTAG ]; then
+    error_out code=2 message="Version is already tagged. Update meson.build with the new version."
+fi
+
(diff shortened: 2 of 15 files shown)


sgallagher:
I added a couple new patches to rework how we're generating the `Release:` field for the Packit builds. I think this will be far better now.

.packit.yml
```diff
@@ -2,6 +2,8 @@ specfile_path: libmodulemd.spec
 upstream_package_name: libmodulemd
 downstream_package_name: libmodulemd
 upstream_tag_template: libmodulemd-{version}
+current_version_command:
+  - git describe --tags --match "*.*"  --abbrev=0 |perl -lne 'print "$1\n" if /^libmodulemd-(.*)/'
```
mmathesius:
```suggestion
  - git describe --tags --match "*.*"  --abbrev=0 | sed -ne 's/^libmodulemd-//p'
```
I love `perl` as much as the next person, but `sed` always seems more appropriate in a shell script.

modulemd/tests/test-version.sh
```diff
@@ -0,0 +1,36 @@
...
+
+function common_finalize {
+    exitcode=$?
+
+    return $exitcode
+}
+
+trap common_finalize EXIT
+
+jq --version >/dev/null || error_out code=127 message="Install 'jq' to use this script"
```
mmathesius: As you've probably noticed already, `jq` needs to be installed during the container build process. And probably added to the SPEC file.

.packit.yml
sgallagher: Huh... I was trying to make `sed` work here and it just wasn't coming together for me. I found the `perl` form on the web.

I agree, I'll use the `sed` form. (Better for dependencies too)

modulemd/tests/test-version.sh
sgallagher: Yeah, saw that in the CI logs. Oops.

sgallagher: It doesn't need to be added to the spec file because that builds with `-Ddeveloper_build=false`, suppressing this test.

mmathesius: Ah. I overlooked the non-developer build from the SPEC.

mmathesius:
> Also, I simply cannot get the `git describe ...` commands to work as expected on my workstation. The only tags shown to me are `libmodulemd-0.1.0`, `libmodulemd-0.2.0`, and `libmodulemd-0.2.1` which is very strange. I'm still trying to figure that out so I can carefully check out the scripts locally.

I figured this one out. My github fork of libmodulemd (and yours) only had info for a few of the original tags. The master repository has all of them. (And I just updated my fork with the latest tags.)

mmathesius:
I did some experimenting to sort out the packit error. It doesn't seem to like anything other than a single token for `current_version_command`. The following patch worked for me to be able to run `packit srpm` locally:

```
diff --git a/.packit.yml b/.packit.yml
index fba2d4a..7514322 100644
--- a/.packit.yml
+++ b/.packit.yml
@@ -3,7 +3,7 @@ upstream_package_name: libmodulemd
 downstream_package_name: libmodulemd
 upstream_tag_template: libmodulemd-{version}
 current_version_command:
-  - git describe --tags --match "*.*"  --abbrev=0 | sed -ne 's/^libmodulemd-//p'
+  - ./.packit_version.sh
 
 synced_files:
   - .packit.yml
diff --git a/.packit_version.sh b/.packit_version.sh
new file mode 100755
index 0000000..d1b8201
--- /dev/null
+++ b/.packit_version.sh
@@ -0,0 +1,2 @@
+#!/bin/sh
+git describe --tags --match '*.*'  --abbrev=0 | sed -ne 's/^libmodulemd-//p'
```

sgallagher:
@mmathesius Thanks for the packit fix. I rolled it in and now both the Travis and Packit tests appear to be working!

PR #452, opened 2020-02-13
Release script enhancements

sgallagher:
Two patches to improve `release.sh`

The first patch moves `release.sh` into the `contrib/release-tools` directory and out of the root. It also adds the ability to have `meson` update the version after tagging the current state for the release. They are then pushed together to avoid CI complaining in the new [version test](https://github.com/fedora-modularity/libmodulemd/blob/master/modulemd/tests/test-version.sh).

The second patch changes the release.sh to require a remote named `upstream`. Previously, we were pushing to whatever remote happened to be the default for the branch we were on.

### DIFF
--- a/contrib/release-tools/release.sh
+++ b/contrib/release-tools/release.sh
@@ -24,6 +24,9 @@ function common_finalize {
 
 trap common_finalize EXIT
 
+git remote get-url upstream |grep "fedora-modularity/libmodulemd.git\$" -q || error_out code=127 message="There must be a git remote named 'upstream' available. The user running this script must have privilege to push commits to the 'upstream' remote"
+[ $(git branch --show-current) = master ] || error_out code=127 message="This script may only be run on the 'master' branch"
+
 hub --version >/dev/null || error_out code=127 message="Install 'hub' to use this script"
 jq --version >/dev/null || error_out code=127 message="Install 'jq' to use this script"
 
@@ -68,7 +71,7 @@ if [ $($SCRIPT_DIR/semver compare 0.50.0 $(meson --version)) = -1 ]; then
 fi
 
 # Make sure everything is up-to-date on Github
-git push --follow-tags || error_out code=5 message="Couldn't push the new tags to Github"
+git push --follow-tags upstream master || error_out code=5 message="Couldn't push the new tags to Github"
 
 echo "libmodulemd $NEWVERSION" > $TMPDIR/github_message
 echo >> $TMPDIR/github_message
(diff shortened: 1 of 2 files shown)


contrib/release-tools/release.sh
```diff
@@ -23,6 +24,9 @@ function common_finalize {
 
 trap common_finalize EXIT
 
+git remote |grep "^upstream\$" -q || error_out code=127 message="There must be a git remote named 'upstream' available. The user running this script must have privilege to push commits to the 'upstream' remote"
```
mmathesius:
```suggestion
git remote get-url upstream >/dev/null || error_out code=127 message="There must be a git remote named 'upstream' available. The user running this script must have privilege to push commits to the 'upstream' remote"
```
Optional. A little more direct.

contrib/release-tools/release.sh
```diff
@@ -23,6 +24,9 @@ function common_finalize {
...
+git branch --show-current |grep "^master\$" -q || error_out code=127 message="This script may only be run on the 'master' branch"
```
mmathesius:
```suggestion
[ $(git branch --show-current) = master ] || error_out code=127 message="This script may only be run on the 'master' branch"
```
Optional. Also a little more direct.

contrib/release-tools/release.sh
sgallagher: Ah, I didn't know about that. In that case, we can be more strict here and verify that it contains the string `fedora-modularity/libmodulemd.git` as well, to be sure that the remote is correct.

contrib/release-tools/release.sh
sgallagher: I didn't know about that syntax. Handy!

sgallagher:
@mmathesius 
I made the requested changes and also added one more patch to have the `release.sh` script refuse to work from a repo with uncommitted changes.

contrib/release-tools/release.sh
```diff
@@ -23,6 +24,12 @@ function common_finalize {
 
 trap common_finalize EXIT
 
+# Don't release if the git repo has uncommitted changes
+git diff --quiet || error_out code=127 message="There are uncommitted changes. Refusing to release."
+
+git remote get-url upstream |grep "fedora-modularity/libmodulemd.git\$" -q || error_out code=127 message="There must be a git remote named 'upstream' available. The user running this script must have privilege to push commits to the 'upstream' remote"
```
mmathesius:
```suggestion
[[ $(git remote get-url upstream) =~ fedora-modularity/libmodulemd.git$ ]] || error_out code=127 message="There must be a git remote named 'upstream' available. The user running this script must have privilege to push commits to the 'upstream' remote"
```
As lone as you're using `bash`, you can do the regex check right in the shell. (This is optional, of course.)

PR #454, opened 2020-02-13
Drop clang-tidy formatter

sgallagher:
It's unreliable and segfaults in CI frequently. Disable it.

### DIFF
--- a/modulemd/meson.build
+++ b/modulemd/meson.build
@@ -14,7 +14,6 @@
 developer_build = get_option('developer_build')
 test_dirty_git = get_option('test_dirty_git')
 test_installed_lib = get_option('test_installed_lib')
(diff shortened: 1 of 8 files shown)


modulemd/meson.build
```diff
@@ -419,15 +385,6 @@ clang_args = [ '-i' ]
 test('clang_format', clang_format,
      args : clang_args + clang_files)
```
mmathesius:
```suggestion
     args : clang_format_args + clang_files)
```
Completely optional: it would be more consistent with the style of the rest of this file if `clang_args` was renamed to `clang_format_args` here and two lines above--but there's no good reason to change it.

sgallagher: I'm not too worried about that, but thanks for the thorough review!

PR #456, opened 2020-02-14
Don't run CI on tags

sgallagher:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: c
 compiler: gcc
 sudo: required
 
+# Don't run Travis against tags, just new builds.
+# Tagged builds will always fail the test for whether we remembered to bump
+# the version number after a release.
+if: tag IS blank
+
 services:
   - docker
 


sgallagher:
> I wondered where those Travis failures where coming from!
> 
> I'm good with this.
> 
> But I'm also wondering if there's a way to have the builds still run, but skip the version test. Something like having Travis set an environment variable such as `MMD_SKIP_VERSION_CHECK` and then have `test-version.sh` exit with 77 if it is set. (I can't seem to tell if Travis can conditionally set an environment variable like that.)

I considered that, but I rejected it for the simple reason that a tagged build should always be a duplicate anyway. It seemed a bit wasteful to run it twice. Though, now that the `release.sh` automatically pushes a version bump to master, I suppose the CI has never been run against that *exact* tag...

If you feel strongly about it, it's not too hard to add. Have a look at [`travis-coverity.sh`](https://github.com/fedora-modularity/libmodulemd/blob/master/.travis/travis-coverity.sh#L60) for how we pass the `COVERITY_SCAN_TOKEN`.

mmathesius:
> If you feel strongly about it, it's not too hard to add. Have a look at [`travis-coverity.sh`](https://github.com/fedora-modularity/libmodulemd/blob/master/.travis/travis-coverity.sh#L60) for how we pass the `COVERITY_SCAN_TOKEN`.

No strong feelings about it. It was mostly a curiosity. I had somehow missed the documentation on all the [Travis environment variables](https://docs.travis-ci.com/user/environment-variables/#default-environment-variables)--particularly `TRAVIS_TAG`. :confused: Now it makes more sense to me!

PR #464, opened 2020-03-27
additional valgrind suppression for Python 3.9

mmathesius:
Fixes BZ#1817665


sgallagher:
That suppression looks *really* broad. I’m concerned that we may actually have noticed a real memory leak in Python 3.

vstinner:
By the way, if you are curious: here is an article telling the rationale for not deleting the GIL at exit ;-) https://vstinner.github.io/daemon-threads-python-finalization-python32.html

PR #467, opened 2020-04-05
build: Make pygobject override_dir configurable

jtojnar:
On systems that use different prefixes for each package like Nix or Guix, determining the path to install the PyGObject overrides will fail. Let's add a build option for setting it manually.

### DIFF
--- a/bindings/python/meson.build
+++ b/bindings/python/meson.build
@@ -1,3 +1,6 @@
+pygobject_override_dir = get_option('pygobject_override_dir')
+pygobject2_override_dir = get_option('py2_pygobject_override_dir')
+
 get_overridedir = '''
 import os
 import sysconfig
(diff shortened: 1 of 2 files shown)


sgallagher:
@jtojnar I'd prefer not to make these paths configurable, since they are tightly bound to the version of python installed and they can always be auto-detected.

That said, our method of auto-detection was rather excessively picky about where it considered a "correct" overrides directory. Since the value of `gi._overridesdir` must always be correct (or GObject Introspection cannot work at all), we'll rely on that.

The older auto-detection was compatible with pygobject 2.X, but as libmodulemd doesn't support anything that old, it was unnecessary.

Please see (and test) #469 for my proposed alternative approach.

jtojnar:
That will not work for us since `gi._overridesdir` will be something like `/nix/store/g5142vy7c74iccsy7gp3qxnqpg0c7nlb-python3.7-pygobject-3.36.0/lib/python3.7/site-packages/gi/overrides/` which is immutable so we cannot install to it. We need to install it under this package’s own prefix like `/nix/store/p7gzfyhygv2lrrv352xgcv58lmy79nbf-libmodulemd-2.9.2/lib/python3.7/site-packages/gi/overrides/`.

pkg-config supports this use case superbly where we can just do `pkg-config pygobject-3.0 --variable=overridesdir --define-variable=prefix=$prefix` (see https://www.bassi.io/articles/2018/03/15/pkg-config-and-paths/) but unfortunately, pygobject does not support that.

I do not really see a way to allow this without making the path configurable at build time, or changing pygobject to allow overriding the prefix.

sgallagher:
Does Nix do something special where pygobject3 loads from multiple locations? Because the whole point of the `gi._overridesdir` is to indicate the path from which it would load overrides.

My guess is that, even if you install into `/nix/store/p7gzfyhygv2lrrv352xgcv58lmy79nbf-libmodulemd-2.9.2/lib/python3.7/site-packages/gi/overrides/`, pygobject3 won't actually see it (and therefore won't apply overrides).

My suspicion is that pygobject3 is packaged incorrectly in Nix, because the overridesdir must always be writable by other packages.

With your patch applied to specify the overrides directory, can you try running the following command: `python3 -c "import gi; from gi.repository import Modulemd; assert('_overrides_module' in dir(Modulemd))"`. If that succeeds, then pygobject3 can read the overrides content from the libmodulemd-specific path. If it `assert()`s, then there's an issue with pygobject3's packaging that you will need to resolve on their end.

jtojnar:
I believe `gi._overridesdir` is just for people that want to install things but is not used by pygobject itself. Overrides are plain old Python modules, just starting with `gi.overrides.` prefix, so installing them to path that gets added to `PYTHONPATH` will suffice.

```ShellSession
$ nix-shell -p gobject-introspection 'python3.withPackages (p: with p; [pygobject3 six (toPythonModule libmodulemd).py])' --run "python3 -c \"import gi; from gi.repository import Modulemd; assert('_overrides_module' in dir(Modulemd))\""
-c:1: PyGIWarning: Modulemd was imported without specifying a version first. Use gi.require_version('Modulemd', '2.0') before import to ensure that the right version gets loaded.
```

sgallagher:
OK, that is helpful at least. I'll tweak my PR to also allow overriding by meson option. Thanks for the explanation and the investigation. I'm closing this PR and I'll add the appropriate options on #469

PR #469, opened 2020-04-07
Make python overrides path detection more robust and allow setting them explicitly

sgallagher:
The original auto-detection was copied from another project and
was compatible with very old (<3.0) pygobject. Since we aren't
supporting any version that old, we can simplify the detection
and make it work for systems that don't necessarily use the
/usr/lib[64]/pythonX.Y/gi/overrides path, such as NixOS and
Guix.

Closes #467

### DIFF
--- a/bindings/python/meson.build
+++ b/bindings/python/meson.build
@@ -1,44 +1,23 @@
-get_overridedir = '''
-import os
-import sysconfig
-
-libdir = sysconfig.get_config_var('SCRIPTDIR')
-
-if not libdir:
-  libdir = '/usr/lib'
-
-try:
-  import gi
-  overridedir = gi._overridesdir
-except ImportError:
-  purelibdir = sysconfig.get_path('purelib')
-  overridedir = os.path.join(purelibdir, 'gi', 'overrides')
-
-if overridedir.startswith(libdir): # Should always be True..
-  overridedir = overridedir[len(libdir) + 1:]
-
-print(overridedir)
-'''
-
 # Python 3
-ret = run_command([python3, '-c', get_overridedir])
+if gobject_overrides_dir_py3 == ''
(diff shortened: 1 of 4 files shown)


sgallagher:
@jtojnar Would you mind reviewing these patches and see if they work for you? I changed the option format from your #467 to make it more future-compatible (we might need overrides for non-python languages eventually and I want to have consistent option names).

bindings/python/meson.build
```diff
@@ -1,45 +1,37 @@
...
-except ImportError:
-  purelibdir = sysconfig.get_path('purelib')
-  overridedir = os.path.join(purelibdir, 'gi', 'overrides')
-
-if overridedir.startswith(libdir): # Should always be True..
-  overridedir = overridedir[len(libdir) + 1:]
+import gi
+overridedir = gi._overridesdir
 
 print(overridedir)
```
jtojnar, line 5: Perhaps just print it directly?

jtojnar:
One thing that might make sense is to allow disabling Python 3 overrides as well to make it symmetric.

bindings/python/meson.build
sgallagher: Yeah, probably not a bad idea. Just a leftover from the previous approach.

sgallagher:
> One thing that might make sense is to allow disabling Python 3 overrides as well to make it symmetric.

I considered that, but I cannot really come up with a scenario where we wouldn't want Python 3 support to be built. We require python 3 on the system to build (meson uses it).

I'm going to leave it unconditional until and unless a use-case turns up.

jtojnar:
Yeah, I consider it pretty unfortunate that Meson leaks its Python to projects, especially since one of its goals was not to leak its implementation to allow third party implementations of the language. But I guess it is inevitable on systems where most of packages are installed to single global hierarchy.

PR #475, opened 2020-06-15
Drop custom valgrind test

sgallagher:
Instead of running valgrind as a test case, make it a specialized
CI task.

This patch also adds "suite" tags to each of the tests, so they
can be run individually.


.travis/fedora/travis-tasks.sh
```diff
@@ -10,12 +10,51 @@ RETRY_CMD=/builddir/.travis/retry-command.sh
...
+
 # Build the code under GCC and run standard tests
 meson --buildtype=debug \
       $COMMON_MESON_ARGS \
       travis
 
-meson test -C travis --num-processes=$PROCESSORS --print-errorlogs -t 5
+meson test --suite smoketest \
+           -C travis \
+           --num-processes=$PROCESSORS \
```
mmathesius, line 28: Should `$COMMON_MESON_ARGS` also be included for each of the `meson test ...` runs? Although, the way things were re-worked by this PR, the dirty test flag set by `$COMMON_MESON_ARGS` is superfluous everywhere except for the `dirty` suite below--in which case `-Dtest_dirty_git=true` could be hard coded since the shell conditional has already determined if it should be run or not.

.travis/mageia/travis-tasks.sh
```diff
@@ -4,7 +4,7 @@
 set -e
 set -x
 
-
+PROCESSORS=$(/usr/bin/getconf _NPROCESSORS_ONLN)
 COMMON_MESON_ARGS="-Dtest_dirty_git=${DIRTY_REPO_CHECK:-true} -Ddeveloper_build=false"
```
mmathesius, line 8: Related to my comments for `.travis/fedora/travis-tasks.sh`, should `$COMMON_MESON_ARGS` be carried through to each of the `meson test ...` runs? Or should it just be dropped. It appears that for mageia, the dirty git tests are never run, and `developer_build` enables running of the `smoketest` and `formatters` suites--which are explicitly called out below.

.travis/fedora/travis-tasks.sh
sgallagher: No, the meson args are only for setting up the buildroot. The `meson test` calls wouldn't know what to do with those flags.

I hard-coded the `-Dtest_dirty_git=true` in the meson buildroot command.

.travis/mageia/travis-tasks.sh
sgallagher: I dropped the formatter suite but left the smoketest suite (since it still runs the header-import test).

I'm not actually sure why Mageia wasn't running the dirty test. I'm adding the dirty test to the 'ci' suite so it doesn't need a separate entry in the `travis-tasks.sh` files. Hopefully that will force it to work on Mageia too.

sgallagher:
I updated `README.md` as requested. I also tweaked the suites a bit more so that the `smoketest` and `dirty` suites are part of the `ci` suite, which reduced the duplication in `travis-tasks.sh`.

README.md
```diff
@@ -329,14 +329,19 @@ cause the application to `abort()` on programming errors that would be logged
...
 
-If you are trying to iterate quickly, you can temporarily skip the valgrind
-memory tests by running the test suite with:
 ```
-MMD_SKIP_VALGRIND=True ninja test
+meson test --suite=ci_valgrind \
+  --wrap="valgrind \
+          --leak-check=full \
+          --suppressions=/usr/share/glib-2.0/valgrind/glib.supp \
+          --suppressions=$SRCDIR/contrib/valgrind/libmodulemd-python.supp
```
mmathesius, line 339:
```suggestion
          --suppressions=../contrib/valgrind/libmodulemd-python.supp"
```

sgallagher: I tweaked that section a little to make the path clearer.

README.md
```diff
@@ -329,14 +329,21 @@ cause the application to `abort()` on programming errors that would be logged
...
-memory tests by running the test suite with:
+### Running tests with valgrind
+Assuming your current working directory is `debugbuild` as described above:
 ```
-MMD_SKIP_VALGRIND=True ninja test
+meson test --suite=ci_valgrind \
+  --wrap="valgrind \
+          --leak-check=full \
+          --suppressions=/usr/share/glib-2.0/valgrind/glib.supp \
+          --suppressions=../contrib/valgrind/libmodulemd-python.supp
```
mmathesius, line 339:
```suggestion
          --suppressions=../contrib/valgrind/libmodulemd-python.supp"
```

The closing double quote went missing again. Assuming CI passes, everything else with this PR looks great! Thanks!

sgallagher:
Merged

PR #476, opened 2020-06-23
Add new modulemd-eol

kontura:
This adds a new document type according to: https://fedoraproject.org/wiki/Changes/Module_Obsoletes_and_EOL

I tried to guess what API dnf might want for this, but we will see later, there might be changes needed. 

Couple of things I think could be problematic:
 - I made a separate commit that converts the date ISO 8601 format (specified for the modulemd-eol yaml) into the internal guint64 representation, but the API still uses guint64 directly.
 - The EOLs are stored in modulemd-module in a hash with keys such as: `stream:context`, these keys are kind of awkward, maybe a GPtrArray through which we would search could be better?


modulemd/include/modulemd-2.0/modulemd-eol.h
```diff
@@ -0,0 +1,224 @@
...
+ *
+ * Since: 2.10
+ */
+const gchar *
+modulemd_eol_get_module_context (ModulemdEol *self);
+
+/**
+ * modulemd_eol_set_module_context:
+ * @self: This #ModulemdEol object.
+ * @context: The name of the module context to which this eol apply.
```
sgallagher, line 116:
```suggestion
 * @context: The name of the module context to which this eol applies.
```

modulemd/modulemd-docs.xml
```diff
@@ -79,6 +80,7 @@
        <xi:include href="xml/modulemd-subdocument-info-private.xml"/>
        <xi:include href="xml/modulemd-translation-private.xml"/>
        <xi:include href="xml/modulemd-translation-entry-private.xml"/>
+       <xi:include href="xml/modulemd-eol-private.xml"/>
```
sgallagher, line 83: Please also add `xml/api-index-2.10.xml` below, or these functions will not appear in the index.

modulemd/include/modulemd-2.0/modulemd-eol.h
sgallagher: Also, it appears that the code is just fine with `context` being `NULL`, so you need to add `(nullable)` to the docstring here.

modulemd/include/modulemd-2.0/modulemd-eol.h
```diff
@@ -0,0 +1,224 @@
...
+ *
+ * Since: 2.10
+ */
+gboolean
+modulemd_eol_get_reset (ModulemdEol *self);
+
+/**
+ * modulemd_eol_set_eol_date:
+ * @self: This #ModulemdEol object.
+ * @eol_date: The end-of-life date for this stream. If not specified, the stream is EOLed immediately.
```
sgallagher, line 159:
```suggestion
 * @eol_date: The end-of-life date for this stream. If set to zero, the stream is EOLed immediately.
```

modulemd/include/modulemd-2.0/modulemd-module-index-merger.h
```diff
@@ -103,6 +103,22 @@ G_BEGIN_DECLS
...
+ *   #ModulemdModuleIndex will be the eol object in the resulting merged
+ *   #ModulemdModuleIndex.
+ * - If the #ModulemdModuleIndex inputs have different priorities (not common),
+ *   then the eol from the highest priority #ModulemdModuleIndex will be
+ *   used and the others entirely discarded. The `modified` value will not be
+ *   considered at all. (Priority is intended for providing a total override,
+ *   including an on-disk configuration).
+ * - If the repos have the same priority (such as "fedora" and "updates" in the
+ *   Fedora Project) newer eol (by modifed value) is used.
+ * - If the repos have the same priority and modifed value the behavior is
```
sgallagher, line 120:
```suggestion
 * - If the repos have the same priority and modifed value, the behavior is
```

modulemd/modulemd-module.c
```diff
@@ -751,6 +767,67 @@ modulemd_module_get_translation (ModulemdModule *self, const gchar *stream)
...
+            {
+              continue;
+            }
+          modulemd_module_stream_associate_eol (stream, neweol);
+        }
+      else
+        {
+          modulemd_module_stream_associate_eol (stream, neweol);
+        }
+    }
```
sgallagher, line 827: I think you need to define the behavior for a mixture of documents that do or do not have context specified.

For example, as written right now, the result of having the following two `modulemd-eol` documents will behave differently depending on the order they appear in the modulemd YAML stream:

```yaml
document: modulemd-eol
version: 1
data:
  name: nodejs
  stream: 10
  context: deadbeef
  obsoleted_by:
    module: nodejs
    stream: 12
```
vs.
```yaml
document: modulemd-eol
version: 1
data:
  name: nodejs
  stream: 10
  obsoleted_by:
    module: nodejs
    stream: 14
```

Read in the order above, the result would be that `nodejs:10` was obsoleted by `nodejs:14` (because the no-context document would overwrite the context-including one that was read earlier).

Reverse the order and the result would be that `nodejs:10:deadbeef` is obsoleted by `nodejs:12` while any other context of `nodejs:10` will be obsoleted by `nodejs:14`.

Please figure out how to do this without ordering differences.

kontura: Thanks for the batch of feedback, I have pushed an update.

I have also added a modified date check, so that the associated eol is the newest one. If there are identical dates it prefers the more specific (with context) eol.
This way the ordering problem still happens if there are conflicting eols with the same modified date, module, stream and context values, so I have added `g_info` message for this case.

I will also add this logic to merging indexes.

kontura: Added the logic to merging indexes.

modulemd/modulemd-eol.c
```diff
@@ -0,0 +1,1196 @@
...
+}
+
+
+void
+modulemd_eol_set_reset (ModulemdEol *self, gboolean reset)
+{
+  g_return_if_fail (MODULEMD_IS_EOL (self));
+  g_return_if_fail (!(reset && self->obsoleted_by_module_name));
+  g_return_if_fail (!(reset && self->obsoleted_by_module_stream));
+  g_return_if_fail (!(reset && self->eol_date));
```
sgallagher, line 322: `g_return_if_fail()` is intended to catch coding errors (it calls `g_abort()` in the debug profile). Don't perform such checks in the set function anyway; the reason we have the validate function is to catch exactly these sorts of issues once all the data is loaded.

modulemd/modulemd-eol.c
```diff
@@ -0,0 +1,1196 @@
...
+  self->module_context = g_strdup (module_context);
+
+  g_object_notify_by_pspec (G_OBJECT (self), properties[PROP_MODULE_CONTEXT]);
+}
+
+void
+modulemd_eol_set_eol_date (ModulemdEol *self, guint64 eol_date)
+{
+  g_return_if_fail (MODULEMD_IS_EOL (self));
+  g_return_if_fail (!(eol_date && self->reset));
```
sgallagher, line 374: Remove this

modulemd/modulemd-eol.c
```diff
@@ -0,0 +1,1196 @@
...
+  g_object_notify_by_pspec (G_OBJECT (self), properties[PROP_EOL_DATE]);
+}
+
+
+void
+modulemd_eol_set_message (ModulemdEol *self, const gchar *message)
+{
+  g_return_if_fail (MODULEMD_IS_EOL (self));
+  g_return_if_fail (message);
+  g_return_if_fail (g_strcmp0 (message, O_DEFAULT_STRING));
```
sgallagher, line 387: Don't attempt to validate this here. It will cause problems for anyone attempting to initialize on instantiation. make this a check in `modulemd_eol_validate()`.

modulemd/modulemd-eol.c
```diff
@@ -0,0 +1,1196 @@
...
+
+  g_object_notify_by_pspec (G_OBJECT (self), properties[PROP_MESSAGE]);
+}
+
+void
+modulemd_eol_set_obsoleted_by_module_name (
+  ModulemdEol *self, const gchar *obsoleted_by_module_name)
+{
+  g_return_if_fail (MODULEMD_IS_EOL (self));
+  g_return_if_fail (!(obsoleted_by_module_name && self->reset));
```
sgallagher, line 400: See above

modulemd/modulemd-eol.c
```diff
@@ -0,0 +1,1196 @@
...
+  g_object_notify_by_pspec (G_OBJECT (self),
+                            properties[PROP_OBSOLETED_BY_MODULE_NAME]);
+}
+
+void
+modulemd_eol_set_obsoleted_by_module_stream (
+  ModulemdEol *self, const gchar *obsoleted_by_module_stream)
+{
+  g_return_if_fail (MODULEMD_IS_EOL (self));
+  g_return_if_fail (!(obsoleted_by_module_stream && self->reset));
```
sgallagher, line 414: See above

modulemd/modulemd-eol.c
```diff
@@ -0,0 +1,1196 @@
...
+}
+
+void
+modulemd_eol_set_obsoleted_by (ModulemdEol *self,
+                               const gchar *obsoleted_by_module_name,
+                               const gchar *obsoleted_by_module_stream)
+{
+  g_return_if_fail (MODULEMD_IS_EOL (self));
+  g_return_if_fail (obsoleted_by_module_name);
+  g_return_if_fail (obsoleted_by_module_stream);
```
sgallagher, line 430: These are redundant and also should probably not `abort()` if they fail. Just check that they are valid in `modulemd_eol_validate()`.

modulemd/modulemd-eol.c
```diff
@@ -0,0 +1,1196 @@
...
+
+  properties[PROP_MODIFIED] = g_param_spec_uint64 (
+    "modified",
+    "Modified",
+    "The last modified time represented as a 64-bit integer "
+    "(such as 201807011200)",
+    0,
+    G_MAXUINT64,
+    0,
+    G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS);
```
sgallagher, line 557: Why is this not `G_PARAM_CONSTRUCT`?

modulemd/modulemd-eol.c
```diff
@@ -0,0 +1,1196 @@
...
+    G_MAXUINT64,
+    0,
+    G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS | G_PARAM_CONSTRUCT);
+
+  properties[PROP_MESSAGE] = g_param_spec_string (
+    "message",
+    "Message",
+    "A string describing the change, reason, etc.",
+    O_DEFAULT_STRING,
+    G_PARAM_READWRITE | G_PARAM_STATIC_STRINGS | G_PARAM_CONSTRUCT_ONLY);
```
sgallagher, line 601: Why is this `G_PARAM_CONSTRUCT_ONLY`?

kontura: I thought since the message is mandatory it can only be set upon object construction just like module name, module stream etc

modulemd/modulemd-eol.c
kontura: I have removed the checks, but since these values are optional they should not be checked in `modulemd_eol_validate()` right? They are always valid.

modulemd/modulemd-eol.c
sgallagher: You're right, I missed that.

modulemd/modulemd-eol.c
sgallagher: In `modulemd_eol_validate()` you should at least check that they are either both present or both absent. Having just one or the other doesn't make sense (and that should be mentioned in the spec too).

kontura: Added

kontura:
I have made `modulemd_eol_get_modified` public API because it is needed in libdnf.

modulemd/include/modulemd-2.0/modulemd-eol.h
```diff
@@ -0,0 +1,238 @@
...
+ *
+ * Since: 2.10
+ */
+const gchar *
+modulemd_eol_get_module_context (ModulemdEol *self);
+
+/**
+ * modulemd_eol_set_module_context:
+ * @self: This #ModulemdEol object.
+ * @context: (nullable): The name of the module context to which this eol applies.
```
mmathesius, line 116:
```suggestion
 * @module_context: (nullable): The name of the module context to which this eol applies.
```
Documentation should match actual parameter name.

modulemd/modulemd-eol.c
```diff
@@ -0,0 +1,1207 @@
...
+  modulemd_eol_set_obsoleted_by_module_name (
+    o, modulemd_eol_get_obsoleted_by_module_name (self));
+  modulemd_eol_set_obsoleted_by_module_stream (
+    o, modulemd_eol_get_obsoleted_by_module_stream (self));
+
+  return g_steal_pointer (&o);
+}
+
+gboolean
+modulemd_eol_validate (ModulemdEol *self, GError **error)
```
dmach, line 114: Just for the record, @kontura found the issues in my test data and knows about them:

You should validate the module (==name),  stream, and context fields.
Also obsoleted_by / {module, stream} should be validated.
I managed to use ':' character in the stream and it's definitely a forbidden character (NSVCA separator).

DNF also did not report a missing mandatory field (modified), but that might be due to that part of code being in progress.

modulemd/tests/test-modulemd-modulestream.c
```diff
@@ -3174,6 +3175,47 @@ module_stream_v1_regression_content_license (void)
...
+  stream = modulemd_module_stream_new (
+    MD_MODULESTREAM_VERSION_LATEST, "foo", "latest");
+  g_assert_nonnull (stream);
+  e = modulemd_eol_new (1, 2, "testmodule", "teststream", "testmessage");
+  g_assert_nonnull (e);
+
+  g_assert_null (modulemd_module_stream_get_eol (stream));
+
+  modulemd_module_stream_associate_eol (stream, e);
+
```
mmathesius, line 3193:
```suggestion
  g_clear_object (&e);
  
```

modulemd/tests/test-modulemd-modulestream.c
```diff
@@ -3174,6 +3175,47 @@ module_stream_v1_regression_content_license (void)
...
+  g_clear_object (&e);
+
+  e = modulemd_eol_new (1, 2, "testmodule", "teststream", "testmessage");
+  modulemd_eol_set_reset (e, TRUE);
+  g_assert_nonnull (e);
+  modulemd_module_stream_associate_eol (stream, e);
+  e = modulemd_module_stream_get_eol (stream);
+  g_assert_null (e);
+
```
mmathesius, line 3207:
```suggestion
  g_clear_object (&e);
  
```

sgallagher:
These patches are failing tests with SIGABRT.

kontura:
Yes, of the top of my head I think it was [this](https://github.com/fedora-modularity/libmodulemd/pull/476/commits/aa3ade20793e863aabf2504b23e3038290b0ebc9) `g_clear_object` that is causing that, I will fix it. I am actually now working on a bigger update and putting it all together with libdnf and dnf, hopefully it will be ready by the end of the week.

mmathesius:
> Yes, of the top of my head I think it was [this](https://github.com/fedora-modularity/libmodulemd/pull/476/commits/aa3ade20793e863aabf2504b23e3038290b0ebc9) `g_clear_object` that is causing that, I will fix it. I am actually now working on a bigger update and putting it all together with libdnf and dnf, hopefully it will be ready by the end of the week.

That could do it. I think my suggestion landed a couple lines below where I intended.

kontura:
Sorry for the delay, I got sidetracked.
I have added several commits. We decided to rename eols to obsoletes as it was more consistent in dnf and it also avoids a conflict here with existing eol. 
I have also changed the way obsoletes are stored in modules from a hash table to an array, because I have realized particular `stream:context` combination can have multiple obsoletes. Only one of which is the newest active (relevant to user), but there can be others set in the future using `eol_date` and we don't want to loose this information. This way its still accessible through particular module.
This needed appropriate change in merging.

There is also few other improvements. 
I tried to keep the original commits intact so that you can review only the new ones. I plan to squash the commits once you think its ready, or I can do it now if you prefer.

modulemd/include/modulemd-2.0/modulemd-module-index-merger.h
```diff
@@ -103,6 +103,18 @@ G_BEGIN_DECLS
...
+ * and modified values behaves as follows (when accessed through module):
+ *
+ * - Any obsoletes object that is provided by a single #ModulemdModuleIndex
+ *   will be the obsoletes object in the resulting merged #ModulemdModuleIndex.
+ * - If the #ModulemdModuleIndex inputs have different priorities (not common),
+ *   then the obsoletes from the highest priority #ModulemdModuleIndex will be
+ *   used and the others entirely discarded. (Priority is intended for providing
+ *   a total override, including an on-disk configuration).
+ * - If the repos have the same priority (such as "fedora" and "updates" in the
+ *   Fedora Project) the behaviour is undefined.
```
sgallagher, line 117: Please explain that the `modified` value on the `modulemd-obsoletes` document will be checked for entries in the same priority level.

modulemd/include/modulemd-2.0/modulemd-module-stream-v1.h
```diff
@@ -257,6 +257,23 @@ const gchar *
...
+ * there is no obsoletes set for this stream. Every obsoletes (even with reset) can
+ * be accessed from the streams module.
+ *
+ * Since: 2.10
+ */
+ModulemdObsoletes *
+modulemd_module_stream_v1_get_obsoletes_resolved (
+  ModulemdModuleStreamV1 *self);
+
+
```
sgallagher, line 276: I recommend *against* implementing this for ModuleStreamV1 documents. We do not have any of these in the repos at this point. (All RHEL 8.0 module streams are V2 and Fedora no longer has any supported releases with V1 content in them). A better approach would be to document that the use of `modulemd_module_index_add_obsoletes()` will automatically update the stream object to at least V2 as a side-effect. As it is, the ModuleIndex always upgrades every object inside it to match the highest Stream version it sees, so there is precedent for this.

Let's try to avoid enhancing ModuleStreamV1 ever again, please.

modulemd/include/modulemd-2.0/modulemd-module-stream.h
```diff
@@ -468,4 +469,20 @@ gboolean
...
+ * this object. If the associated obsoletes has reset attribute set, this
+ * function doesn't return it. From outside obsoletes with reset looks like
+ * there is no obsoletes set for this stream. Every obsoletes (even with reset) can
+ * be accessed from the streams module.
+ *
+ * Since: 2.10
+ */
+ModulemdObsoletes *
+modulemd_module_stream_get_obsoletes_resolved (ModulemdModuleStream *self);
+
```
sgallagher, line 487: As an addendum to my comment above, that would also mean that we don't need to make this a virtual method.

modulemd/include/modulemd-2.0/modulemd-module.h
```diff
@@ -304,4 +305,34 @@ modulemd_module_get_defaults (ModulemdModule *self);
 ModulemdTranslation *
 modulemd_module_get_translation (ModulemdModule *self, const gchar *stream);
 
+
+/**
+ * modulemd_module_get_obsoletes:
+ * @self: (in): This #ModulemdModule object.
+ *
+ * Returns: (transfer none) (element-type ModulemdObsoletes): All obsoletes attached to this module.
```
sgallagher, line 313: If you are doing `(transfer none)`, please indicate that this is returning the internal data and must not be modified or freed.

modulemd/include/modulemd-2.0/modulemd-obsoletes.h
```diff
@@ -0,0 +1,254 @@
...
+ *
+ * Since: 2.10
+ */
+void
+modulemd_obsoletes_set_modified (ModulemdObsoletes *self, guint64 modified);
+
+/**
+ * modulemd_obsoletes_set_reset:
+ * @self: This #ModulemdObsoletes object.
+ * @reset: (in): Whether to cancel/reset all previously specified obsoletes.
```
sgallagher, line 140: Please make this documentation more clear. It took me multiple readings to realize that this is setting an attribute of the object rather than clearing out the object.

modulemd/include/modulemd-2.0/modulemd-module-index-merger.h
kontura: I think this is already covered in the `Merging #ModulemdObsoletes entries with identical module, stream, context and modified values behaves as follows (when accessed through module)` because it isn't checked otherwise.

Or do you mean when associating it with a stream? Because I think that is a different thing and I would rather describe that `modulemd_module_stream_get_obsoletes_resolved` returns the latest active obsoletes valid for this stream (which is missing in the docs and I really should add it).

modulemd/include/modulemd-2.0/modulemd-module-stream-v1.h
kontura: oh, of course, I can drop the commit and add the description.

modulemd/include/modulemd-2.0/modulemd-module-stream.h
kontura: So do I understand correctly then there would be no `modulemd_module_stream_get_obsoletes_resolved` only `modulemd_module_stream_v2_get_obsoletes_resolved`? And I should move the `obsoletes` pointer from the `ModulemdModuleStreamPrivate` struct into `_ModulemdModuleStreamV2`?

kontura:
I have addressed some of the review and also tweaked the merging to better reflect the documented state. Now `modulemd_module_index_add_obsoletes` actually overrides existing duplicate (same module, stream, context and modified time) obsoletes.

modulemd/tests/ModulemdTests/merger.py
```diff
@@ -450,6 +450,71 @@ def test_merge_incompatible_streams(self):
...
+        module = merged_idx.get_module("nodejs")
+        self.assertEqual(len(module.get_obsoletes()), 2)
+
+        all_streams = merged_idx.search_streams_by_nsvca_glob()
+        self.assertEqual(len(all_streams), 1)
+        self.assertEqual(all_streams[0].props.module_name, "nodejs")
+
+        obsoletes = all_streams[0].get_obsoletes_resolved()
+        # obsoletes is None because it has reset: true
+        self.assertEqual(obsoletes, None)
```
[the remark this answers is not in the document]
kontura: oh, I was testing that the order of associating of indexes doesn't matter and I must have left it identical by mistake, will switch the order. Thanks!

kontura:
> Need to figure out where those memory leaks reported by valgrind are coming from.

I have pushed an update which fixes it but I have to say I wasn't able to reproduce the problem by using the [command specified in readme](https://github.com/fedora-modularity/libmodulemd#running-tests-with-valgrind). It showed up only when I used a command copied from the Travis CI, something like:

`G_DEBUG='fatal-warnings,fatal-criticals' MESON_SOURCE_ROOT='/root/dbox/libmodulemd' MESON_BUILD_ROOT='/root/dbox/libmodulemd/debugbuild2/' G_MESSAGES_DEBUG='all' LC_ALL='C' TEST_DATA_PATH='/root/dbox/libmodulemd/modulemd/tests/test_data' valgrind --error-exitcode=1 --errors-for-leak-kinds=definite --leak-check=full --show-leak-kinds=definite --suppressions=/usr/share/glib-2.0/valgrind/glib.supp --suppressions=/root/dbox/libmodulemd/contrib/valgrind/libmodulemd-python.supp /root/dbox/libmodulemd/debugbuild2/modulemd/module`

(I have updated the paths to work for my system)

mmathesius:
> > Need to figure out where those memory leaks reported by valgrind are coming from.
> 
> I have pushed an update which fixes it but I have to say I wasn't able to reproduce the problem by using the [command specified in readme](https://github.com/fedora-modularity/libmodulemd#running-tests-with-valgrind). It showed up only when I used a command copied from the Travis CI, something like:
> 
> `G_DEBUG='fatal-warnings,fatal-criticals' MESON_SOURCE_ROOT='/root/dbox/libmodulemd' MESON_BUILD_ROOT='/root/dbox/libmodulemd/debugbuild2/' G_MESSAGES_DEBUG='all' LC_ALL='C' TEST_DATA_PATH='/root/dbox/libmodulemd/modulemd/tests/test_data' valgrind --error-exitcode=1 --errors-for-leak-kinds=definite --leak-check=full --show-leak-kinds=definite --suppressions=/usr/share/glib-2.0/valgrind/glib.supp --suppressions=/root/dbox/libmodulemd/contrib/valgrind/libmodulemd-python.supp /root/dbox/libmodulemd/debugbuild2/modulemd/module`
> 
> (I have updated the paths to work for my system)

Interesting. I tried this myself in several combinations of options and Fedora releases and confirmed the same thing.

modulemd/include/modulemd-2.0/modulemd-module-stream.h
kontura: @sgallagher is the code ok in the current state in regards to `modulemd_module_stream_get_obsoletes_resolved`? (I have just dropped the v1 version) Can I squash the commits now?

sgallagher: > So do I understand correctly then there would be no `modulemd_module_stream_get_obsoletes_resolved` only `modulemd_module_stream_v2_get_obsoletes_resolved`? And I should move the `obsoletes` pointer from the `ModulemdModuleStreamPrivate` struct into `_ModulemdModuleStreamV2`?

Sorry, I missed the notification for this question. Yes, this is what I meant. The `ModuleStream` object is meant to only provide the set of functions that *all* streams must have. Basically, this just means the set that are required to add them to a `ModuleIndex`. So please remove this from ModuleStream and implement it in ModuleStreamV2.

kontura: Ok I have squashed the commits since it was getting more confusing I think, but most importantly I moved all the obsoletes from `ModuleStream` to `ModuleStreamV2`. I had to also add the automatic stream upgrading to `modulemd_module_add_obsoletes` and `modulemd_module_add_stream` since they can both associate obsoletes with a stream (also updated their docs).

kontura:
I fixed a memory leak (forgot to move obsoletes finalization from stream to stream v2).

modulemd/tests/test-modulemd-modulestream.c
```diff
@@ -3174,6 +3175,46 @@ module_stream_v1_regression_content_license (void)
...
+  g_assert_nonnull (stream);
+  o = modulemd_obsoletes_new (1, 2, "testmodule", "teststream", "testmessage");
+  g_assert_nonnull (o);
+
+  g_assert_null (modulemd_module_stream_v2_get_obsoletes_resolved (stream));
+
+  modulemd_module_stream_v2_associate_obsoletes (stream, o);
+  g_clear_object (&o);
+
+  o = modulemd_module_stream_v2_get_obsoletes_resolved (stream);
```
mmathesius, line 3194: Suggestion to add here:
```
  g_assert_nonnull (o);
```

modulemd/tests/test-modulemd-modulestream.c
```diff
@@ -3174,6 +3175,46 @@ module_stream_v1_regression_content_license (void)
...
+  g_assert_cmpstr (modulemd_obsoletes_get_module_name (o), ==, "testmodule");
+  g_assert_cmpstr (modulemd_obsoletes_get_module_stream (o), ==, "teststream");
+  g_assert_null (modulemd_obsoletes_get_module_context (o));
+
+  o = modulemd_obsoletes_new (1, 2, "testmodule", "teststream", "testmessage");
+  modulemd_obsoletes_set_reset (o, TRUE);
+  g_assert_nonnull (o);
```
mmathesius, line 3201: `g_assert_nonnull(o)` should be moved up a  line.

mmathesius (commented):
I just have a couple minor nit-picks in one of the tests. Otherwise LGTM.

kontura:
> I just have a couple minor nit-picks in one of the tests. Otherwise LGTM.

Test updated. :slightly_smiling_face:

mmathesius:
> > I just have a couple minor nit-picks in one of the tests. Otherwise LGTM.
> 
> Test updated.

Looks good. Thanks!

modulemd/include/modulemd-2.0/modulemd-module-stream.h
kontura: @sgallagher can you check the update from my previous comment? Or is there something else that needs fixing? If possible I would like to get this finished. :slightly_smiling_face:

sgallagher (approved):

Issue #479, opened 2020-07-29
modulemd-validator doesn't work with gziped files


FrostyX:
$ modulemd-validator foo.yaml.gz 
    foo.yaml.gz failed to validate
    foo.yaml.gz could not be read in its entirety: Parser error

PR #480, opened 2020-07-29
validator: Add support for compressed YAML

sgallagher:
Fixes https://github.com/fedora-modularity/libmodulemd/issues/479

### DIFF
--- a/modulemd/modulemd-validator.c
+++ b/modulemd/modulemd-validator.c
@@ -114,7 +114,6 @@ parse_file (const gchar *filename, GPtrArray **failures, GError **error)
   MMD_INIT_YAML_PARSER (parser);
   MMD_INIT_YAML_EVENT (event);
   g_autoptr (FILE) yaml_stream = NULL;
-  int saved_errno;
   g_autoptr (ModulemdModuleIndex) index = NULL;
 
   if (options.verbosity >= MMD_VERBOSE)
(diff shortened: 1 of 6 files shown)


sgallagher:
> This looks like a faithful reproduction of the compression handling code found in `modulemd_module_index_update_from_file()` in `modulemd-module-index.c`.
> 
> But I must be missing something. Why can't `parse_file()` be reduced to a small wrapper that simply calls `modulemd_module_index_update_from_file()`?

It needs to be able to call [`modulemd_index_update_from_parser()`](https://fedora-modularity.github.io/libmodulemd/latest/modulemd-2.0-Modulemd.ModuleIndex-(Private).html#modulemd-module-index-update-from-parser) because that function takes one extra argument: `autogen_module_name`. Without this argument, it can't parse `document: modulemd` that doesn't have the module and stream names set (such as those you would find in dist-git).

sgallagher:
> It needs to be able to call [`modulemd_index_update_from_parser()`](https://fedora-modularity.github.io/libmodulemd/latest/modulemd-2.0-Modulemd.ModuleIndex-(Private).html#modulemd-module-index-update-from-parser) because that function takes one extra argument: `autogen_module_name`. Without this argument, it can't parse `document: modulemd` that doesn't have the module and stream names set (such as those you would find in dist-git).

I suppose an alternative approach would be to create a private `modulemd_update_from_file_ext()` function that both could use. That would probably be better, as it would avoid the code duplication. I will just do that.

mmathesius:
> > It needs to be able to call [`modulemd_index_update_from_parser()`](https://fedora-modularity.github.io/libmodulemd/latest/modulemd-2.0-Modulemd.ModuleIndex-(Private).html#modulemd-module-index-update-from-parser) because that function takes one extra argument: `autogen_module_name`. Without this argument, it can't parse `document: modulemd` that doesn't have the module and stream names set (such as those you would find in dist-git).

_That's_ what I was missing. So close...

> I suppose an alternative approach would be to create a private `modulemd_update_from_file_ext()` function that both could use. That would probably be better, as it would avoid the code duplication. I will just do that.

I'd be happy to take this on--unless you really want to (or almost have it done already).

sgallagher:
> > > It needs to be able to call [`modulemd_index_update_from_parser()`](https://fedora-modularity.github.io/libmodulemd/latest/modulemd-2.0-Modulemd.ModuleIndex-(Private).html#modulemd-module-index-update-from-parser) because that function takes one extra argument: `autogen_module_name`. Without this argument, it can't parse `document: modulemd` that doesn't have the module and stream names set (such as those you would find in dist-git).
> 
> _That's_ what I was missing. So close...
> 
> > I suppose an alternative approach would be to create a private `modulemd_update_from_file_ext()` function that both could use. That would probably be better, as it would avoid the code duplication. I will just do that.
> 
> I'd be happy to take this on--unless you really want to (or almost have it done already).

Thanks for the offer, but I just did it. See the new version I just pushed.

modulemd/modulemd-validator.c
```diff
@@ -114,36 +114,16 @@ parse_file (const gchar *filename, GPtrArray **failures, GError **error)
   MMD_INIT_YAML_PARSER (parser);
   MMD_INIT_YAML_EVENT (event);
   g_autoptr (FILE) yaml_stream = NULL;
```
mmathesius, line 116:
```suggestion
```
Now extraneous.

PR #483, opened 2020-09-15
Add tests for modulemd_load_string()

mmathesius:
I added tests and coded them so they currently pass. But I question if the current behavior of `modulemd_load_string()` is actually correct. Particularly note my double-slash comments...

### DIFF
--- a/modulemd/tests/test-modulemd-common.c
+++ b/modulemd/tests/test-modulemd-common.c
@@ -82,6 +82,75 @@ test_modulemd_load_file (void)
 }
 
 
+static void
+test_modulemd_load_string (void)
+{
+  const gchar *yaml_string = NULL;
+  g_autoptr (GError) error = NULL;
+  g_autoptr (ModulemdModuleIndex) idx = NULL;
+  g_autofree gchar *output = NULL;
+
+  /* This function is a wrapper around lower-level functions, so it should be
+   * okay to just test basic success and failure here.
+   */
+
+  /* Trivial modulemd */
+  g_clear_error (&error);
+  g_clear_object (&idx);
+  yaml_string =
+    "---\n"
+    "document: modulemd\n"
+    "version: 2\n"
+    "data:\n"
+    "  name: trivialname\n"
+    "  stream: trivialstream\n"
+    "  summary: Trivial Summary\n"
+    "  description: >-\n"
+    "    Trivial Description\n"
+    "  license:\n"
+    "    module: MIT\n"
+    "...\n";
+  idx = modulemd_load_string (yaml_string, &error);
+  g_assert_no_error (error);
+  g_assert_nonnull (idx);
+
+  /* Make sure loaded index dumps to string cleanly */
+  g_clear_error (&error);
+  g_clear_pointer (&output, g_free);
+  output = modulemd_module_index_dump_to_string (idx, &error);
+  g_assert_no_error (error);
+  g_assert_nonnull (output);
+
+
+  /* NULL string should raise an exception */
+  g_clear_error (&error);
+  g_clear_object (&idx);
+  modulemd_test_signal = 0;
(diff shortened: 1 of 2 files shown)


modulemd/tests/test-modulemd-common.c
```diff
@@ -82,6 +82,84 @@ test_modulemd_load_file (void)
...
+  g_clear_error (&error);
+  g_clear_object (&idx);
+  modulemd_test_signal = 0;
+  signal (SIGTRAP, modulemd_test_signal_handler);
+  idx = modulemd_load_string (NULL, &error);
+  g_assert_cmpint (modulemd_test_signal, ==, SIGTRAP);
+  g_assert_null (idx);
+
+
+  // An empty string returns non-NULL but empty index. Is that right?
```
sgallagher, line 135: That's a really good question, actually.

With the previous behavior (`ModuleIndex.new()` followed by `ModuleIndex.update_from_*()`), you would initialize an empty `ModuleIndex` first and then updating from an empty string would populate it with zero new entries. An empty string is valid YAML (it would appear as a YAML stream of zero documents). The same would be true of a zero-length file. So what you're getting here is the same as you would get there: an empty `ModuleIndex`. I think this is therefore correct.

modulemd/tests/test-modulemd-common.c
```diff
@@ -82,6 +82,84 @@ test_modulemd_load_file (void)
...
+  g_clear_error (&error);
+  g_clear_object (&idx);
+  yaml_string = "";
+  idx = modulemd_load_string (yaml_string, &error);
+  g_assert_no_error (error);
+  g_assert_nonnull (idx);
+
+  // Outputting the empty index returned above returns a validation error...
```
sgallagher, line 143: Yes, it returns `MMD_ERROR_VALIDATE` and the `error->message` should contain "Index contains no modules".

modulemd/tests/test-modulemd-common.c
```diff
@@ -82,6 +82,84 @@ test_modulemd_load_file (void)
...
+  g_assert_null (output);
+
+
+  /* Invalid YAML string */
+  g_clear_error (&error);
+  g_clear_object (&idx);
+  yaml_string = "Hello, World!\n";
+  idx = modulemd_load_string (yaml_string, &error);
+  // Should it actually return MMD_YAML_ERROR_UNPARSEABLE?
+  // Or should it just assert it's anything in the MODULEMD_YAML_ERROR domain?
```
sgallagher, line 157: Probably we should clarify the definition of those two errors.

`MMD_YAML_ERROR_UNPARSEABLE` means that libyaml has thrown an error trying to interpret the data. For example, `modulemd_load_string ('[', &error);` would throw it because it encountered the start of an array but never the end of it.

It's basically the difference between "violates the YAML spec" and "violates the modulemd spec", if that helps.

sgallagher (changes requested):

modulemd/tests/test-modulemd-common.c
mmathesius: OK. It just seems asymmetric that it's valid to load an empty string/file, but not valid to dump it back out.

modulemd/tests/test-modulemd-common.c
mmathesius: Got it. I'll try to add some clarifying text for those two errors in `modulemd/include/modulemd-2.0/modulemd-errors.h`.

modulemd/tests/test-modulemd-common.c
mmathesius: I'm just going to drop this test that attempts to output the empty index since it's existing behavior and really does nothing to test `modulemd_load_string()`.

sgallagher (approved):

PR #486, opened 2020-10-07
[WIP] Add modulemd-packager and modulemd v3 document specs

sgallagher:

### DIFF
--- a/modulemd/tests/test_data/upgrades/packager_v3_to_stream_v3.yaml
+++ b/modulemd/tests/test_data/upgrades/packager_v3_to_stream_v3.yaml
@@ -14,6 +14,8 @@ data:
     license:
         module:
             - MIT
+    xmd:
+        some_key: some_data
     dependencies:
         platform: f32
         buildrequires:
@@ -49,6 +51,17 @@ data:
     filter:
         rpms:
             - baz-nonfoo
+    buildopts:
+        rpms:
+            macros: |
+                %demomacro 1
+                %demomacro2 %{demomacro}23
+            whitelist:
+                - fooscl-1-bar
+                - fooscl-1-baz
+                - xxx
+                - xyz
+        arches: [i686, x86_64]
     components:
         rpms:
             bar:
@@ -93,6 +106,8 @@ data:
     license:
         module:
             - MIT
+    xmd:
+        some_key: some_data
     dependencies:
         platform: f33
     references:
@@ -152,4 +167,11 @@ data:
                 ref: somecoolbranchname
                 buildorder: 100
 ...
-
+---
+document: modulemd-defaults
+version: 1
+data:
+  module: testmodule
+  profiles:
+    'teststream': [minimal]
+...
--- a/modulemd/tests/test_data/upgrades/packager_v3_to_stream_v3.yaml
+++ b/modulemd/tests/test_data/upgrades/packager_v3_to_stream_v3.yaml
@@ -79,7 +79,7 @@ data:
 ...
 ---
 document: modulemd
-version: 2
+version: 3
 data:
     name: testmodule
     stream: teststream
@@ -152,3 +152,4 @@ data:
                 ref: somecoolbranchname
                 buildorder: 100
 ...
+
--- a/modulemd/tests/test_data/upgrades/packager_v3_to_stream_v3.yaml
+++ b/modulemd/tests/test_data/upgrades/packager_v3_to_stream_v3.yaml
@@ -1,0 +1,154 @@
+---
+document: modulemd
+version: 3
+data:
+    name: testmodule
+    stream: teststream
+    version: 1
+    context: CTX1
+    arch: x86_64
+    summary: An example module
+    description: >-
+        A module for the demonstration of the metadata format. Also,
+        the obligatory lorem ipsum dolor sit amet goes right here.
+    license:
+        module:
+            - MIT
+    dependencies:
+        platform: f32
+        buildrequires:
+            appframework: v1
+        requires:
+            appframework: v1
+    references:
+        community: http://www.example.com/
+        documentation: http://www.example.com/
+        tracker: http://www.example.com/
+    profiles:
+        container:
+            rpms:
+                - bar
+                - bar-devel
+        minimal:
+            description: Minimal profile installing only the bar package.
+            rpms:
+                - bar
+        buildroot:
+            rpms:
+                - bar-devel
+        srpm-buildroot:
+            rpms:
+                - bar-extras
+    api:
+        rpms:
+            - bar
+            - bar-extras
+            - bar-devel
+            - baz
+            - xxx
+    filter:
+        rpms:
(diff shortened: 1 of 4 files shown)


sgallagher:
Open Question: Is the `servicelevels` data useful? Should we drop it in v3?

sgallagher:
> Open Question: Is the `servicelevels` data useful? Should we drop it in v3?

Discussed at Modularity meeting: this information is not useful in the stream document and should be dropped.

sgallagher:
ATTN: @mcurlej

sgallagher:
@mcurlej I remember that at one point we were discussing keeping a dictionary of context->dependencies in the metadata. I *think* we ultimately decided against that, but I don't remember the entire story.

sgallagher:
I just pushed an update that also has a view of what the output of v2 and v3 stream documents should look like, given the `modulemd-packager` v3 spec as input. The implementation should derive from that.

modulemd/tests/test_data/upgrades/packager_v3_to_stream_v3.yaml
```diff
@@ -0,0 +1,154 @@
...
+        modules:
+            includedmodule:
+                rationale: Included in the stack, just because.
+                repository: https://pagure.io/includedmodule.git
+                ref: somecoolbranchname
+                buildorder: 100
+...
+---
+document: modulemd
+version: 2
```
mmathesius, line 82:
```suggestion
version: 3
```

mmathesius (changes requested):
The example output stream documents will be greatly helpful. But I'm pretty sure there's an off-by-one error!

modulemd/tests/test_data/upgrades/packager_v3_to_stream_v3.yaml
sgallagher: Good catch, I'll fix that in the next push.

sgallagher:
Question for you, @mmathesius : while we're creating a new stream format, do you think it might make sense to also rename (or alias) the document type to `document: modulemd-stream`? Now that we have `modulemd-defaults`, `modulemd-translations` and `modulemd-obsoletes`, having the streams be `document: modulemd` seems inconsistent and confusing.

Perhaps we can have libmodulemd accept either string to represent streams and use the `document: modulemd-stream` as the canonical version going forward?

sgallagher:
I also realized I forgot the important part of the `modulemd-packager` v3 -> `modulemd-stream` v2 conversion: it needs to be able to produce a single v2 document so it can be read by older MBS unchanged. I'll fix that up too.

mmathesius:
> Question for you, @mmathesius : while we're creating a new stream format, do you think it might make sense to also rename (or alias) the document type to `document: modulemd-stream`? Now that we have `modulemd-defaults`, `modulemd-translations` and `modulemd-obsoletes`, having the streams be `document: modulemd` seems inconsistent and confusing.
> 
> Perhaps we can have libmodulemd accept either string to represent streams and use the `document: modulemd-stream` as the canonical version going forward?

I, too, find the current naming inconsistent. Are you suggesting the rename/alias only for v3? Or allow it as an input alias for all versions, continue to use the old name for pre-v3 output, and use the new name for v3 (and later) output?

mmathesius:
> I also realized I forgot the important part of the `modulemd-packager` v3 -> `modulemd-stream` v2 conversion: it needs to be able to produce a single v2 document so it can be read by older MBS unchanged. I'll fix that up too.

Thanks. I should have caught that.

sgallagher:
> I, too, find the current naming inconsistent. Are you suggesting the rename/alias only for v3? Or allow it as an input alias for all versions, continue to use the old name for pre-v3 output, and use the new name for v3 (and later) output?

I mean the latter, yes. In practical terms, it means changing [modulemd_yaml_parse_document_type_internal()](https://github.com/fedora-modularity/libmodulemd/blob/main/modulemd/modulemd-yaml-util.c#L741) to treat either name as valid and then updating the output routines to ensure that versions older than v3 continue to use the older name.

modulemd/tests/test_data/upgrades/packager_v3_to_stream_v3.yaml
```diff
@@ -152,4 +167,11 @@ data:
                 ref: somecoolbranchname
                 buildorder: 100
 ...
-
+---
```
mmathesius, line 170: What is the trigger for this modulemd-defaults document appearing? Did you intend to include `data.profiles.minimal.default: true` in `modulemd_packager_v3.yaml`?

sgallagher: Yes, did I miss that?

sgallagher: Fixed.

yaml_specs/modulemd_packager_v3.yaml
```diff
@@ -221,6 +221,10 @@ data:
             description: Minimal profile installing only the bar package.
             rpms:
                 - bar
+            # default (boolean):
+            # Whether this profile should be installed if none is explicitly
+            # specified
+            default: True
```
mmathesius, line 227:
```suggestion
            # Whether this profile should be installed if none is explicitly
            # specified. No more than one configuration may have `default: true`.
            #
            # Defaults to "false" if not specified.
            #
            # Type: OPTIONAL
            default: true
```

yaml_specs/modulemd_packager_v3.yaml
```diff
@@ -0,0 +1,486 @@
...
+        #
+        # Type: Optional
+        minimal:
+            # A verbose description of the module, optional
+            description: Minimal profile installing only the bar package.
+            rpms:
+                - bar
+            # default (boolean):
+            # Whether this profile should be installed if none is explicitly
+            # specified. No more than one configuration may have `default: true`.
```
mmathesius, line 226:
```suggestion
            # specified.
```

Apparently this restriction on the `default` property shouldn't be there, since any number of profiles can be installed by default.

mcurlej:
@sgallagher wrt the dictionary, the first proposal was that there will be a `contexts` property which would be a dict and the keys would be names of the contexts. But as we renamed the property to `builds` it did not make sense to have the keys names of the contexts.  Then we agreed on `configurations` and and array makes more sense as the name of a context can be a part of the configuration and we still have a connection between modulemd-packager and modulemd.

sgallagher:
> @sgallagher wrt the dictionary, the first proposal was that there will be a `contexts` property which would be a dict and the keys would be names of the contexts. But as we renamed the property to `builds` it did not make sense to have the keys names of the contexts. Then we agreed on `configurations` and and array makes more sense as the name of a context can be a part of the configuration and we still have a connection between modulemd-packager and modulemd.

Right, I think I confused two different topics. In any case, I think that means this approach is what we agreed on.

PR #490, opened 2020-10-23
Do not treat OpenMandriva tests as blocking

sgallagher:


Conan-Kudo:
It looks like something has gone haywire on the published repositories... @berolinux, can you take a look here?

PR #491, opened 2020-10-26
Add common mmd_parse_xmd() function

sgallagher:

### DIFF
--- a/modulemd/modulemd-module-stream-v1.c
+++ b/modulemd/modulemd-module-stream-v1.c
@@ -1532,9 +1532,6 @@ modulemd_module_stream_v1_parse_components (
   gboolean strict,
   GError **error);
 
-static GVariant *
-modulemd_module_stream_v1_parse_raw (yaml_parser_t *parser, GError **error);
-
 
 ModulemdModuleStreamV1 *
 modulemd_module_stream_v1_parse_yaml (ModulemdSubdocumentInfo *subdoc,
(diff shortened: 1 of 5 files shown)


sgallagher:
These were 100% copy-pasted and I'm going to need it in the `PackagerV3` as well, so it made sense to deduplicate it and add it to the yaml utility functions.

modulemd/modulemd-module-stream-v1.c
```diff
@@ -2516,45 +2512,6 @@ modulemd_module_stream_v1_parse_module_components (
...
-
-    default:
-      MMD_YAML_ERROR_EVENT_EXIT (error,
-                                 event,
-                                 "Unexpected YAML event in raw parsing: %s",
-                                 mmd_yaml_get_event_name (event.type));
-      break;
-    }
-
-  return g_steal_pointer (&variant);
```
mmathesius, line 2554: I'm not familiar with `GVariant`s. I see v1 was returning `g_steal_pointer (&variant)` while v2 (and the v3 clone and your new common function) return `g_variant_ref_sink (variant)`. Are they pretty much the same?

mmathesius (approved):
Just a question. Otherwise LGTM. Thanks!

modulemd/modulemd-module-stream-v1.c
sgallagher: In this particular instance, they are the same. `g_variant_ref_sink()` is more correct here, but they essentially end up compiling down to the same code.

sgallagher:
Some of the tests hit a timeout. Resuming them before we merge.

modulemd/modulemd-module-stream-v1.c
sgallagher: Put another way: we should probably fix it for v1, but since v1 is dead and buried it likely doesn't really matter.

mmathesius: Thanks for the explanation. I assume by "v1" you are referring to libmodulemd V1, since this PR now fixes if for _stream_ v1 in libmodulemd V2.

PR #492, opened 2020-10-27
Add YAML emit capabilities to BuildConfig object

mmathesius:
I found it worth my time to add this while working on the StreamV2 to StreamV3 upgrade dependency stream expansion so I could visualize the data structures.

I also happened to notice and fix a small issue with the implementation of the `MMD_INIT_YAML_STRING()` and `MMD_REINIT_YAML_STRING()` macros.

### DIFF
--- a/modulemd/include/private/modulemd-build-config.h
+++ b/modulemd/include/private/modulemd-build-config.h
@@ -299,6 +299,25 @@ modulemd_build_config_parse_yaml (yaml_parser_t *parser,
 
 
 /**
+ * modulemd_build_config_emit_yaml:
+ * @self: This #ModulemdBuildConfig object.
+ * @emitter: (inout): A libyaml emitter object positioned where a BuidConfig
+ * belongs in the YAML document.
+ * @error: (out): A #GError that will return the reason for an emission or
+ * validation error.
+ *
+ * Returns: TRUE if the BuildConfig was emitted successfully. FALSE and sets
+ * @error appropriately if the YAML could not be emitted.
+ *
+ * Since: 2.10
+ */
+gboolean
+modulemd_build_config_emit_yaml (ModulemdBuildConfig *self,
+                                 yaml_emitter_t *emitter,
+                                 GError **error);
+
+
+/**
  * modulemd_build_config_validate:
  * @self: (in): This #ModulemdBuildConfig object.
  * @error: (out): A #GError explaining any validation failure.
--- a/modulemd/include/private/modulemd-build-config.h
+++ b/modulemd/include/private/modulemd-build-config.h
@@ -346,4 +346,20 @@ modulemd_build_config_validate (ModulemdBuildConfig *buildconfig,
  */
 ModulemdBuildConfig *
 modulemd_build_config_copy (ModulemdBuildConfig *self);
+
+/**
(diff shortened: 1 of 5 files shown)


modulemd/modulemd-build-config.c
```diff
@@ -364,6 +364,56 @@ modulemd_build_config_parse_yaml (yaml_parser_t *parser,
...
+                                 yaml_emitter_t *emitter,
+                                 GError **error)
+{
+  MODULEMD_INIT_TRACE ();
+  int ret;
+  g_autoptr (GError) nested_error = NULL;
+  MMD_INIT_YAML_EVENT (event);
+
+  ret = mmd_emitter_start_mapping (
+    emitter, YAML_BLOCK_MAPPING_STYLE, &nested_error);
```
sgallagher, line 378: Before we actually start emitting content, please run `modulemd_build_config_validate()` to ensure that the contents are acceptable to output.

modulemd/modulemd-build-config.c
```diff
@@ -364,6 +364,56 @@ modulemd_build_config_parse_yaml (yaml_parser_t *parser,
...
+  EMIT_HASHTABLE_KEY_VALUES_IF_NON_EMPTY (
+    emitter, error, "buildrequires", self->buildrequires);
+  EMIT_HASHTABLE_KEY_VALUES_IF_NON_EMPTY (
+    emitter, error, "requires", self->requires);
+
+  if (self->buildopts != NULL)
+    {
+      EMIT_SCALAR (emitter, error, "buildopts");
+      EMIT_MAPPING_START (emitter, error);
+      if (!modulemd_buildopts_emit_yaml (self->buildopts, emitter, error))
```
sgallagher, line 398: Use the nested_error here and propagate it, please.

I prefer to have them clearly differentiated in case we ever decide that certain errors can be ignored.

modulemd/include/private/modulemd-build-config.h
```diff
@@ -346,4 +346,21 @@ modulemd_build_config_validate (ModulemdBuildConfig *buildconfig,
  */
 ModulemdBuildConfig *
 modulemd_build_config_copy (ModulemdBuildConfig *self);
+
+/**
+ * modulemd_build_config_equals:
+ * @self_1: A #ModulemdBuildConfig object.
+ * @self_2: A #ModulemdBuildConfig object.
+ *
+ * Check if @self_1 and @self_2 are equal objects.
```
sgallagher, line 355: Rather than "equal objects", can you explain that we're testing whether they have identical *contents*? "Equal objects" sounds to me like we're checking to see if they are the same instance.

sgallagher (changes requested):
Just a few minor changes, but this looks quite good.

modulemd/modulemd-build-config.c
mmathesius: Shouldn't the validation be done just before starting to emit content of any higher-level stand-alone document format that makes use of BuildConfig? Such upper level validation should already be calling `modulemd_build_config_validate()`.

sgallagher: Oh right. You are correct of course.

sgallagher (approved):
LGTM :+1:

PR #497, opened 2020-11-03
Implement PackagerV3 mapping to StreamV2/StreamV3

mmathesius:
This PR provides low level functionality to map PackagerV3 to StreamV2/StreamV3. Also a few corrections to the example YAML files for testing the mapping, as well as minor bug fixes for handling of the Profile 'default' attribute.

Basic tests are included. Of course more could be added later.

This PR should provide the functionality needed to adjust the higher level functions to accept a modulemd-packager v3 document and return it as if it were a modulemd-stream v2 object. However, those adjustments are not included in this PR.

### DIFF
--- a/modulemd/include/private/modulemd-packager-v3.h
+++ b/modulemd/include/private/modulemd-packager-v3.h
@@ -20,7 +20,10 @@
 #include "modulemd-2.0/modulemd-profile.h"
 #include "modulemd-2.0/modulemd-subdocument-info.h"
 
+#include "private/modulemd-module-stream-v2-private.h"
+#include "private/modulemd-module-stream-v3-private.h"
 #include "private/modulemd-build-config.h"
+#include "private/modulemd-module-index-private.h"
 
 G_BEGIN_DECLS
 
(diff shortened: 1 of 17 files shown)


mmathesius:
> This PR should provide the functionality needed to adjust the higher level functions to accept a modulemd-packager v3 document and return it as if it were a modulemd-stream v2 object. However, those adjustments are not included in this PR.

My latest commit just added those adjustments to this PR.

mmathesius:
My latest commit provides the functionality for adding an modulemd-packager v3 document to an index--which will add it as StreamV2 or StreamV3 according to the current mdversion of the index.

This commit should wrap up the functionality for this PR!

modulemd/include/private/modulemd-packager-v3.h
```diff
@@ -715,6 +718,86 @@ modulemd_packager_v3_get_rpm_component (ModulemdPackagerV3 *self,
...
+modulemd_packager_v3_to_stream_v3 (ModulemdPackagerV3 *self, GError **error);
+
+/**
+ * modulemd_packager_v3_to_stream_v2_ext:
+ * @self: (in): This #ModulemdPackagerV3 object.
+ * @error: (out): A #GError that will return the reason for a conversion error.
+ *
+ * Returns: (transfer full): A newly-allocated #ModulemdModuleIndex object
+ * containing one or more #ModulemdModuleStreamV3 objects and possibly a
+ * #ModulemdDefaultsV1 object corresponding to the #ModulemdPackagerV3 object
```
sgallagher, line 790:
```suggestion
 * #ModulemdDefaults object corresponding to the #ModulemdPackagerV3 object
```

modulemd/modulemd-module-stream.c
```diff
@@ -259,18 +262,35 @@ modulemd_module_stream_read_yaml (yaml_parser_t *parser,
...
+            }
+
+          /* convert packager v3 to stream v2 for backwards compatibility */
+          stream = MODULEMD_MODULE_STREAM (
+            modulemd_packager_v3_to_stream_v2 (packager_v3, &nested_error));
+          if (!stream)
+            {
+              g_propagate_error (error, g_steal_pointer (&nested_error));
+              return NULL;
+            }
```
sgallagher, line 280: We need to conditionalize this, because if we are expecting to be v3, we don't want to downgrade and upgrade later. (Among other things, this would result in losing the per-configuration buildopts).

modulemd/modulemd-packager-v3.c
```diff
@@ -738,6 +739,449 @@ modulemd_packager_v3_get_rpm_component (ModulemdPackagerV3 *self,
...
+
+  module = modulemd_module_index_get_module (index, module_names[0]);
+  module_streams = modulemd_module_get_all_streams (module);
+
+  if (module_streams->len != 1)
+    {
+      g_set_error_literal (error,
+                           MODULEMD_ERROR,
+                           MMD_ERROR_UPGRADE,
+                           "Stream v2 upgrade must return a single stream.");
```
sgallagher, line 1012: I suspect this is the wrong error message.

sgallagher (changes requested):

modulemd/modulemd-module-stream.c
mmathesius: I thought we discussed doing this was the key to provided backwards compatibility to existing consumers of libmodulemd. (eg., RHEL 8) Do you have any suggestions for how the API should be modified/extended so the preference for StreamV2 vs StreamV3 can be indicated?

sgallagher: The conversion from PackagerV3->StreamV2 is slightly lossy due to the buildopts. If we don't want to lose that, we'll need to retain that data to re-add it for a later upgrade.

Maybe we could have `modulemd_packager_v3_to_stream_v2()` store a copy of the original `PackagerV3` object internally and in the event that the consumer calls `modulemd_modulestream_upgrade()`, we produce the results converted from the Packager object instead of the StreamV2 object? We'd have to make sure they don't get out of sync; maybe we can have it so that the stored Packager object is freed if any "write" to the StreamV2 object occurs?

modulemd/include/private/modulemd-packager-v3.h
```diff
@@ -715,6 +718,86 @@ modulemd_packager_v3_get_rpm_component (ModulemdPackagerV3 *self,
...
+ * @error: (out): A #GError that will return the reason for a conversion error.
+ *
+ * Returns: (transfer full): A newly-allocated #ModulemdDefaults object
+ * corresponding to the #ModulemdPackagerV3 object @self if @self contains any
+ * profiles marked as default. NULL if no default profiles.
+ *
+ * Since: 2.10
+ */
+ModulemdDefaults *
+modulemd_packager_v3_to_defaults (ModulemdPackagerV3 *self, GError **error);
```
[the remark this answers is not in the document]
mmathesius: Thank you for the suggestions. I went with option 2.

modulemd/modulemd-module-stream.c
mmathesius: I implemented this per our earlier discussion by creating a new configuration setting and API for specifying the desired default stream metadata version, and using that setting to determine if the packager v3 document gets returned as a StreamV2 or StreamV3 object.

Note: since this function returns a single Stream, it will fail if the packager v3 document contains multiple build configurations--which causes it to expand to multiple StreamV3s.

modulemd/modulemd-module-index.c
```diff
@@ -117,6 +118,7 @@ modulemd_module_index_init (ModulemdModuleIndex *self)
 {
   self->modules =
     g_hash_table_new_full (g_str_hash, g_str_equal, g_free, g_object_unref);
+  self->stream_mdversion = modulemd_get_default_stream_mdversion ();
```
mmathesius, line 121: @sgallagher After our discussions today, it may be best to leave this unset (`MD_DEFAULTS_VERSION_UNSET`)--at least until we work out how to deal with the bigger stream upgrade problem.

sgallagher: Yeah, leave it unset for now.

sgallagher (approved):
I think this looks good. If you want to clean up the commit history a little, we'll get this merged. We can fix any issues as bugs.

modulemd/modulemd-module-index.c
mmathesius: I reverted that change.

mmathesius:
> I think this looks good. If you want to clean up the commit history a little, we'll get this merged. We can fix any issues as bugs.

Thank you for the review.

I reverted the change to set the default `stream_mdversion` when creating a new ModulemdModuleIndex, and squashed the commits.

PR #498, opened 2020-11-04
CI: replace RPM creation with packit

sgallagher:

Also clean up the unnecessary Fedora `%changelog`.


.travis/fedora/travis-tasks.sh
```diff
@@ -67,29 +67,24 @@ fi
...
+createrepo_c $arch
 
 $RETRY_CMD dnf -y install --nogpgcheck \
                --allowerasing \
-               --repofrompath libmodulemd-travis,rpmbuild/RPMS \
-               python3-libmodulemd \
-               "libmodulemd-devel > 2"
+               --repofrompath libmodulemd-travis,$arch \
+               $arch/python3-libmodulemd*.rpm \
+               $arch/libmodulemd-devel*.rpm
```
mmathesius, line 80: I still don't know much about packit. :wink: But, why did you have to change these lines to install the created RPMs explicitly by globbed pathnames?  Aside from that, everything else looks good to me (as far as I can tell).

sgallagher: Because the way we were setting the version and the way packit does it is different.

We were getting the version from meson.build and using `meson dist` to generate the tarball for the RPM build. With packit, it generates the tarball from git and names it based on `git describe`. As a result, the packit version will generally sort lower than what is in the public repositories, so I changed it to install the files directly and use the repo as a backup for resolving dependencies. This will be more robust in general, since we won't have to worry as much about remembering to bump the version in meson.build at release time (though that's automated now through [`contrib/release-tools/release.sh`](https://github.com/fedora-modularity/libmodulemd/blob/main/contrib/release-tools/release.sh)).

mmathesius: That makes sense. Thanks for the explanation!

PR #499, opened 2020-11-10
Always output quoted stream name

sgallagher:
Some module streams are named after versions such as "5.30" which
are coerced into numbers (5.3) in other YAML parsers (such as the
default configuration of pyyaml). Since this format is common, we
should avoid outputting ambiguous stream names.

I'm not aware of other fields that would be ambiguous in this way.

h/t @james-antill


james-antill:
Nice, that was fast! It feels a bit weird to quote everything instead of just the entries that need it (and thus having to change all the test cases) ... but I assume this way is like 666x easier to implement. so LGTM.

mmathesius (approved):
Looks good to me. I agree with @james-antill that it looks a little odd to have just stream names quoted, but it makes sense and gets the job done.

sgallagher:
I considered having it only quote the string if the value could pass through `g_ascii_strtod()` unchanged, but that would have required an unpleasant degree of additional work for something that is ultimately cosmetic.

james-antill:
It's technically worse than just strtod(), in both directions, because "true", "nan" or "null" might have similar problems but "2.6" works fine (because it's the same output when you convert it back to a string).

On my end, for the next yaml parser after my code, all I did was:

    if stream.endswith("0"):
        stream = '"' + stream + '"'

...because it works fine if "1.23" is parsed as a float, but not "5.30". I've google'd a bunch, and the default parsers all seem to be aware of the [default schemas](https://perlpunk.github.io/yaml-test-schema/schemas.html) and go to a bunch of effort to make sure things are quoted when required but don't seem to have any API to easily say "wtf will this string be parsed as" or "do I need to double quote this so it'll be parsed as a string". Sigh.

PR #501, opened 2020-11-11
Change the specification of `ref`

sgallagher:
Loosen the restriction on how the build system must treat an
unspecified `ref` in a component. It must always fall back to the
repo's default branch (which may not be 'master' these days), but
if build-system rules exist to pick a different branch, that is
up to the build-system to decide.

Fixes https://pagure.io/modularity/issue/142

CC @voxik

### DIFF
--- a/yaml_specs/modulemd_packager_v3.yaml
+++ b/yaml_specs/modulemd_packager_v3.yaml
@@ -380,8 +380,10 @@ data:
                 # ref:
                 # Use this specific commit hash, branch name or tag for
                 # the build.  If ref is a branch name, the branch HEAD
-                # will be used.  If no ref is given, the master branch
-                # is assumed.
+                # will be used.  If no ref is given, the build system is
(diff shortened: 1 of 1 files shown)


voxik:
This makes sense to me and opens the possibility for infrastructure to be more clever about the refs, so +1.

However, the language "but it must fall back to the default branch" feels to strong to my taste. What about: "If no ref is given, the build system is permitted to make reasonable assumptions about the correct default value, otherwise the fallback is default branch for the git repository (commonly 'master')."

sgallagher:
>  "If no ref is given, the build system is permitted to make reasonable assumptions about the correct default value, otherwise the fallback is default branch for the git repository (commonly 'master')."

I accepted this suggestion, rebased and pushed another version. I think we're ready to commit this now. (No need to wait for CI; it's a comment-only change).

mmathesius (approved):
LGTM

PR #504, opened 2020-11-16
Handle XMD for ModuleStreamV3 in python

sgallagher:
I missed when reviewing ModuleStreamV3 that it didn't support the python override to make dealing with XMD more sane. This corrects that and also makes some minor changes to the test runner to ensure that we are always testing the correct version of the override.

I modified the `test_xmd()` routine from the python tests so it will now test XMD on all available stream versions.

I also dropped the `$RETRY_CMD` from the attempt to install `python2-libmodulemd` for performance reasons. (No need to retry repeatedly when we expect to fail most of the time.)


mmathesius (approved):
LGTM. Thank you for finding and fixing that.

PR #508, opened 2020-11-18
Upgrade helper

sgallagher:
Add Modulemd.UpgradeHelper object
    
This is the implementation of an object that will hold known
module:stream values to provide a way to resolve ambiguous
upgrades from ModuleStreamV2 dependencies.

### DIFF
--- a/modulemd/include/modulemd-2.0/modulemd-upgrade-helper.h
+++ b/modulemd/include/modulemd-2.0/modulemd-upgrade-helper.h
@@ -1,0 +1,104 @@
+/*
+ * This file is part of libmodulemd
+ * Copyright (C) 2020 Red Hat, Inc.
+ *
+ * Fedora-License-Identifier: MIT
+ * SPDX-2.0-License-Identifier: MIT
+ * SPDX-3.0-License-Identifier: MIT
+ *
+ * This program is free software.
+ * For more information on the license, see COPYING.
+ * For more information on free software, see <https://www.gnu.org/philosophy/free-sw.en.html>.
+ */
+
+#pragma once
+
+#include <glib-object.h>
+
+G_BEGIN_DECLS
+
+/**
+ * SECTION: modulemd-upgrade-helper
+ * @title: Modulemd.UpgradeHelper
+ * @stability: stable
+ * @short_description: Helpers to provide cues to #ModulemdModuleStream
+ * upgrades.
+ */
+
+#define MODULEMD_TYPE_UPGRADE_HELPER (modulemd_upgrade_helper_get_type ())
+
+G_DECLARE_FINAL_TYPE (ModulemdUpgradeHelper,
+                      modulemd_upgrade_helper,
+                      MODULEMD,
+                      UPGRADE_HELPER,
+                      GObject)
+
+
+/**
+ * modulemd_upgrade_helper_new:
+ *
+ * Returns: (transfer full): A newly-allocated #ModulemdUpgradeHelper object.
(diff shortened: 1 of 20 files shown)


sgallagher:
This is the basic object; I still need to work on associating it with the `ModuleIndex` and `Module` objects. But if you could review the object in the meantime, that would be helpful.

modulemd/include/modulemd-2.0/modulemd-upgrade-helper.h
```diff
@@ -0,0 +1,104 @@
...
+
+
+/**
+ * modulemd_upgrade_helper_get_known_streams_as_strv: (rename-to modulemd_upgrade_helper_get_known_streams)
+ * @self: (in): This #ModulemdUpgradeHelper
+ * @module_name: (in): The name of the module to return a list of known streams
+ * for.
+ *
+ * Returns: (transfer full): A list of known streams to provide clues to the
+ * stream upgrade process.
```
mmathesius, line 96:
```suggestion
 * stream upgrade process. NULL if there are no known streams for @module_name.
```
It's probably worth adding a quick test that querying an unknown module returns NULL, too.

mmathesius (commented):
Just one nit-pick on documentation and a suggested test to add.

modulemd/include/modulemd-2.0/modulemd-upgrade-helper.h
sgallagher: I went the other way; it turns out that PyGObject always returns `[ ]` there, even if it gets NULL back. I decided to keep it consistent and I'll return a zero-length array to C.

sgallagher:
I fixed up the requested issue and added the test.

I also included two other patches, another one for fixing up Travis tests (since this PR showed it was still not picking up the overrides properly) and also clean up some memory issues detected by clang.

modulemd/modulemd-yaml-util.c
```diff
@@ -504,6 +504,17 @@ modulemd_yaml_parse_string_set (yaml_parser_t *parser, GError **error)
       yaml_event_delete (&event);
     }
 
+  /* Work around false-positive in clang static analysis which thinks it's
+	   * possible for this function to return NULL and not set error.
+	   */
```
mmathesius, line 509: Odd comment indentation...

mmathesius (approved):
Aside from the noted odd comment indentation, this looks good to me. Thanks!

modulemd/modulemd-yaml-util.c
sgallagher: Fixed

modulemd/modulemd-yaml-util.c
```diff
@@ -504,6 +504,17 @@ modulemd_yaml_parse_string_set (yaml_parser_t *parser, GError **error)
       yaml_event_delete (&event);
     }
 
+  /* Work around false-positive in clang static analysis which thinks it's
+	 * possible for this function to return NULL and not set error.
+	 */
```
mmathesius, line 509: Indentation is still weird--caused by a hard tab at the beginning of those two lines.

sgallagher: Thanks, it looked fine on my editor because the tabs were four spaces wide. Fixed now.

mmathesius: No problem. I thought `clang-format` expanded hard tabs...

sgallagher: It doesn't touch tabs that are inside a comment block.

PR #509, opened 2020-11-19
mmdv3 enhancements

mmathesius:
A small improvement to enable diagnosing which sub-document(s) in a failed batch load are at fault.

Also fix for a bug in `modulemd-validator` that I ran across while borrowing some of its code.


sgallagher (approved):
LGTM, I'll merge it once CI is finished.

PR #511, opened 2020-11-19
mmdv3 improvements

mmathesius:
Background: When setting the default stream_mdversion to 3, a lot of tests are going to need major updating because a lot of the current StreamV2 test data will not upgrade to StreamV3. Either due to lack of any dependencies, or failed stream expansion for modules that specify `[]` or excluded streams. Even with the steam expansion helper addition, tests will still need a lot of work to incorporate that.

With this PR, I have revised many of the tests to make them easier to debug when making future changes for the stream_mdversion 3 default tests. For example, getting a test failure that simply says a function call didn't return TRUE isn't very helpful. Asserting that there is no gerror returned is better to do first, since the test will at least display what the error was.

I also added a function to dump sub-document failures, and called it in many existing tests, so the causes of StreamV2 to StreamV3 conversion failures can be viewed.

I did make one change to non-test code, and that is to detect and report when StreamV2 is being upgraded to StreamV3 and has no dependencies. That was previously being caught later with a difficult to comprehend "Unable to deduplicate expanded dependencies" message.

### DIFF
--- a/modulemd/include/private/modulemd-subdocument-info-private.h
+++ b/modulemd/include/private/modulemd-subdocument-info-private.h
@@ -139,3 +139,16 @@ modulemd_subdocument_info_get_data_parser (ModulemdSubdocumentInfo *self,
                                            yaml_parser_t *parser,
                                            gboolean strict,
                                            GError **error);
+
+
+/**
+ * modulemd_subdocument_info_debug_dump_failures:
+ * @failures: (in) (element-type ModulemdSubdocumentInfo): An array containing
+ * any subdocuments from the YAML file that failed to parse.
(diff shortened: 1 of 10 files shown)


sgallagher (approved):
Thanks! I've slowly been updating tests to compare  `error` before checking for the error case, but I hadn't gotten around to a mass-update.

mmathesius:
I was accidentally returning FALSE instead of NULL in the new error catch I added to `modulemd_module_stream_expand_v2_to_v3_deps()`. Just pushed a fixup for that. Still investigating why tests have decided to fail in Travis but not locally.

PR #515, opened 2020-11-24
Sort expanded dependencies for stream v2 to v3 upgrades

mmathesius:
When modulemd-stream v2 documents are upgraded to modulemd-stream v3, they may undergo stream expansion of the buildtime and runtime dependencies. This PR implements strict sorting of stream expanded dependencies so they are in a sensible, predictable, and reliable order before contexts are auto-generated.

Having this ordering in place also facilitates the implementation of meaningful stream expansion and upgrade tests--which are also provided by this PR.


sgallagher (approved):
Looks great to me!

PR #520, opened 2020-12-04
Drop ModuleStreamV3

sgallagher:
Signed-off-by: Stephen Gallagher <sgallagh@redhat.com>

PR #521, opened 2020-12-04
Add static_context attribute to StreamV2

sgallagher:
This PR depends on #520 

Only the final patch in the set is new.


sgallagher:
Don't review yet; I haven't implemented the YAML parser and emitter changes yet.

sgallagher:
Should be ready to review now.

modulemd/modulemd-module-stream-v2.c
```diff
@@ -2856,6 +2987,10 @@ modulemd_module_stream_v2_emit_yaml (ModulemdModuleStreamV2 *self,
       return FALSE;
     }
 
+  if (modulemd_module_stream_v2_is_static_context (self))
+    {
+    }
```
mmathesius, line 2992: I'm only reviewing the last commit at the moment. Looks like the emit code didn't make it into this commit. Some specific tests for parsing/emitting static_context would be in order--which would have caught this. Everything else that's here looks good too me!

sgallagher: Whoops! :anguished: Yeah, I think I got distracted and forgot to add the emit. Thanks for catching that.

I've added a test as well now.

mmathesius (approved):
The last commit looks good to me now. Thanks!

PR #522, opened 2020-12-07
PackagerV3 updates

mmathesius:
Several updates in preparation for completing the PackagerV3 API implementation:

- Removed a few StreamV3 leftovers
- Included `static_context` in the comparison for `modulemd_module_stream_v2_equals()`
- Moved the PackagerV3 and BuildConfig objects from private to public API
- Added YAML emit capabilities to PackagerV3 object


mmathesius:
@sgallagher I have a question, too. In release 2.10, `modulemd_module_stream_upgrade()` (which returns a single Stream object) was deprecated for new function `modulemd_module_stream_upgrade_ext()` (which returns a Module object). With StreamV3 being rolled back, the reason for the deprecation no longer exists. However, the new _ext() version is used internally and I see no need to change it. But, perhaps, should the deprecation be reverted and leave the two interfaces to stream upgrades equally valid?

sgallagher:
> @sgallagher I have a question, too. In release 2.10, `modulemd_module_stream_upgrade()` (which returns a single Stream object) was deprecated for new function `modulemd_module_stream_upgrade_ext()` (which returns a Module object). With StreamV3 being rolled back, the reason for the deprecation no longer exists. However, the new _ext() version is used internally and I see no need to change it. But, perhaps, should the deprecation be reverted and leave the two interfaces to stream upgrades equally valid?

I think we still want to deprecate it (though we won't remove it). The experiment with StreamV3 showed that the original assumption that upgrades would be one-to-one was faulty, so regardless of the fact that we are not going to have that problem with V1->V2 doesn't change the fact that we need to account for that behavior for potential future cases.

Also, StreamV1 doesn't really exist in the wild, so the upgrade routine is kind of meaningless.

modulemd/modulemd-packager-v3.c
```diff
@@ -1628,3 +1628,146 @@ modulemd_packager_v3_parse_module_components (yaml_parser_t *parser,
...
+  if (!modulemd_yaml_emit_document_headers (
+        emitter, MODULEMD_YAML_DOC_PACKAGER, MD_PACKAGER_VERSION_THREE, error))
+    {
+      return FALSE;
+    }
+
+  /* Start data: */
+  EMIT_MAPPING_START (emitter, error);
+
+  /* TODO: emit data */
```
sgallagher, line 1654: You probably want to drop the TODO here.

modulemd/modulemd-packager-v3.c
```diff
@@ -1628,3 +1628,146 @@ modulemd_packager_v3_parse_module_components (yaml_parser_t *parser,
...
+  if (self->xmd != NULL)
+    {
+      EMIT_SCALAR (emitter, error, "xmd");
+      if (!modulemd_yaml_emit_variant (emitter, self->xmd, error))
+        {
+          return FALSE;
+        }
+    }
+
+  /* TODO: fix this */
```
sgallagher, line 1692: This seems suspicious.

sgallagher (changes requested):
This looks pretty good to me, modulo a few TODO comments left in.

modulemd/modulemd-packager-v3.c
mmathesius: Yup. That's a leftover.

modulemd/modulemd-packager-v3.c
mmathesius: This is also a leftover from my tailed attempt to use `EMIT_HASHTABLE_VALUES_IF_NON_EMPTY()`.

mmathesius:
@sgallagher My latest push removes the extraneous "TODO"s and adds a commit with a batch of documentation fixups.

sgallagher (approved):
LGTM

PR #524, opened 2020-12-08
new PackagerV3 APIs

mmathesius:
This PR provides the additional PackagerV3-related APIs and associated tests:
* `PackagerV3.convert_to_index()`
* `Modulemd.read_packager_file()`
* `Modulemd.read_packager_string()`

A few extraneous StreamV3 test data files are also cleaned up.


modulemd/include/modulemd-2.0/modulemd-packager-v3.h
```diff
@@ -722,4 +723,28 @@ ModulemdComponentRpm *
...
+
+/**
+ * modulemd_packager_v3_convert_to_index:
+ * @self: (in): This #ModulemdPackagerV3 object.
+ * @error: (out): A #GError that will return the reason for a conversion error.
+ *
+ * Note: If buildopts (#ModulemdBuildopts) are in use in one or more build
+ * configurations in the #ModulemdPackagerV3 object @self, only the buildopts
+ * present in the first listed configuration (if any) will be applied to the
+ * #ModulemdModuleStreamV2 object in the returned index.
```
sgallagher, line 735: I think I was unclear in the design document for how this function should work.

In our original StreamV3 plans, we were trying to avoid changing MBS behavior, so we were forced to convert the PackagerV3 into a single StreamV2 object with dependencies in the MSE style. This is the behavior we want to happen transparently if they read in a PackagerV3 document using the [`ModuleStream.read_file()`](https://fedora-modularity.github.io/libmodulemd/latest/ModulemdModuleStream.html#modulemd-module-stream-read-file) or [`ModuleStream.read_string()`](https://fedora-modularity.github.io/libmodulemd/latest/ModulemdModuleStream.html#modulemd-module-stream-read-string) functions. We will need to document this as a limitation of the `ModuleStream.read_*()` API and recommend the use of the `Modulemd.read_packager_*()` API instead (I think by deprecating the `ModuleStream.read_*()` functions in favor of their `Modulemd.read_packager_*()` counterparts).

However, if MBS will use new API here, we can (and should) just convert it into a ModuleIndex of multiple StreamV2 objects, one for each build configuration specified in the PackagerV3. This will allow us to have separate buildopts for each of them.

modulemd/include/private/modulemd-packager-v3-private.h
```diff
@@ -67,28 +67,6 @@ modulemd_packager_v3_to_defaults (ModulemdPackagerV3 *self,
...
- * #ModulemdDefaults object corresponding to the #ModulemdPackagerV3 object
- * @self. NULL if there was an error doing the mapping and sets @error
- * appropriately.
- *
- * Since: 2.11
- */
-ModulemdModuleIndex *
-modulemd_packager_v3_to_stream_v2_ext (ModulemdPackagerV3 *self,
-                                       GError **error);
-
```
sgallagher, line 91: We will probably still want this function around internally so we can handle the implicit conversion for the `ModuleStream.read_*()` functions.

sgallagher (changes requested):
I have no issues with any of the code that you've written here, but I think I miscommunicated to you how the `Modulemd.read_packager_*()` functions should work (see inline comments).

The good news is that you've already implemented the *harder* of the two codepaths, so the other one we need should be very quick.

modulemd/include/modulemd-2.0/modulemd-packager-v3.h
mmathesius: ACK. I'll rename this function back to private function `modulemd_packager_v3_to_stream_v2_ext()` and create a new implantation of `modulemd_packager_v3_convert_to_index()` with that behavior.

modulemd/modulemd-packager-v3.c
```diff
@@ -983,6 +983,198 @@ modulemd_packager_v3_to_stream_v2_ext (ModulemdPackagerV3 *self,
...
+
+      COPY_HASHTABLE_BY_VALUE_ADDER (v2_stream,
+                                     self,
+                                     rpm_components,
+                                     modulemd_module_stream_v2_add_component);
+
+      COPY_HASHTABLE_BY_VALUE_ADDER (v2_stream,
+                                     self,
+                                     module_components,
+                                     modulemd_module_stream_v2_add_component);
```
sgallagher, line 1089: This chunk of code is probably going to need to be reused for the `ModuleStream.read_*()` codepath, so it is probably best to extract it into its own function. Probably not a terrible idea to reuse it for `ModuleStreamV2.copy()` as well. Then we only have to worry about a single place to update in the future.

sgallagher (changes requested):
Looks pretty good to me, though I noted one place where you will probably want to separate a chunk of code and reuse it for the other codepath (and possibly a third place).

modulemd/modulemd-packager-v3.c
mmathesius: Redundancy eliminated for both chunks. But I didn't attempt to re-use it for `ModuleStreamV2.copy()` due to the incompatible object types.

sgallagher (approved):

PR #529, opened 2020-12-11
Work around libseccomp issue in CI

sgallagher:
See https://bugzilla.redhat.com/show_bug.cgi?id=1906575 for more
details.

Fedora Rawhide CI runs were failing (but for unrelated reasons the error was being caught and the CI task was reporting "green"). This works around an issue with Rawhide/F34 and CI hosts running with `libseccomp < 2.4.4`.


mmathesius (approved):
That's a strange problem. But this workaround LGTM!

PR #530, opened 2020-12-11
CI: Add Github Actions for aarch64, ppc64le and s390x

sgallagher:
We cannot run valgrind on these because it is incompatible with
the qemu-static virtualization, but this will allow us to at
least run all of the standard CI tests on those architectures.


mmathesius (commented):
Looks good. Although CI failed for aarch64 with a complaint about `ldd` exiting with status `1`. I haven't dug into that, but could it be related to the `ldd` tweak made by PR #529?

sgallagher:
It appears to be a different ldd issue, I applied the workaround but it's still failing.

PR #531, opened 2020-12-15
Report and drop packager documents encountered in repodata

mmathesius:
If `modulemd-packager` documents are found while reading directly into a module index, issue a warning and drop them.

CC: @mcurlej

### DIFF
--- a/modulemd/modulemd-module-index.c
+++ b/modulemd/modulemd-module-index.c
@@ -155,51 +155,12 @@ add_subdoc (ModulemdModuleIndex *self,
   switch (doctype)
     {
     case MODULEMD_YAML_DOC_PACKAGER:
-      if (modulemd_subdocument_info_get_mdversion (subdoc) <
-          MD_PACKAGER_VERSION_TWO)
-        {
-          g_set_error (error,
-                       MODULEMD_YAML_ERROR,
-                       MMD_YAML_ERROR_PARSE,
-                       "Invalid mdversion for a packager document");
-          return FALSE;
-        }
-
-      if (modulemd_subdocument_info_get_mdversion (subdoc) ==
-          MD_PACKAGER_VERSION_THREE)
-        {
-          packager = modulemd_packager_v3_parse_yaml (subdoc, error);
-
-          index =
-            modulemd_packager_v3_to_stream_v2_ext (packager, &nested_error);
-          if (!index)
-            {
-              g_propagate_error (error, g_steal_pointer (&nested_error));
-              return FALSE;
-            }
-
-          if (autogen_module_name)
-            {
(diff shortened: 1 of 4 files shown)


modulemd/tests/test-modulemd-moduleindex.c
```diff
@@ -223,24 +224,30 @@ module_index_test_read (void)
...
+   * using modulemd_module_index_update_from_file() should trigger a warning
+   * and ignore the document.
    */
   yaml_path = g_strdup_printf ("%s/yaml_specs/modulemd_packager_v2.yaml",
                                g_getenv ("MESON_SOURCE_ROOT"));
+  modulemd_test_signal = 0;
+  signal (SIGTRAP, modulemd_test_signal_handler);
   ret = modulemd_module_index_update_from_file (
     index, yaml_path, TRUE, &failures, &error);
+  g_assert_cmpint (modulemd_test_signal, ==, SIGTRAP);
```
sgallagher, line 239: Why are you expecting a signal here? If we are throwing a signal on unexpected input, that's going to crash DNF...

sgallagher (changes requested):

modulemd/modulemd-module-index.c
```diff
@@ -155,51 +155,10 @@ add_subdoc (ModulemdModuleIndex *self,
...
-          g_clear_object (&index);
-          break;
-        }
-
-      /* Fall through intentionally
-       * We will handle the v2 packager format below
-       */
+      g_warning (
+        "modulemd-packager document ignored while reading into a module "
+        "index");
```
sgallagher, line 160: Don't issue a `g_warning()` here. This will call `assert()` if `G_DEBUG=fatal-warnings`.

sgallagher: What we actually want to do here is treat this the same way we treat an unknown document type below: set the gerror and return `FALSE`.

sgallagher (changes requested):

modulemd/modulemd-module-index.c
mmathesius: I implemented per our chat discussion: "If we see a PackagerV3 in the regular Index read, drop it on the floor (with a g_warning())". Apparently that's not what was intended.

I'll revise tomorrow.

sgallagher: Sorry about that, I did indeed say "use `g_warning()`". I had forgotten that it `assert()`s and that what we really wanted to do was treat it the same as any other unknown subdocument.

mmathesius: No worries. It should be better now.

sgallagher (approved):

Issue #535, opened 2020-12-19
read_packager_file doesn't allow setting module_name


carlwgeorge:
The CentOS team routinely needs to read modulemd files with no name ([example](https://git.centos.org/modules/virt/blob/98fad7bdd0e4c84796ab04535d08dcf516333a34/f/SOURCES/modulemd.src.txt)).  We currently do that using `Modulemd.ModuleStream.read_file` in Python, which allows us to pass in an optional module name argument.

```python
source_modulemd = Modulemd.ModuleStream.read_file('SOURCES/modulemd.src.txt', True, module_name)
```

Since the 2.11 update on my Fedora workstation, I now get a deprecation warning.  The [documentation](https://fedora-modularity.github.io/libmodulemd/latest/ModulemdModuleStream.html#modulemd-module-stream-read-file) says that the recommended replacement is `Modulemd.read_packager_file`.  However, that method does not allow setting the module name during invocation.  Can `Modulemd.read_packager_file` add the ability to pass in an optional module name?

Edit: I spoke with @sgallagher about this and he asked me to provide more context.  The CentOS team has to rebuild RHEL modules based on the modulemd.src.txt files that are exported to https://git.centos.org.  Those files must be passed through a translation tool that converts the internal RHEL git branch names to public git commit references in https://git.centos.org.  That translation tool attempts to add the missing name field, as well as naming the new file using the module name for the expected input to our MBS instance.  That is why we require the ability to manually specify the module name when reading the modulemd.src.txt file.  It exists in the deprecated method, and we would like it to exist in the new method so it is a proper replacement for our needs.

mmathesius:
Something like that should be possible--at least from the Python perspective. However, we need to be careful changing the existing API in the C code.

@mcurlej @sgallagher What are your thoughts on this RFE?

P.S. With the year-end holidays upon us, don't expect an movement on this before the beginning of the new year.

mmathesius:
I had a few spare moments so I worked on this a bit. As noted in the associated draft PR, I have provided the C implementation to go along with this request, but I don't know how to integrate it with the Python API.

mcurlej:
@mmathesius I am ok with this, just please make it so it is the same API as in `Modulemd.ModuleStream.read_file` so it is not mandatory but only optional.

mmathesius:
I figured out how to integrate the enhanced functionality into the Python API--with the extra arguments being optional. See the additional commit in my PR.

[...]

PR #536, opened 2020-12-21
Enable Modulemd.read_packager_file/string() to specify module and stream name overrides

mmathesius:
This is the C implementation to address RFE #535 that adds new `modulemd_read_packager_file_ext()` and `modulemd_read_packager_string_ext()` API calls.

However, the Python implementation is still missing. I am sure it is possible to integrate this with the Python API so the current `Modulemd.read_packager_file()` and `Modulemd.read_packager_string()` methods automatically do the correct thing if extra module and stream name parameters are provided, but I have no idea how to get GObject to do that.

### DIFF
--- a/modulemd/include/modulemd-2.0/modulemd.h
+++ b/modulemd/include/modulemd-2.0/modulemd.h
@@ -362,6 +362,33 @@ modulemd_read_packager_file (const gchar *yaml_path,
                              GError **error);
 
 /**
+ * modulemd_read_packager_file_ext:
+ * @yaml_path: (in): A path to a YAML file containing a packager document.
+ * @object: (out): (transfer full): A newly allocated #ModulemdModuleStreamV2 or
+ * #ModulemdPackagerV3 object initialized with the content from @yaml_path.
+ * @module_name: (in) (nullable): An optional module name to override the
+ * document on disk. Mostly useful in cases where the name is being
+ * auto-detected from git.
+ * @module_stream: (in) (nullable): An optional module stream name to override
+ * the document on disk. Mostly useful in cases where the name is being
+ * auto-detected from git.
+ * @error: (out): A #GError containing additional information if this function
+ * fails in a way that prevents program continuation.
+ *
+ * Returns: @MODULEMD_TYPE_MODULE_STREAM_V2, @MODULEMD_TYPE_PACKAGER_V3, or
+ * @G_TYPE_INVALID. Returns the matching GObject through the @object parameter.
+ * If the return value is @G_TYPE_INVALID, returns the reason as @error.
+ *
+ * Since: 2.11
+ */
+GType
+modulemd_read_packager_file_ext (const gchar *yaml_path,
+                                 GObject **object,
+                                 const gchar *module_name,
+                                 const gchar *module_stream,
+                                 GError **error);
+
+/**
  * modulemd_read_packager_string:
  * @yaml_string: (in): A YAML string containing a packager document.
  * @object: (out): (transfer full): A newly allocated #ModulemdModuleStreamV2 or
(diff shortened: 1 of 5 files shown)


mmathesius:
I figured out the Python GObject implementation to automatically use the enhanced versions for `Modulemd.read_packager_file()` and `Modulemd.read_packager_string()`.

A remaining question, however... Should the enhanced versions of the functions also include a `strict` parameter like `Modulemd.ModuleStream.read_file()` and `Modulemd.ModuleStream.read_string()` have?

sgallagher (changes requested):
You are overthinking the solution here. You should be able to just drop the GObject annotation for `modulemd_read_packager_file()` and add `(rename-to modulemd_read_packager_file)` to the `modulemd_read_packager_file_ext()` implementation.

This will avoid the need to provide a custom override (which we want to minimize whenever possible, since that has to be done per-language-binding, whereas fixing it in the annotations fixes it for all languages.)

sgallagher:
> A remaining question, however... Should the enhanced versions of the functions also include a `strict` parameter like `Modulemd.ModuleStream.read_file()` and `Modulemd.ModuleStream.read_string()` have?

`read_packager_file()` should *always* be strict. We may need to provide a different way to interact with documents as requested in #535, however. I need to understand better what they're using that information for. We should define the API around their use-case.

mmathesius:
> You are overthinking the solution here. You should be able to just drop the GObject annotation for `modulemd_read_packager_file()` and add `(rename-to modulemd_read_packager_file)` to the `modulemd_read_packager_file_ext()` implementation.
> 
> This will avoid the need to provide a custom override (which we want to minimize whenever possible, since that has to be done per-language-binding, whereas fixing it in the annotations fixes it for all languages.)

It doesn't seem to be possible to drop the custom Python overrides for `Modulemd.ModuleStream.read_file()` and `Modulemd.ModuleStream.read_string()` because otherwise they will be returning a tuple again--which is what originally necessitated adding the overrides with the `ModulemdUtil.strip_gtype()` wrapper trickiness.

sgallagher:
> It doesn't seem to be possible to drop the custom Python overrides for `Modulemd.ModuleStream.read_file()` and `Modulemd.ModuleStream.read_string()` because otherwise they will be returning a tuple again--which is what originally necessitated adding the overrides with the `ModulemdUtil.strip_gtype()` wrapper trickiness.

I didn't say drop the overrides for `read_packager_file()`, I said we can just have the extended version of the function *be* read_packager_file() so far as the language bindings are concerned.

sgallagher (approved):
LGTM

The test failures appear to be a bug in the test runner that I need to track down.

sgallagher:
Actually... hang on. I'm somewhat concerned about the packit failures in the same place. We need to figure out why that's broken...

mmathesius:
> I think I found the issue. Have a look at [sgallagher@754ed25](https://github.com/sgallagher/libmodulemd/commit/754ed250fc10b8fd13a2da2937b582d8e20270b7)
> 
> Basically, we need to skip the test if the current installed environment doesn't have the appropriate override.

D'oh! That problem again. I should have remembered to include that.

Would you like me to add your commit to this PR? Actually, I think it would make it a little easier to comprehend if the test functions started out with the logic reversed, such as:
```
     if not (
         "_overrides_module" in dir(Modulemd)
         and hasattr(gi.overrides.Modulemd, "read_packager_file")
     ):
         return
```
What do you think?

> I still need to figure out why the "Other Architectures" aren't picking up the modified overrides, but that's a different problem.

I can try to look into that.

sgallagher:
> > I think I found the issue. Have a look at [sgallagher@754ed25](https://github.com/sgallagher/libmodulemd/commit/754ed250fc10b8fd13a2da2937b582d8e20270b7)
> > Basically, we need to skip the test if the current installed environment doesn't have the appropriate override.
> 
> D'oh! That problem again. I should have remembered to include that.
> 
> Would you like me to add your commit to this PR? Actually, I think it would make it a little easier to comprehend if the test functions started out with the logic reversed, such as:
> 
> ```
>      if not (
>          "_overrides_module" in dir(Modulemd)
>          and hasattr(gi.overrides.Modulemd, "read_packager_file")
>      ):
>          return
> ```
> 
> What do you think?

If you think it reads more cleanly that way, be my guest. I'd prefer if you integrated my changes (or this variant) directly into the patch that creates `common.py`. Otherwise it could potentially break a future `git bisect`.
> 
> > I still need to figure out why the "Other Architectures" aren't picking up the modified overrides, but that's a different problem.
> 
> I can try to look into that.

It's not urgent and we don't often change the overrides. We can file a bug on it and look into it whenever we get a free hour.

PR #540, opened 2021-01-12
When adding obsoletes ensure index has stream mdversion at least 2

kontura:
We were getting ready to finally merge the obsoletes part in libdnf and dnf but running the tests I have created before showed this issue with the new libmodulemd version.

When there is an index with just obsoletes it doesn't have a `stream_mdversion` and its not possible to add streams due to: https://github.com/fedora-modularity/libmodulemd/blob/main/modulemd/modulemd-module.c#L381

This is currently blocking us and if I understand correctly there is no workaround.

### DIFF
--- a/modulemd/modulemd-module-index.c
+++ b/modulemd/modulemd-module-index.c
@@ -1602,6 +1602,17 @@ modulemd_module_index_merge (ModulemdModuleIndex *from,
             }
         }
 
+      if (obsoletes_array->len > 0)
+        {
+          /* Obsoletes need at least MD_MODULESTREAM_VERSION_TWO */
+          if (!modulemd_module_index_upgrade_streams (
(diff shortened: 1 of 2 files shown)


j-mracek:
@sgallagher May I ask you for a review and a release ASAP? It resolves a critical problem. Thanks a lot.

sgallagher:
I think this is probably the wrong place to make this change. We want to set the version as soon as the Obsoletes is added to the Index, not wait until we are attempting to merge it. I'll see about creating a better solution.

sgallagher:
I merged #541 and I've built updated Fedora and EPEL 7 packages and submitted them as updates.

kontura:
> I merged #541 and I've built updated Fedora and EPEL 7 packages and submitted them as updates.

@sgallagher Thank you!

PR #541, opened 2021-01-12
When adding obsoletes ensure index has stream mdversion at least 2

sgallagher:

@kontura @j-mracek

This modifies #540 so that the check occurs when the Obsoletes data is added to the ModuleIndex, rather than waiting until a merge is attempted. I left your test unmodified to demonstrate that it still works to solve the same problem.


kontura:
I see, so with this version the check in `modulemd_module_index_merge` is no longer needed right? Was it left here by mistake?

I have tested it with the check just in `modulemd_module_index_add_obsoletes` and it works and even our tests for dnf using obsoletes work. :+1:

sgallagher:
Hmm, I thought I removed that section when I squashed my fixup, but I guess I screwed up the merge. One moment.

sgallagher:
Pushed with that section dropped.

Issue #539, opened 2021-01-11
packager format does not accept buildorder


mikem23:
I performed a simple test of using the packager format by taking a simple modulemd yaml file and changing the document value to modulemd-packager. The yaml was rejected with this error:

```
GLib.Error('Unexpected key in rpm component body: buildorder [line 34 col 9]', 'modulemd-yaml-error-quark', 8)
```

It seems like the packager format should support any field that a developer would need to specify.

mikem23:
This makes me wonder what other necessary fields not be rejected. I haven't run exhaustive tests. Is there a clear way to compare the two formats?

sgallagher:
> This makes me wonder what other necessary fields might be rejected. I haven't run exhaustive tests. Is there a clear way to compare the two formats?

This should actually be the only one that we dropped and it was semi-intentional.

We've been planning for the last [two years](https://pagure.io/fm-orchestrator/issue/1216) to support `buildafter` as a replacement for `buildorder`, so when I originally drafted the PackagerV2 format, I dropped `buildorder` from it. However, MBS has still not actually implemented support for `buildafter` and I overlooked that when we created the PackagerV3 format, large sections of which are shared with PackagerV2.

So we have two options here:

1) Reintroduce support for `buildorder` into PackagerV3
2) Finally implement `buildafter` in MBS.

I'd prefer that we implement `buildafter`, since it will be far simpler for packagers to understand and maintain than the `buildorder`. However, if this is impossible to fit into the schedule, we can extend PackagerV3 to support it the same way that ModuleStreamV2 did.

mcurlej:
@sgallagher So i am confused was this removed in v2 or in v3?

sgallagher:
It was removed in the PackagerV2 format, which we never ended up using in Fedora. The ModuleStreamV2 format still supported both `buildorder` and `buildafter`.

We then did not implement `buildorder` into the PackagerV3 format because it was not part of PackagerV2, which we were using as our baseline.

To proceed from here, we need to do one or the other of the options I described [above](https://github.com/fedora-modularity/libmodulemd/issues/539#issuecomment-758286707).

mcurlej:
@sgallagher oh ok, thanks for the explanation. I am for the 2nd option, as `buildorder` was discontinued, some time ago.

mcurlej:
[...]

PR #543, opened 2021-01-13
[WIP] Spec updates for reintroducing 'buildorder' in PackagerV3

mmathesius:
This PR contains the proposed YAML spec file updates for reintroducing `buildorder` in PackagerV3 to address #539.

This PR is for review and comment only. DO NOT MERGE, as it will cause CI to fail until the code implementation is complete.

CC: @mikem23 @mcurlej


sgallagher (changes requested):
Please update the PackagerV2 spec as well. We should be consistent there (even if we never end up using that document).

sgallagher (approved):
LGTM

mmathesius:
Closing this PR since the updated specs are now included with the implementation PR #544.

PR #544, opened 2021-01-13
Enable component 'buildorder' for modulemd-packager v2 and v3 documents

### LINKED ISSUE #539: packager format does not accept buildorder

mmathesius:
Spec, implementation, and test updates to resolve #539.

CC: @mikem23 @mcurlej


sgallagher (approved):
Looks good to me. Great work!

PR #545, opened 2021-01-13
Autoformatting improvements

sgallagher:
We don't need to stand up a heavy container image just to run the
autoformatters. We'll just use the ones available in the host
environment.

Also switch to https://github.com/marketplace/actions/add-commit
for handling the git commit.

The second patch drops the formatters from the internal test suite,
simplifying it greatly.

The last patch was autogenerated due to tweaks made by python-black.


mmathesius (changes requested):
Nice optimization and cleanup!

Somehow I overlooked the `add-commit` action in the Actions Marketplace in my previous iteration, so I kind of rolled my own.

Just one nitpick to give OpenMandriva a better chance of having working CI again someday.

.github/workflows/formatters.yaml
```diff
@@ -0,0 +1,30 @@
+name: Automatic Code Formatters
+on: push
```
mmathesius, line 2: Also checking... This is intended to run on _every_ push? Not just pushes to the main branch?

sgallagher: Yes, that was my intention. I want it to show up when someone pushes to their private branches as well, so they might notice the extra patch before sending a pull request.

.ci/openmandriva/ci-tasks.sh
```diff
@@ -21,6 +20,12 @@ meson test --suite ci \
            --print-errorlogs \
            -t 5
 
+# Disable the verbose tests and run the supported tests through valgrind
+CC=clang CXX=clang++ meson --buildtype=debugoptimized \
+      -Dverbose_tests=false \
+      $COMMON_MESON_ARGS \
+      ci
```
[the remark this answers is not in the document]
sgallagher: Thanks for catching that.

.github/workflows/formatters.yaml
mmathesius: ACK. But I'm going to miss having `ninja test` clean up formatting _before_ pushing.

mmathesius (approved):

.github/workflows/formatters.yaml
sgallagher: Train yourself to use `ninja clang-format` and you'll be mostly fine. I don't want to implement a `python-black` runner mostly because different versions of it make different decisions. Hence why we're only running it once in github actions rather than on each distro like we did in Travis CI.

sgallagher:
I just made one minor change and added `--target-version py36` to the python auto-formatter so that it doesn't ever accidentally reformat the python code in a way that would break libmodulemd on RHEL 7.

mmathesius:
> I just made one minor change and added `--target-version py36` to the python auto-formatter so that it doesn't ever accidentally reformat the python code in a way that would break libmodulemd on RHEL 7.

:+1:

PR #546, opened 2021-01-14
Build documentation with Github Actions

sgallagher:
Also ensures that static analysis and documentation generation will
occur only on the main branch of the upstream repository.

I also modified the doc generation script so that in the future we can
add an "on-release" Action that will generate a snapshot of the API
docs automatically.

### DIFF
--- a/.ci/docs/ci-tasks.sh
+++ b/.ci/docs/ci-tasks.sh
@@ -26,26 +26,11 @@ pushd doc-generation/modulemd/html
 /builddir/contrib/doc-tools/fix-xref.sh
 popd
 
-$RETRY_CMD git clone https://sgallagher:$DOC_TOKEN@github.com/fedora-modularity/fedora-modularity.github.io
-rsync -avh --delete-before --no-perms --omit-dir-times /builddir/doc-generation/modulemd/html/* fedora-modularity.github.io/libmodulemd/latest
+MODULEMD_VERSION=${MODULEMD_VERSION:-latest}
 
-pushd fedora-modularity.github.io
-
-git add libmodulemd/latest
-
-# Check to see if there are any changes
-set +e
-git commit -m "Updating libmodulemd docs for $GITHUB_SHA" --dry-run
-err=$?
-if [ $err = 0 ]; then
-    set -e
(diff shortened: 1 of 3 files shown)


.github/workflows/upstreamed.yaml
```diff
@@ -20,12 +21,34 @@ jobs:
...
+      - name: Checkout code repo
         uses: actions/checkout@v2
 
-      - name: Publish
-        env:
-          DOC_TOKEN: ${{ secrets.DOC_TOKEN }}
+      - name: Checkout documentation repo
+        uses: actions/checkout@v2
+        with:
+          repository: ${{ github.actor }}/fedora-modularity.github.io
```
mmathesius, line 33: Will this work when run by anyone except `sgallagher`? (Unless the person triggering this also has an appropriate fork and updates `secrets.DOC_TOKEN` to match. At least that's my understanding of how this works.)

Perhaps there should also be a `secrets.DOC_USER` that pairs with `secrets.DOC_TOKEN` that gets used here? (I can't seem to find a way to set per-repository non-secret variables other than hard coding them in the workflow file.)

sgallagher: Oops, that was testing code. That should just say `fedora-modularity`. I was using my own fork to test and that was an easy way to do it...

.ci/docs/ci-tasks.sh
```diff
@@ -26,26 +26,9 @@ pushd doc-generation/modulemd/html
...
-    git commit -m "Updating libmodulemd docs for $GITHUB_SHA"
-    $RETRY_CMD git push origin master
-fi
-set -e
-
-popd #fedora-modularity.github.io
+mkdir -p /builddir/fedora-modularity.github.io/libmodulemd/$MODULEMD_VERSION
+rsync -avh --delete-before --no-perms --omit-dir-times \
+    /builddir/doc-generation/modulemd/html/* \
+    /builddir/fedora-modularity.github.io/libmodulemd/$MODULEMD_VERSION
```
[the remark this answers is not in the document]
sgallagher: I'll just have it default to "latest".

mmathesius (approved):

Issue #549, opened 2021-02-12
Regex parsing for context in Stream/Packager v2


mcurlej:
Hi All,

we need to adjust the regex for context parsing. We need to investigate the regex which parses context.  The regex needs to be `a-zA-Z0-9.-_` to accomodate for contex suffixes. Can you take a look?

The specs:
https://github.com/fedora-modularity/libmodulemd/blob/main/yaml_specs/modulemd_stream_v2.yaml#L91

sgallagher:
There is no regex there. See https://github.com/fedora-modularity/libmodulemd/blob/main/modulemd/modulemd-module-stream-v2.c#L1237

You should be able to just add a check for `.`, `-`, and `_` to https://github.com/fedora-modularity/libmodulemd/blob/main/modulemd/modulemd-module-stream-v2.c#L1260

(Also, if 10 characters is insufficient, update MMD_MAXCONTENTLEN).

If you change this parsing, please also update the specification for the stream and packager YAML formats.

ppisar:
I believe the requested change is available now in commit 0132015b5729b6077d49fc5beb2c662a563c6e6d.

ppisar:
The tests added in  #550 does not work on RHEL 8.  #552 should fix them.

ppisar:
This issue was resolved with 0132015b5729b6077d49fc5beb2c662a563c6e6d and 6d02f70dc6d958c1d5743ec9ee8794459e671fd1 commits.

PR #550, opened 2021-04-13
modulemd v2: Relax context up to 13 characters including underscore

ppisar:
cb507415b158b79df44d9b11ca69642c2c546f0b (Add BuildConfig object)
commit broke scratch builds:

$ rhpkg module-scratch-build --file perl-YAML.yaml --buildrequires platform:el8.5.0
Submitting the module build...
Could not execute module_scratch_build: The build failed with:
modulemd-error-quark: Could not validate stream to emit: Non-alphanumeric character in stream context 'a6d43775_1' (1)

The reason is that a scratch build automatically adds an underscore
suffix (e.g.  "_1)" to the context value.

This patch fixes it by allowing the underscore character and
prolonging a limit of 10 characters with 3 additional characters (i.e.
up to 99 scratch builds). We keep v3 context strict as it was (no
underscore, 10 characters) to prevent packagers from specifying
a context in v3 which could not be scratch-built later.
A specification was amended accordingly.

This dichotomy is dictated by the fact that v3 is an input-only
format, while v2 is both input and output format for MBS.

The implemented tests discovered that reading v2 document with
a static context does not perform a validation in contrast to reading
a v3 document. This patch does not change it. It instead marks the
affected tests as TODO and test an explicit validation in addition.

<https://github.com/fedora-modularity/libmodulemd/issues/549>

### DIFF
--- a/modulemd/include/private/modulemd-module-stream-v2-private.h
+++ b/modulemd/include/private/modulemd-module-stream-v2-private.h
@@ -13,6 +13,7 @@
 
 #pragma once
 
+#include "modulemd-module-stream.h"
 #include "modulemd-module-stream-v2.h"
 #include "modulemd-subdocument-info.h"
 #include <glib-object.h>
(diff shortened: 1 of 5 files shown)


sgallagher:
The changes look fine to me by inspection, but could you rebase this atop the current main branch? I've just pushed some fixes to the test runner.

ppisar:
No problem. Rebased.

ppisar:
Any progress with merging this code?

PR #552, opened 2021-04-22
tests: Do not use g_test_incomplete() broken in old glib

ppisar:
The context validation tests dealt with v2 format not being valided
implicitly by marking the tests as expected to fail.

But g_test_incomplete() seems to be broken in glib2-2.56.4. It will
return a failure exit code if a test expected to fail indeed fails.
This is not how 2.68.1 behaves.

This patch accommodates to both behaviours by not using
the broken g_test_incomplete() at all.

Related to
<https://github.com/fedora-modularity/libmodulemd/issues/549>.

This fixes running the tests on RHEL 8.


ppisar:
Any progress with merging this patch?

PR #561, opened 2021-05-03
validator: Handle a failed g_setenv()

ppisar:
g_setenv() can fail and then --debug option would lost its effect.
This patch raises an error in that case.

CID 370512


ppisar:
CI assessment: Arch Linux fails on downloading Arch base image. F35 Installed RPM Tests fail on a segfault in packit (probably regression in librpm). Overall the tests are fine.

PR #568, opened 2021-06-04
Python support requires GIR support

ppisar:
Previously, configuring and running the tests could fail because
GIR data were not built and thus the tests exhibited installed GIR
data mismatching tests and the library. E.g.:

modulemd/tests/ModulemdTests/modulestream.py", line 504, in test_demodularized_rpms
    stream.add_demodularized_rpm("foo")
AttributeError: 'ModuleStreamV2' object has no attribute 'add_demodularized_rpm'

A trivial fix was adding a build target for GIR as a dependency to the
Python tests. But that arrised a more profound bug -- Let's disable GIR
with -Dskip_introspection=true, then the build fails because Python3
tests are always performed and Python3 binding uses GIR. We simply
cannot perform Python tests without building GIR.

This patch fixes it by forbiding disabling GIR while keeping Python2 or
Python3 enabled. And it also disables the Python3 tests if the Python3
binding is disabled.

### DIFF
--- a/.ci/fedora/ci-tasks.sh
+++ b/.ci/fedora/ci-tasks.sh
@@ -55,6 +55,7 @@ else
     set -e
     meson --buildtype=debug \
           -Dskip_introspection=true \
+          -Dwith_py3=false \
           ci_scanbuild
 
     pushd ci_scanbuild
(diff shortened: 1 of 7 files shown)


sgallagher (approved):
LGTM :+1:

PR #574, opened 2021-07-23
Add centos-stream-9 to Packit configuration

ppisar:
Test Packit with CentOS 9 Stream with this pull request.


sgallagher:
Looks like it didn't actually trigger a build for c9s...

PR #575, opened 2021-07-26
modulemd-module: Document modulemd_module_get_defaults can return NULL

jlebon:
We weren't handling this properly in libdnf. Let's document it to make
it explicit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,9 +103,12 @@ specification for a full list of information that can be retrieved.
 ModulemdModule *module =
   modulemd_module_index_get_module (merged_index, "modulename");
 ModulemdDefaults *defaults = modulemd_module_get_defaults (module);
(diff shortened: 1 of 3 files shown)


ppisar:
Thanks for improving the documentation. I will merge it so that people won't get warnings when calling modulemd_defaults_v1_get_default_stream()  on NULL.

But in general, any libmodulemd *_get_*() function can return NULL meaning that the retrieved value is undefined. E.g. a subsequent modulemd_defaults_v1_get_default_stream() whose return value is printed in the patched example can also return NULL.

jlebon:
Thanks for the review!

> But in general, any libmodulemd __get__() function can return NULL meaning that the retrieved value is undefined.

I think it'd be helpful to document all of these to be explicit.

PR #576, opened 2021-09-08
modulemd-packager-v3: Reject duplicate context

ppisar:
A specification requires modulemd-packager-v3 documents to have
context values unique among all configurations:

document: modulemd-packager
version: 3
data:
  name: foo
  stream: bar
  license: [MIT]
  configurations:
    - context: A
      platform: 1
    - context: A
      platform: 2

That was not enforced and previous configurations were swallowed by the last
one.

This patch fixes it.

I was tempting to augment modulemd_packager_v3_add_build_config(), but
that cannot change a prototype because it's passed to
COPY_HASHTABLE_BY_VALUE_ADDER(). A wrapper for an error-returning
variant would be too cluttery and glib < 2.40 does not report
a replacement. Next, augmenting modulemd_build_config_parse_yaml()
would bring the benefit of reporting a line number, but feeding
a packager object down to the YAML parser would be against separating
a data flow. Thus I simply added a condition just before calling
modulemd_packager_v3_add_build_config().

### DIFF
--- a/modulemd/modulemd-packager-v3.c
+++ b/modulemd/modulemd-packager-v3.c
@@ -1403,6 +1403,7 @@ modulemd_packager_v3_parse_build_configs (yaml_parser_t *parser,
   gboolean done = FALSE;
   g_autoptr (ModulemdBuildConfig) buildconfig = NULL;
   g_autoptr (GError) nested_error = NULL;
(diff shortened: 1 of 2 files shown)


sgallagher (approved):
LGTM

PR #583, opened 2022-01-26
Add support for stripping XMD from an Index

sgallagher:
By specification, the XMD contents are not meant to be used by
consumers. This will provide a simple method of stripping that content
out from the metadata of all streams.

This could be used to sanitize the module metadata prior to including it
in a composed RPM repository.

### DIFF
--- a/modulemd/include/modulemd-2.0/modulemd-module-index.h
+++ b/modulemd/include/modulemd-2.0/modulemd-module-index.h
@@ -667,4 +667,19 @@ modulemd_module_index_upgrade_defaults (ModulemdModuleIndex *self,
                                         GError **error);
 
 
+/**
+ * modulemd_module_index_clear_xmds:
+ * @self: This #ModulemdModuleIndex object.
+ *
+ * Iterates through all #ModulemdModuleStream entries in this
+ * #ModulemdModuleIndex and removes any XMD sections that are present. This is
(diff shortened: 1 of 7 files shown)


ppisar:
There are these three new public API functions:

modulemd_module_index_strip_xmd
modulemd_module_strip_xmd
modulemd_module_stream_v2_clear_xmd

Is there a reason why two of them use "strip" and one of them use "clear" verb? I'd rather see "clear" everywhere as it was already heavily used before. The API should be consistent. And maybe module-level functions could you plural to emphases it removes all xmd objects:

modulemd_module_index_clear_xmds
modulemd_module_clear_xmds
modulemd_module_stream_v2_clear_xmd

Otherwise, I'm fine with this code.

sgallagher:
Thanks, that's a good point. I've just pushed an update that does exactly as you suggest.

sgallagher:
My intention is to propose that we modify createrepo_c to consume this function when saving module metadata to a repository. That way, when we compose a repo, we aren't leaking internal buildsystem details into the public metadata.

ppisar:
Great. I bumped a minor version of libmodulemd and rebased your commit  on top. Your code is in main branch now.
I imaged the use case because I've already reported leaking XMDs to ubi8 repositories from the very initial RHEL 8 module builds and I was rejected with an explanation that Red Hat is not going to redact the historical builds.

PR #585, opened 2022-02-21
Update libmodulemd.spec

JustALawnGnome7:
The "master" branch is no longer the default branch for libmodulemd, which means the description of the package now contains a broken link. Branch "master" has been replaced by "main" as the default branch.


sgallagher (approved):
Nice catch, thanks.

PR #587, opened 2022-03-15
Double-quote string sets and profiles in modulemd-defaults

m10k:
When emitting YAML, libmodulemd does not quote values in (build)dependencies
and default profiles. These fields commonly contain version numbers such as
1.10, which are treated as floating point values by non-libmodulemd parsers,
causing any trailing zeroes to be dropped and the generated YAML file to be
incorrect.

This commit addresses the issue by forcing the YAML-emitting code to emit
double-quoted scalars when emitting string sets and profiles from a
modulemd-defaults-v1 object.


sgallagher:
You will need to update the tests as well to expect the quotes.

m10k:
Sorry, I forgot about those.. I have adjusted all test cases and made sure that they pass when I run them locally.

ppisar:
Thanks for the patch. I will merge it because it improves the situation with profiles and streams.

It unnecessarily quotes all list items (license, rpms, artifacts). That's a less important disturbance. It can be tackled later by only quoting values which look like a non-interger.

m10k:
Thank you!

I tried to keep the change set as small as possible because I'm not familiar with the code base, but that ended up affecting the output of other fields as well. I will create another PR that quotes the output only where it's absolutely necessary, it will probably take a bit longer though.

ppisar:
CI failures are unrelated  (Arch Linux: <https://bugs.archlinux.org/task/74037>, OpenMandriva Cooker <https://github.com/OpenMandrivaAssociation/distribution/issues/2698>). Merging.

PR #588, opened 2022-03-31
tests: Check a return value of modulemd_module_index_update_from_file…

ppisar:
…() in test-modulemd-merger

Discovered by Covscan.
<https://scan4.scan.coverity.com/reports.htm#v52774/p12206/fileInstanceId=115824592&defectInstanceId=14419202&mergedDefectId=353648>


ppisar:
The OpenSuSE failure (missing glib/index.html) is unrelated.

PR #590, opened 2022-04-06
Fix a crash when converting a modulemd-packager with a default profile without a module name or stream

ppisar:
modulemd_packager_v3_convert_to_index() called on a ModulemdPackagerV3
object with an automatic name or a stream and with a default profile
aborted:

~~~~
        (process:41508): libmodulemd-CRITICAL **: 14:54:41.578: modulemd_defaults_set_module_name: assertion 'module_name' failed

        (process:41508): libmodulemd-CRITICAL **: 14:54:41.578: modulemd_defaults_v1_add_default_profile_for_stream: assertion 'stream_name' failed
        Neoprávněný přístup do paměti (SIGSEGV) (core dumped [obraz paměti uložen])
~~~~

That's because a default profile is converted to the index as a
ModulemdDefaultsV1 object and that object requires the modules name and stream.

This patch fixes it by returning a proper error. It also amends
a documentation for modulemd_packager_v3_convert_to_index() so that
users are aware of the requirements.


sgallagher (approved):
This solution looks correct to me. :+1:

PR #594, opened 2022-04-13
Remove installation dependency on six

fuanan-3:
If we give up support for python2.
In python3 architecture,text_type no longer support,str compatible text_type.
see:https://six.readthedocs.io/#six.text_type
Therefore,we no longer need to rely on the six module.

### DIFF
--- a/bindings/python/gi/overrides/Modulemd.py
+++ b/bindings/python/gi/overrides/Modulemd.py
@@ -16,7 +16,6 @@ from ..overrides import override
 
 import functools
 
-from six import text_type
 from gi.repository import GLib
 
 import datetime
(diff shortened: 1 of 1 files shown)


sgallagher:
Dropping support for Python 2 means not being able to build for RHEL/CentOS
7, which could be problematic for Satellite/Pulp deployments on those
systems.

fuanan-3:
Yes, I agree with you,and I will close this pull request.
The modification applies to OS where python2 has been deleted.

ppisar:
Fuanan-3, I can see your point: If you set up libmodulemd without Python2, you don't need six because in Python 3 the instance will be of type str and covered by a if-branch above (isinstance(obj, str)). But as sgallagher correctly noticed, libmodulemd still has Python2 users and cannot unconditionally remove the support.

If you want to get rid of six in case of Python3-only build, then please rewrite your patch so that it keeps working for Python2 configurations.

Issue #593, opened 2022-04-13
Empty profile is parsed as a package with an empty name


ppisar:
Having this document (empty profiles are legal <https://docs.fedoraproject.org/en-US/modularity/policies/packaging-guidelines/#_profiles>):

~~~~
document: modulemd
version: 2
data:
    name: test
    stream: A
    summary: Test.
    description: >
        Test.
    license:
        module: [ MIT ]
    profiles:
        profileA:
          rpms:
~~~~

reading and writing results into:

~~~~
document: modulemd
version: 2
data:
  name: test
  stream: "A"
  summary: Test.
  description: >
    Test.
  license:
    module:
    - MIT
  profiles:
    profileA:
      rpms:
      -
~~~~

This is a bug on parser side:

~~~~
$ cat /tmp/test.py 
#!/usr/bin/python3
import gi
gi.require_version('Modulemd', '2.0')
from gi.repository import Modulemd

module = Modulemd.read_packager_file('/tmp/out')
print("name={}".format(module.get_module_name()))
profile = module.get_profile('profileA')
print("profile={}".format(profile.get_name()))
packages = profile.get_rpms()
for package in packages:
    print("<{}>".format(package))
[test@fedora-37 libmodulemd-devel]$ GI_TYPELIB_PATH=/tmp/build/modulemd LD_LIBRARY_PATH=/tmp/build/modulemd PYTHONPATH=$PWD python3 /tmp/test.py 
name=test
profile=profileA
<>
~~~~

Possibly all rpms lists are affected.

sgallagher:
Your original document is faulty.

```
  profiles:
    profileA:
      rpms:
      -
```
Literally says that there is one entry under `rpms:` and it's an empty scalar.

What you actually want is:
```
  profiles:
    profileA:
      rpms: []
```
This will give you a zero-length array, which will behave as you expect. The docs may need to clarify this, but the way it's behaving is as expected.

ppisar:
My original document is without the trailing hyphen:

~~~~
  profiles:
    profileA:
      rpms:
~~~~

sgallagher:
That's still incorrect, because we expect a YAML_SEQUENCE_START here and we get nothing. The libyaml is probably just trying to treat the lack of content here as a zero-length scalar, but that's pretty much just trying to avoid aborting on an invalid syntax.

PR #595, opened 2022-04-14
Fix parsing empty profiles

ppisar:
Parsing a profile with an undefined rpms list:

~~~~
  document: modulemd
  version: 2
  data:
      name: test
      stream: A
      summary: Test.
      description: >
          Test.
      license:
          module: [ MIT ]
      profiles:
          profileA:
            rpms:
~~~~

lead to a nonempty list containing an empty-name package:

~~~~
  profiles:
    profileA:
      rpms:
      -
~~~~

while user probably wanted:

~~~~
  profiles:
    profileA:
      rpms: []
~~~~

This patch changes interpretation of the top-most excerpt from the
middle one to the last one.

https://github.com/fedora-modularity/libmodulemd/issues/593

### DIFF
--- a/modulemd/modulemd-yaml-util.c
+++ b/modulemd/modulemd-yaml-util.c
@@ -592,15 +592,32 @@ modulemd_yaml_parse_string_set (yaml_parser_t *parser, GError **error)
         case YAML_SCALAR_EVENT:
           g_debug ("Parsing scalar: %s",
                    (const gchar *)event.data.scalar.value);
-          g_hash_table_add (result,
-                            g_strdup ((const gchar *)event.data.scalar.value));
 
-          if (!in_list)
+          if (in_list)
             {
-              /* We got a scalar instead of a sequence. Treat it as a list with
-               * a single entry
-               */
-              done = TRUE;
+              g_hash_table_add (
+                result, g_strdup ((const gchar *)event.data.scalar.value));
+            }
+          else
+            {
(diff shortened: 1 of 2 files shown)


modulemd/modulemd-yaml-util.c
```diff
@@ -592,15 +592,32 @@ modulemd_yaml_parse_string_set (yaml_parser_t *parser, GError **error)
...
+                {
+                  /* Treat it as a list with a single entry
+                   * if it's nonempty. */
+                  g_hash_table_add (
+                    result, g_strdup ((const gchar *)event.data.scalar.value));
+                  done = TRUE;
+                }
+              else
+                {
+                  /* Otherwise it's a bug in libyaml which misreports an empty
```
sgallagher, line 614: It's not technically a bug in libyaml. The [YAML specification](https://yaml.org/spec/1.2.2/#72-empty-nodes) actually requires this explicitly. Best not to confuse a future reader with the assertion that it's a bug. No need to send for re-review just for this.

sgallagher (approved):

modulemd/modulemd-yaml-util.c
ppisar: I corrected the comment. Though I still understand the specification as describing two distinct cases, one resulting into an empty string (7.2) and the other one resulting to an undefined value (7.3). Interesting is that YAML::XS can tell the distinction despite being built on top of the very same libyaml.

PR #600, opened 2022-07-26
Add ARM tests

sgallagher:

### DIFF
--- a/.github/workflows/multiarch.yaml
+++ b/.github/workflows/multiarch.yaml
@@ -15,11 +15,13 @@ jobs:
             distro: fedora_latest
           - arch: s390x
             distro: fedora_latest
+          - arch: aarch64
+            distro: fedora_latest
 
     steps:
       - uses: actions/checkout@v2
 
-      - uses: uraimo/run-on-arch-action@v2.0.7
+      - uses: uraimo/run-on-arch-action@v2
         name: Perform upstream tests
 
         with:


ppisar:
I'm not sure whether fedora_latest on armv7 makes sense. Fedora 36 is the last release built for armv7.

ppisar:
The distribution comes from <https://hub.docker.com/r/arm32v7/fedora/#!>. That means it will be available until F36 end of life. Slightly confusing that other architectures will move to F37 etc. (Compare to <https://hub.docker.com/r/arm64v8/fedora#!>.) I'm fine with it. Maybe a comment there would be helpful when, a years later, somebody will investigate a mysterious difference between architectures.

sgallagher:
I think you're right, I'll drop the armv7 portion.

ppisar:
A nit pick: Please move the new aarch64 entry before ppc64le entry to preserve a lexical ordering.

ppisar:
The newly added test fails (a segfault in ldd). I'm not able to reproduce it locally (up-to-date F36 without updates-testing on aarch64 machine). I installed all build dependencies for rawhides' libmodulemd.spec, cloned this pull request, and followed all commands from the "sed" command (included) in "run" section in .github/workflows/multiarch.yaml.

While I understand that the new test is not a cause of the failure, I'm not going to accept a change which breaks CI. Please fix the failure first.

ppisar:
Maybe the container is out-dated and contains a broken glibc. <https://hub.docker.com/r/arm64v8/fedora/tags> is from 2022-05-12.

PR #601, opened 2022-07-26
Add support for stream overrides in configurations

### LINKED ISSUE #599: Packager v3: Extend configurations to allow stream renaming
It would be useful to add an optional `stream:` key to the `data.configurations` section of the Packager v3 format. This would allow us to submit the module with a custom stream name for certain contexts.

As a real-world example, I currently maintain a variant of the `16` stream for EPEL 9 called `16-epel` which is needed because the RHEL module does not provide `v8-devel`, needed for some EPEL packages. Right now, I have to maintain a completely separate branch in dist-git, despite the modulemd being essentially identical except for the stream name. It would be much more convenient to have an override in the configurations section instead.

sgallagher:
Fixes: https://github.com/fedora-modularity/libmodulemd/issues/599

### DIFF
--- a/modulemd/include/modulemd-2.0/modulemd-build-config.h
+++ b/modulemd/include/modulemd-2.0/modulemd-build-config.h
@@ -111,6 +111,35 @@ modulemd_build_config_get_platform (ModulemdBuildConfig *self);
 
 
 /**
+ * modulemd_build_config_set_stream:
+ * @self: This #ModulemdBuildConfig object.
+ * @stream: A stream name to use as an override for this configuration.
+ *
+ * Set the stream name override for this build configuration.
+ *
+ * Since: 2.15
+ */
+void
+modulemd_build_config_set_stream (ModulemdBuildConfig *self,
+                                  const gchar *stream);
+
+
+/**
+ * modulemd_build_config_get_stream:
+ * @self: This #ModulemdBuildConfig object.
+ *
+ * Get the stream override for this build configuration.
+ *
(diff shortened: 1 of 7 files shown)


ppisar:
I reviewed this patch. I found these minor omissions which should be checked or implemented:

* Increase libmodulemd version
* Merge source code formatting commit
* Handle stream override in modulemd_build_config_equals()
* Handle stream override in modulemd_build_config_compare()

But I really don't like how it breaks modulemd_packager_v3_to_stream_v2() and modulemd_packager_v3_to_stream_v2_ext() (which you adjusted tests for):

The patch changes stream as a side effect in copy_packager_v3_buildconfig_to_stream_v2(). As a result modulemd_packager_v3_to_stream_v2() will assume a stream override of the latest buildconfig as a stream. It breaks modulemd_packager_v3_to_stream_v2() in the opposite direction than it's already broken in regard of buildopts. modulemd_packager_v3_to_stream_v2() adopts the first buildopts. And now it adopts the latest stream.

I understand that it stems from the ill-design of modulemd_packager_v3_to_stream_v2_ext() which produces a single modulemd-v2 object instead of a list the objects, an objects for each modulemd-packager-v3 buildconfig. I would swallow it as a necessary collateral damage. But I don't like that the damage is not in accord with buildopts. Could you please change the patch to:

* Either raise an error if the overrides differ among buildconfigs,
* or use an override of the first buildconfig as it is done with buildopts?
* Or ignore the override completely?

And document handling of the stream override in these conversion functions?

PR #604, opened 2023-01-27
Meson build tidyup

Kangie:
- Set meson_version to actual required version
- Update deprecated functions to supported equivalents

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -15,7 +15,7 @@ project(
 version : '2.14.1',
 default_options : ['buildtype=debugoptimized', 'c_std=c11', 'warning_level=1', 'b_asneeded=true'],
 license : 'MIT',
-meson_version : '>=0.47.0'
+meson_version : '>=0.58.0'
 )
 
 libmodulemd_version = meson.project_version()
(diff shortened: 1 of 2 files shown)


Kangie:
Not sure about fixing this - I'm not inclined to go do packaging work for EPEL to bring their meson up-to-scratch!

I can confirm that this PR is the bare minimum for a 'working' build system; meson 0.47 **cannot** build this due to missing features. I had to bump to >= 0.58.0 due to some of the `gtk-doc` conditionals using features that were unavailable previously:

`uses feature introduced in '0.58.0': error with more than one argument.`

TL;DR: EPEL 7 was probably broken anyway for a while, just masked by CI and not calling the `gtk-doc` codepath.

eli-schwartz:
You could, in theory, wrap said gtk-doc stuff in `if meson.version().version_compare('>=0.58.0')` and the build would then work with older versions of Meson, it would just refuse to build the docs.

I make no judgments about whether or not it's worth doing so, merely noting the option.

ppisar:
Thanks for the cleanup, however we need to support EPEL 7. EPEL 7 contains meson-0.55 and libmodulemd is packaged there under libmodulemd2 name (in order not to hijack RHEL-7-provided libmodulemd package). Therefore we cannot raise the required meson version above 0.55. So <code>meson_version : '>=0.58.0'</code> is a no go. At least now.

ppisar:
This will be addressed with #605.

PR #605, opened 2023-01-27
Increase Meson version to 0.55 and remove deprecated meson functions

ppisar:
meson-0.55.0 is required for gnome.generate_gir(). meson.source_root(), meson.build_root(), glib.get_pkgconfig_variable() were deprecated.

The oldest supported distribution is EPEL 7 with meson-0.55.1 now.

This patch is based on <https://github.com/fedora-modularity/libmodulemd/pull/604>.

### DIFF
--- a/bindings/python/meson.build
+++ b/bindings/python/meson.build
@@ -8,7 +8,7 @@ if with_py3
       check: false)
 
     if ret.returncode() != 0
-      error('Failed to determine Python 3 pygobject overridedir:', ret.stderr())
+      error('Failed to determine Python 3 pygobject overridedir: @0@'.format(ret.stderr()))
     else
       gobject_overrides_dir_py3 = ret.stdout().strip()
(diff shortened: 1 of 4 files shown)


Kangie:
Looks good, though you might want to look at the use of `error()`, the 'issue' with using < 0.58.0 is that you can only provide one parameter, and there's a few places where [that's not the case](https://github.com/fedora-modularity/libmodulemd/pull/605/files#diff-30d8f6be6320feeacf686be94f48c70869b52630e01ea625f0f15adc0d57c3e4L136).

I suppose it'll error out if it can't error properly; it's worth implementing an alternative - f-strings exist, but I think that came later than 0.55.

some string.format magic, perhaps?

ppisar:
Indeed. Old meson reports an argument number mismatch at run-time. I will fix it str.format().

PR #611, opened 2024-01-25
Drop fixxref from gtk-doc

yselkowitz:
GLib 2.79.0 switched from gtk-doc to gi-docgen, whose format and installation location are both not compatible.  This allows the docs to build with the latest GLib at the price of some broken links.

This fixes the F40FTBFS from the mass rebuild.

### DIFF
--- a/fedora/libmodulemd.spec
+++ b/fedora/libmodulemd.spec
@@ -32,7 +32,6 @@ BuildRequires:  pkgconfig(gobject-2.0)
 BuildRequires:  pkgconfig(gobject-introspection-1.0)
 BuildRequires:  pkgconfig(yaml-0.1)
 BuildRequires:  pkgconfig(gtk-doc)
-BuildRequires:  glib2-doc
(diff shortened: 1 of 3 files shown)


ppisar:
Thanks for the patch. I will try to make the fixxref optional not to regress libmodulemd documentation for people with old glib2 < 2.79.0.

I'm also considering to refer to new locations in glib2 ≥ 2.79.0 because they keep providing `*.devhelp2` indices which gtkdoc-fixxref uses for rewriting the links. Though, I briefly checked /usr/share/doc/glib-2.0/glib/glib.devhelp2 and it is missing entries for such basic types as gchar. For other, like struct GError, it uses a shorter "GError" identifier. So I'm still not persuaded to utilize the new locations.

ppisar:
When using the new path, not a single reference to glib and gobject is resolved.

sgallagher:
Closing in favor of #612

PR #612, opened 2024-01-26
Adapt to GLib 2.79 documentation

ppisar:
This patchset first moves a code in the build script, and then implements a special case GLib ≥ 2.79.0. If the special case proves unproductive, we will turned it into a skip later. This is a more compatible approach comparing to PR #611.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -67,23 +67,21 @@ sed = find_program('sed')
 test = find_program('test')
 
 with_docs = get_option('with_docs')
+gtk_doc_referred_paths = []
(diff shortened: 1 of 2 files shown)


sgallagher (approved):
It's unfortunate that the doc links are broken. I think we need to unbreak this immediately, so I'm approving the ticket.

In a follow-up patch, could we modify `.ci/ci-docs.sh` to pin the documentation generation on F39 for now, so that we push the properly-linked version to the web?

PR #613, opened 2024-01-26
CI: Arch: Edit /etc/pacman.conf instead of copying a nonexistent file

ppisar:
Base archlinux image stopped delivering a backed up configuration file for pacman and thus our CI failed:

    Run cp /etc/pacman.conf.pacnew /etc/pacman.conf
    cp: cannot stat '/etc/pacman.conf.pacnew': No such file or directory
    Error: Process completed with exit code 1.

We used to copy that file to change a pacman configuration so that GLib documentation files are not removed when installaing packages. This stopped to working.

This patch edits the file instead.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -233,8 +233,8 @@ jobs:
       image: docker.io/library/archlinux:base
 
     steps:
-      - name: Install the standard pacman config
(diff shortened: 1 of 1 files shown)


ppisar:
"Continuous Integration / Arch Linux " passed. This fix works.

PR #614, opened 2024-01-26
CI: update checkout action

sgallagher:
See https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/


ppisar (approved):
Thanks.

PR #615, opened 2024-01-29
CI: drop s390x and ppc64le, add aarch64

sgallagher:
The Github action we use to run non-x86_64 tests recently dropped support for Fedora on s390x and ppc64le for unknown reasons. This patch removes them from the test matrix, but adds back in aarch64 support.

### DIFF
--- a/.github/workflows/multiarch.yaml
+++ b/.github/workflows/multiarch.yaml
@@ -13,10 +13,6 @@ jobs:
         include:
           - arch: aarch64
             distro: fedora_latest
-          - arch: ppc64le
-            distro: fedora_latest
-          - arch: s390x
-            distro: fedora_latest
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/multiarch.yaml
+++ b/.github/workflows/multiarch.yaml
@@ -11,6 +11,8 @@ jobs:
     strategy:
       matrix:
         include:
+          - arch: aarch64
+            distro: fedora_latest
           - arch: ppc64le
             distro: fedora_latest
           - arch: s390x


ppisar:
I see. uraimo removed themit in commit <https://github.com/uraimo/run-on-arch-action/commit/18a5cd7958b1f33f6a8f10b48fdbcbf02ccda003>.

g-ir-scanner segfaulted in the new Aarch64 test. I will have to check what's wrong with it.

ppisar:
I cannot reproduce the segfault in Koji <https://koji.fedoraproject.org/koji/taskinfo?taskID=112553432>. I will retry the CI test.

sgallagher:
I'm actually going to split this into two patches (one to add aarch64 and the other to drop ppc64le and s390x) to make it easier to revert the drop if we can get those re-added to uraimo's action.

sgallagher:
I looked into this; apparently the `run-on-arch` Action uses qemu-static to run it emulated. I just tried using qemu-static to run a Fedora container manually and perform the build and I hit the same crash. It looks like it's a Qemu emulation bug and not one of ours.

Should we just abandon the `multiarch.yaml` entirely?

ppisar:
Yes. multiarch.yaml does not make sense in the current state.

We could probably augment .packit.yml to test on other architectures. See <https://packit.dev/docs/configuration#aliases> and <https://packit.dev/docs/configuration/upstream/copr_build#available-copr-build-targets>. `copr-cli list-chroots` reports aarch64, ppc64le, s390x, and x86_64 architectures for all Fedoras and ELN for me.

sgallagher:
> Yes. multiarch.yaml does not make sense in the current state.
> 
> We could probably augment .packit.yml to test on other architectures. See https://packit.dev/docs/configuration#aliases and https://packit.dev/docs/configuration/upstream/copr_build#available-copr-build-targets. `copr-cli list-chroots` reports aarch64, ppc64le, s390x, and x86_64 architectures for all Fedoras and ELN for me.

Yeah, that's probably the best approach. In the meantime, I'll close this PR and submit a new one that drops multiarch.yaml entirely.

PR #616, opened 2024-01-30
CI: Drop multiarch tests

sgallagher:
The Github Action we used to emulate other architectures has dropped support for s390x and ppc64le and has emulation bugs on aarch64.


ppisar (approved):
Good.

PR #618, opened 2024-05-15
doc: Adapt GLib documentation path to GLib 2.80.1

DimStar77:
With GLib 2.80.1 (commit 548ec9f1), the installation paths for the documentation
has changed once again.

Rationale there was:
The `gi-docgen` tool is not designed to be used like that. In
particular, when nesting documentation directories, the generated
`*.devhelp2` files (needed by Devhelp to show the documentation) are
nested one directory level too deep for Devhelp to find them, and hence
are useless, and the documentation doesn’t show up in this common
documentation viewer.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -73,12 +73,18 @@ if with_docs
 
   if glib.version().version_compare('<2.79.0')
     glib_docpath = join_paths(glib_prefix, 'share', 'gtk-doc', 'html')
-  else
+    modules = ['glib', 'gobject' ]
+  elif glib.version().version_compare('<2.80.1')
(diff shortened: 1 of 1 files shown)


ppisar:
Thanks for the patch. The changes actually happened in GLib 2.80.1. I will rework your patch and apply it.

DimStar77:
> Thanks for the patch. The changes actually happened in GLib 2.80.1. I will rework your patch and apply it.

oh, right.. 2.80.2 was the GVariant installation path fixup as a fallout from the previous move:
  https://gitlab.gnome.org/GNOME/glib/-/commit/35bccc46f913ce96ba2fe911614e31fbba3039e7

Thanks for catching that

PR #621, opened 2024-12-11
meson fixes

Kangie:
Bump the minimum meson version to reflect what is actually in-use in the project and make a boolean value actually boolean.

### DIFF
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -11,7 +11,7 @@
 #
 # REMEMBER TO UPDATE THE SUMMARY() IN meson.build when adding options here
 
-option('accept_overflowed_buildorder', type : 'boolean', value: 'false',
+option('accept_overflowed_buildorder', type : 'boolean', value: false,
        description : 'Accept overflowed 18446744073709551615 buildorder as -1. This breaks a specification, but some RHEL 8 module builds look like that.')
 
 option('verbose_tests', type : 'boolean', value : true,
--- a/modulemd/meson.build
+++ b/modulemd/meson.build
@@ -284,13 +284,15 @@ endif
 # Test env with release values
 test_release_env = environment()
 test_release_env.set('LC_ALL', 'C')
-if meson.version().version_compare('< 0.56.0')
-    project_source_root = meson.source_root()
-    project_build_root = meson.build_root()
-else
+
+if meson.version().version_compare('>=0.56.0')
     project_source_root = meson.project_source_root()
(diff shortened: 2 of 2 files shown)


ppisar:
We keep compatibility with RHEL 8 where is is meson 0.55.3. That means we cannot increase the minimal version above it.

Kangie:
> We keep compatibility with RHEL 8 where is is meson 0.55.3. That means we cannot increase the minimal version above it.

You already have - you're using features only available since 0.56.0.

eli-schwartz:
> You already have - you're using features only available since 0.56.0.
> 
> Edit: It looks like you have a case handling old meson.

https://github.com/fedora-modularity/libmodulemd/blob/1a19ce74d24b70c59868204fd584394ebc93bc7c/modulemd/meson.build#L287-L293

Meson will automatically parse `if meson.version().version_compare(XXX)` and respect its contract by not emitting warnings inside of that code block for the main project's supported version. However, this suppression only works for `if` nodes, not `else` nodes. Getting it to work correctly in an else node could be tricky as there isn't a direct map from version_compare()'s return value to "is this node being evaluated".

I don't see this style at all, to be honest. Existing uses of version_compare() conditionals inevitably tend to use `>=` comparisons, with the newer style in the `if` and the fallback code in the `else`. Under such usage, meson will avoid warning about too-new features.

eli-schwartz:
>  meson: make boolean options boolean
> 
> Later versions of meson deprecate this "feature", and every
> other boolean value in meson.options is actually a bool.

To help clarify here, the grammar has always said that `option('xxx', type: 'boolean')` accepts a `value: ` kwarg where the value is the *actual* value, not a string representing the value. Same logic applies for e.g. integer options. The reason that passing the string representation worked was actually because of an unintended CLI handling leaking into the wrong layer (since CLI parameters are always inherently strings, it needs to lex the string into rich object tokens).

It was deprecated without replacement in meson 1.1.0, since it has no use case and all existing instances in the wild are assumed to be typos. This PR fixes that typo.

sgallagher (approved):
Both of these patches look good to me now.

Getting rid of that annoying warning about the minimum version is great.

Lastly: we have no control over what version of meson is shipped in RHEL 8, because it's 1) locked due to support guarantees and 2) an OS in maintenance-only mode.

ppisar (approved):
Thanks for fixing the warnings. I confirm it works both with the latest meson as well as in RHEL 8.

Issue #623, opened 2025-04-28
issues with glib >=2.84.1: calls that pass NULL to `g_variant_store` throw critical errors


Kangie:
https://gitlab.gnome.org/GNOME/glib/-/issues/3675

> This manifests into glib critical logs currently on Fedora 42 if one tries to query a yum repo [that has modules](https://www.redhat.com/en/blog/modularity-red-hat-enterprise-linux-9) (it's dead in Fedora, but not yet in RHEL/CentOS Stream):
> 
> ```
> $ podman run -it --rm quay.io/fedora/fedora:42 dnf module list --repofrompath=tmp,'https://composes.stream.centos.org/production/latest-CentOS-Stream/compose/AppStream/$basearch/os'  --disablerepo '*' --enablerepo tmp --refresh
> Trying to pull quay.io/fedora/fedora:42...
> Getting image source signatures
> Copying blob 3130cdbe44de done   |
> Copying config 83e6cd7bf1 done   |
> Writing manifest to image destination
> Updating and loading repositories:
>  tmp                                                                                                                                                                                                                                                                                                                                                 100% |   4.3 MiB/s |   2.0 MiB |  00m00s
> 
> (process:1): GLib-CRITICAL **: 19:07:04.977: g_variant_store: assertion 'data != NULL' failed
> 
> (process:1): GLib-CRITICAL **: 19:07:04.978: g_variant_store: assertion 'data != NULL' failed
> ```

https://github.com/fedora-modularity/libmodulemd/blob/c6ffc068985ded1e3cb6875762eaa8803b63052d/modulemd/modulemd-util.c#L361-L372

The first fix that occurs to me is checking whether `variant` is `NULL` and, if so, returning a `g_variant_new_maybe` instead.

Bug: https://bugs.gentoo.org/954643

ppisar:
Thanks for the report. A minimal reproducer is processing a modulemd-v2 document with /data/xmd element having an empty flow mapping as a value:

~~~~
$ cat test.yaml 
document: modulemd
version: 2
data:
  name: test
  stream: "1"
  version: 1
  context: 1
  arch: x86_64
  summary: text
  description: text
  license:
    module:
    - MIT
  xmd: {}
$ modulemd-validator test.yaml 

(modulemd-validator:2103): GLib-CRITICAL **: 12:52:39.667: g_variant_store: assertion 'data != NULL' failed
test.yaml validated successfully
~~~~

PR #624, opened 2025-04-28
return `g_variant_new_maybe` for `NULL` variants

Kangie:
From glib 2.84.1, `g_variant_store` will not accept `NULL` values instead failing an assertion and throwing a critical error.

In the case where `variant` == `NULL`, instead we can wrap and return a `g_variant_new_maybe` with a child of `NULL` to achieve the same effective result.

Closes: https://github.com/fedora-modularity/libmodulemd/issues/623

### DIFF
--- a/modulemd/modulemd-util.c
+++ b/modulemd/modulemd-util.c
@@ -361,6 +361,12 @@ destroy_variant_data (gpointer data)
 GVariant *
 modulemd_variant_deep_copy (GVariant *variant)
 {
+  // g_variant_store will not accept NULL from 2.84.1
+  if (variant == NULL)
+    {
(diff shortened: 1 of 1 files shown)


jlebon:
Ahh I had missed this patch and opened https://github.com/fedora-modularity/libmodulemd/pull/625.

Note the patch is not correct. The variant itself is valid. It's simply that it has no children. Because GVariants don't include type information when serialized (they're expected to be provided OOB), there is nothing to serialize. The type does survive the deep copy though as written in https://github.com/fedora-modularity/libmodulemd/pull/625.

Kangie:
Let's use #625

PR #629, opened 2025-05-06
CI: Use Packit to build on non-x86_64 architectures

ppisar:
This patch enables Packit to build on AArch64, 32-bit x86, PPC64LE, and s390x platforms on Fedora Rawhide distribution. It also reenables running tests on AArch64 in the upstream spec file.

This way we should get test coverage for some common non-x86_64 architectures at a reasonable speed.

### DIFF
--- a/.packit.yml
+++ b/.packit.yml
@@ -19,6 +19,11 @@ jobs:
   metadata:
     targets:
     - fedora-all
+    - fedora-rawhide-aarch64
+    - fedora-rawhide-i386
+    - fedora-rawhide-ppc64le
+    # fedora-rawhide-x86_64 is included in fedora-all
+    # fedora-rawhide-s390x has too long wait queue
     - epel-7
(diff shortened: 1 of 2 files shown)


ppisar:
All the architectures finished quickly except s390x. 25 minutes and still unfinished. However, Packit shows the build successfully finshed in 74 seconds. Maybe some event loss between GitHub and Packit.

sgallagher:
> All the architectures finished quickly except s390x. 25 minutes and still unfinished. However, Packit shows the build successfully finshed in 74 seconds. Maybe some event loss between GitHub and Packit.

It looks like the buildSRPM step finished, but it hasn't been able to get scheduled on an s390x builder to do the actual RPM builds yet. This is a common problem; the s390x builders have a very limited capacity, so if they end up busy (such as with kernel and webkit builds), other requests can languish for a while.

PR #637, opened 2026-04-01
Drop python-six dependency

bluca:
In Ubuntu and Debian we no longer ship Python 2.7, so the translation module is an unnecessary dependency. In Python 3 the text type is always 'str', so it is not needed. Simply set the value to 'unicode' on Python2, and 'str' on Python3.

### DIFF
--- a/.ci/archlinux/Dockerfile.deps.tmpl
+++ b/.ci/archlinux/Dockerfile.deps.tmpl
@@ -13,7 +13,6 @@ RUN pacman -Syu --needed --noconfirm \
     libyaml \
     meson \
     python-gobject \
-    python-six \
 && pacman -Scc --noconfirm
 
 RUN ln -sf /builddir/bindings/python/gi/overrides/Modulemd.py $(python3 -c "import gi; import os; os.makedirs(gi._overridesdir, exist_ok=True); print(gi._overridesdir)")/Modulemd.py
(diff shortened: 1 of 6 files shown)


ppisar:
The enhancement is a good direction, but I worry it makes the only important case -- when using Python2 without six module installed difficult to debug: In that case it won't report an import failure. Instead it will assign `str` to the variable and happily continue. That means the code will behave differently.

I'd like rather see and error on Python2 without six module. That means adding a condition on Python major version. In in that case it's probably simpler to set the variable to `str` or `unicode` directly then importing the six module. Could rework the patch like that?

Also fedora/libmodulemd.spec for RPM packages still carry the dependency on the six module. That should also be adjusted.

bluca:
> The enhancement is a good direction, but I worry it makes the only important case -- when using Python2 without six module installed difficult to debug: In that case it won't report an import failure. Instead it will assign `str` to the variable and happily continue. That means the code will behave differently.
> 
> I'd like rather see and error on Python2 without six module. That means adding a condition on Python major version. In in that case it's probably simpler to set the variable to `str` or `unicode` directly then importing the six module. Could rework the patch like that?
> 
> Also fedora/libmodulemd.spec for RPM packages still carry the dependency on the six module. That should also be adjusted.

Sure, done

modulemd/tests/test_data/f29-updates.yaml
```diff
@@ -13519,7 +13519,6 @@ data:
       - python3-rpm-generators
       - python3-rpm-macros
       - python3-setuptools
-      - python3-six
```
ppisar, line 13522: Files under module/tests are test data. You shouldn't change them.

bluca: sorry just grepped without thinking through

ppisar (approved):
Great. Thanks for the cooperation.

